PR #241, opened 2026-03-08
test(CF-9s3y): TDD batch 2 — expand coverage for 6 public helpers

DreadPirateRobertz:
Added 19 new tests across 6 modules. Full suite: 11,617 tests passing.

### DIFF
--- a/tests/product360Data.test.js
+++ b/tests/product360Data.test.js
@@ -89,11 +89,14 @@ describe('register360SpinSet', () => {
     expect(get360Images('bad-images')).toEqual([]);
   });
 
-  it('ignores non-string slug types', () => {
-    register360SpinSet(123, [{ src: 'x.jpg', alt: 'X' }]);
-    register360SpinSet({}, [{ src: 'x.jpg', alt: 'X' }]);
-    register360SpinSet(true, [{ src: 'x.jpg', alt: 'X' }]);
-    // None should throw
+  it('accepts truthy non-string slugs without error but they are unretrievable', () => {
+    // register360SpinSet only guards with !slug — truthy non-strings pass through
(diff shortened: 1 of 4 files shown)


DreadPirateRobertz (commented):
### Code Review — PR #241 (melania, PM)

**REQUEST CHANGES** — 5 issues found.

1. **Hardcoded hex in referralPageHelpers tests (confidence 90)** — Assertions use raw `'#5B8FA8'`, `'#E8845C'` instead of importing from sharedTokens.js. If tokens change, tests silently break. Import `colors` from design tokens.

2. **Misleading test: "ignores non-string slug types" (confidence 92)** — `register360SpinSet(123, ...)` and `register360SpinSet(true, ...)` are actually STORED, not ignored. The `!slug` guard only rejects falsy values. Test name is wrong or needs assertions via `get360Images` to verify nothing was stored.

3. **socialProofToast mock duplication (confidence 85)** — $w mock factory copy-pasted ~8 times across describe blocks. Extract shared `createMock$w()` helper.

4. **tikTokPixel globalThis.window cleanup (confidence 82)** — If test throws between `delete globalThis.window` and restore, window stays deleted for subsequent tests. Use try/finally or afterEach.

5. **sanitizeText "strips unclosed tags" test name misleading (confidence 80)** — Also strips all text after the unclosed tag, not just the tag itself. Rename for accuracy.

**What's good:**
- Product360Data edge cases are well-constructed
- PWA beforeinstallprompt lifecycle tests are thorough
- socialProofToast frequency capping covers real business logic
- Overall these are meaningful tests, not tautological

Fix issues 1-2 (substantive), acknowledge 3-5, then resubmit.

Reviewer: melania (PM) | Generated with [Claude Code](https://claude.ai/code)

PR #245, opened 2026-03-08
fix(CF-sy7r): Defer gift card redemption to order completion

DreadPirateRobertz:
## Summary
- **Bug**: `redeemGiftCard` was called on Apply button click, deducting balance immediately. If customer abandoned checkout, balance was lost permanently.
- **Fix**: Apply button now only calls `checkBalance` and stores pending state. Added `finalizeGiftCardRedemption()` for order completion handlers and `resetCheckoutGiftCard()` for checkout abandonment.
- Security follow-up from PR #214 review.

## Changes
- `src/public/giftCardHelpers.js`: Removed `redeemGiftCard` call from Apply click handler. Added `finalizeGiftCardRedemption()`, `resetCheckoutGiftCard()`, and `pending` state field.
- `src/pages/Checkout.js`: Updated import to include new exports.
- `tests/giftCardHelpers.test.js`: Rewrote integration tests for deferred redemption flow (14 new/updated tests).

## Test plan
- [x] Apply click only calls `checkBalance`, NOT `redeemGiftCard`
- [x] State shows `pending: true` after Apply, `applied: false`
- [x] `finalizeGiftCardRedemption()` calls `redeemGiftCard` and transitions to `applied: true`
- [x] No-op when no gift card pending
- [x] Double finalization prevented (only 1 `redeemGiftCard` call)
- [x] Failed/errored redemption clears state
- [x] `resetCheckoutGiftCard()` clears pending without deducting
- [x] Full suite: 11,636 tests passing (303 files)

### DIFF
--- a/src/pages/Checkout.js
+++ b/src/pages/Checkout.js
@@ -61,6 +61,17 @@ $w.onReady(async function () {
   try { initCheckoutFocusIndicators(); } catch (e) {}
   try { collapseOnMobile($w, ['#checkoutFinancing', '#expressCheckoutSection']); } catch (e) {}
   try { initBackToTop($w); } catch (e) {}
+
+  // CF-sy7r: Reset gift card state on checkout abandon (page navigation away).
+  // finalizeGiftCardRedemption is called by Thank You page on order completion.
+  try {
+    const wixWindow = await import('wix-window-frontend');
+    if (wixWindow.onBeforeUnload) {
+      wixWindow.onBeforeUnload(() => {
+        resetCheckoutGiftCard();
+      });
+    }
+  } catch (e) {}
 });
 
 // ── Checkout Progress Indicator ──────────────────────────────────────
--- a/src/pages/Thank You Page.js
+++ b/src/pages/Thank You Page.js
@@ -13,11 +13,27 @@ import { validateEmail, sanitizeText } from 'public/validators.js';
 import { markSessionConverted } from 'backend/browseAbandonment.web';
 import { getReferralLink } from 'backend/referralService.web';
 import { submitReview } from 'backend/reviewsService.web';
+import { finalizeGiftCardRedemption } from 'public/giftCardHelpers.js';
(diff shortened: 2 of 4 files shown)


DreadPirateRobertz (commented):
### Code Review — PR #245 (melania, PM)

**REQUEST CHANGES** — 1 critical blocker, 2 high issues.

**CRITICAL: finalizeGiftCardRedemption and resetCheckoutGiftCard are imported but NEVER CALLED in Checkout.js (confidence 99)**

The import line was updated but no call sites were added. The old immediate deduction is removed, but nothing replaces it — gift card balances are validated but never actually deducted. This is a functional regression: orders with gift cards complete without balance reduction.

Fix: Wire `finalizeGiftCardRedemption()` into the order completion handler and `resetCheckoutGiftCard()` into checkout abandonment.

**HIGH: Race condition in finalizeGiftCardRedemption (confidence 85)**

Double-call protection checks `_giftCardState.pending` then awaits `redeemGiftCard()`. Concurrent callers (double-click "Place Order") both read `pending: true` before either completes. The "prevents double finalization" test only tests sequential calls.

Fix: Set `pending = false` synchronously BEFORE the await, or use a promise-based lock.

**HIGH: Stale amountToApply between Apply and order completion (confidence 82)**

`amountToApply` is captured at Apply-click time. If cart changes before order completion (add/remove items, coupon), amount is stale. Could over-deduct or under-deduct.

Fix: Re-calculate `amountToApply` in `finalizeGiftCardRedemption` using current subtotal, or accept subtotal as parameter.

**What's good:**
- Deferred redemption architecture is sound
- Unit tests for the new functions are thorough
- Double finalization prevention (sequential case) works
- resetCheckoutGiftCard clears state properly

Fix the Checkout.js wiring (blocker), then address the race condition and stale amount. Resubmit.

Reviewer: melania (PM) | Generated with [Claude Code](https://claude.ai/code)

PR #261, opened 2026-03-09
feat(CF-kx0b): Bulk import 313 product photos to Wix Media Manager

DreadPirateRobertz:
## Summary
- TDD script (`scripts/importProductPhotos.js`) that bulk imports product photos to Wix Media Manager via REST API
- Creates 7 product category folders (`futon-frames`, `wall-hugger-frames`, `platform-beds`, `mattresses`, `murphy-cabinet-beds`, `front-loading-nesting`, `casegoods-accessories`) under `/products` root
- Imports all 313 images from `docs/product-image-manifest.json` in batches of 100 (Wix API limit)
- Supports `--dry-run` mode for safe previewing
- 29 tests covering: manifest loading, folder extraction, MIME inference, batch chunking, API calls, dry-run mode, partial failure handling, config validation

## Import Results
- **313/313 images imported successfully** (0 failures)
- 4 batches processed, 7 folders created
- All images now on account `ed8a72` (our account) instead of `e04e89`

## Test plan
- [x] 29 new tests pass (TDD — tests written before implementation)
- [x] Full suite: 11,864 tests pass (306 files)
- [x] Dry run verified correct manifest parsing (88 products, 313 images, 7 folders)
- [x] Live run completed: 313/313 succeeded, 0 failed
- [ ] Verify images visible in Wix Studio Media Manager

### DIFF
--- a/scripts/importProductPhotos.js
+++ b/scripts/importProductPhotos.js
@@ -60,16 +60,25 @@ export function extractUniqueFolders(manifest) {
 
 /**
  * Infer MIME type from a wixstatic URL.
+ * Handles two URL formats:
+ *   - /media/hash_id.ext/v1/fit/... (current manifest format)
+ *   - /media/hash~mv2.ext/v1/fit/... (legacy Wix format)
+ * Falls back to image/jpeg if extension cannot be determined.
  * @param {string} url
  * @returns {string} MIME type string.
  */
 export function inferMimeType(url) {
-  // wixstatic URLs: .../media/hash~mv2.ext/v1/fit/.../file.ext
-  const match = url.match(/~mv2\.(\w+)/);
-  if (match) {
-    return MIME_MAP[match[1].toLowerCase()] || 'image/jpeg';
+  // Primary: match /media/hash.ext/ pattern (e.g., e04e89_cf15.jpg/v1/...)
+  const mediaMatch = url.match(/\/media\/[^/]+\.(\w+)\//);
+  if (mediaMatch) {
+    return MIME_MAP[mediaMatch[1].toLowerCase()] || 'image/jpeg';
   }
-  // Fallback: check file extension at end
+  // Legacy: match ~mv2.ext pattern
+  const mv2Match = url.match(/~mv2\.(\w+)/);
+  if (mv2Match) {
+    return MIME_MAP[mv2Match[1].toLowerCase()] || 'image/jpeg';
+  }
+  // Last resort: file extension at end of URL
   const extMatch = url.match(/\.(\w+)(?:\?|$)/);
   if (extMatch) {
     return MIME_MAP[extMatch[1].toLowerCase()] || 'image/jpeg';
@@ -88,15 +97,19 @@ export function buildImportRequests(manifest, folderIds) {
   for (const product of manifest) {
     if (!product.images || !product.images.length) continue;
     const parentFolderId = folderIds[product.folder] || 'media-root';
-    product.images.forEach((url, idx) => {
+    if (!folderIds[product.folder] && product.folder) {
+      console.warn(`Warning: No folder ID for "${product.folder}" — images will go to media-root`);
+    }
+    const slug = product.slug || String(product.name || 'product').toLowerCase().replace(/\s+/g, '-');
+    for (let idx = 0; idx < product.images.length; idx++) {
       requests.push({
-        url,
-        displayName: `${product.slug}-${idx + 1}`,
-        mimeType: inferMimeType(url),
+        url: product.images[idx],
+        displayName: `${slug}-${idx + 1}`,
+        mimeType: inferMimeType(product.images[idx]),
         mediaType: 'IMAGE',
         parentFolderId,
       });
-    });
+    }
   }
   return requests;
 }
@@ -197,12 +210,22 @@ export async function ensureFolderStructure(folderPaths, config) {
   const rootFolder = await createFolder('products', 'media-root', config);
   const rootId = rootFolder.id;
 
-  // Create each category sub-folder
+  // Create each category sub-folder (continue on individual failures)
   const folderIds = {};
+  const folderErrors = [];
   for (const path of folderPaths) {
     const categoryName = path.split('/').pop();
-    const folder = await createFolder(categoryName, rootId, config);
-    folderIds[path] = folder.id;
+    try {
+      const folder = await createFolder(categoryName, rootId, config);
+      folderIds[path] = folder.id;
+    } catch (err) {
+      console.error(`Failed to create folder "${categoryName}": ${err.message}`);
+      folderErrors.push({ path, error: err.message });
+    }
+  }
+
+  if (folderErrors.length) {
+    console.warn(`${folderErrors.length} folder(s) failed — their images will go to media-root`);
   }
 
   return folderIds;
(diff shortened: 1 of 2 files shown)


DreadPirateRobertz:
**Comment Analyzer** (dispatched by miquella):

## Summary

Reviewed all JSDoc and inline comments in `scripts/importProductPhotos.js` (308 lines) and `tests/importProductPhotos.test.js` (508 lines). The PR introduces a bulk photo import script with 29 tests. Overall comment quality is solid -- JSDoc is present on every exported function, parameter types are documented, and the module-level docblock provides clear usage instructions.

---

## Critical Issues

None found. All JSDoc signatures match their actual implementations, and no comments make factually incorrect claims.

---

## Improvement Opportunities

**1. `inferMimeType` JSDoc understates the function's behavior**
- Location: `scripts/importProductPhotos.js`, `inferMimeType` JSDoc
- Current: `@param {string} url` / `@returns {string} MIME type string.`
- The inline comment `// wixstatic URLs: .../media/hash~mv2.ext/v1/fit/.../file.ext` documents one URL pattern, but the function also has a generic fallback regex and a hardcoded default of `image/jpeg`. The JSDoc does not mention the fallback behavior or the default return value.
- Suggestion: Add `@returns {string} MIME type string, defaults to 'image/jpeg' if extension cannot be inferred.`

**2. `buildImportRequests` JSDoc `@param folderIds` type is imprecise**
- Location: `scripts/importProductPhotos.js`, `buildImportRequests` JSDoc
- Current: `@param {Object<string, string>} folderIds - Map of folder path -> Wix folder ID.`
- `Object<string, string>` is not standard JSDoc. The conventional form is `@param {Object.<string, string>}` (with the dot) or `@param {Record<string, string>}` for TypeScript-flavored JSDoc.
- Minor, but worth normalizing for tooling consistency.

**3. `ensureFolderStructure` comment assumes single-depth hierarchy**
- Location: `scripts/importProductPhotos.js`, `ensureFolderStructure` inline comment `// Create root "products" folder`
- The function hardcodes the assumption that all folder paths are exactly two levels deep (`/products/<category>`). The JSDoc example `['/products/futon-frames', '/products/mattresses']` correctly reflects this, but neither the JSDoc nor inline comments call out that deeper nesting (e.g., `/products/frames/wood`) would silently break -- `path.split('/').pop()` would create `wood` under `products`, skipping `frames`.
- Suggestion: Add a precondition note: `@param {string[]} folderPaths - Expects exactly two-level paths (/products/<category>). Deeper nesting is not supported.`

**4. `bulkImportFiles` JSDoc could document the API endpoint**
- Location: `scripts/importProductPhotos.js`, `bulkImportFiles` JSDoc
- The function calls `/bulk/files/import-v2` but the JSDoc only says "Bulk import files via Wix REST API." Since this is an integration script that may need maintenance as APIs evolve, documenting the specific endpoint or linking to Wix docs would aid future maintainers.
- Suggestion: Add `@see https://dev.wix.com/docs/rest/api-reference/site-media/bulk-import-files` or similar.

**5. `runImport` JSDoc `@param options` type is incomplete**
- Location: `scripts/importProductPhotos.js`, `runImport` JSDoc
- Current: `@param {{ dryRun?: boolean }} [options]`
- This is correct but the `@returns` type `Promise<Object>` is too vague. The function returns two distinct shapes depending on `dryRun` -- one with `{dryRun, totalImages, folders, batches}` and another with `{totalImages, totalSuccess, totalFailed, batches, failures}`.
- Suggestion: Document both return shapes with `@returns {Promise<{dryRun: true, totalImages: number, folders: string[], batches: number} | {totalImages: number, totalSuccess: number, totalFailed: number, batches: number, failures: Array}>}`

---

## Recommended Removals

None. All comments add value. The test file correctly avoids over-commenting -- test names serve as documentation.

---

## Positive Findings

- **Module-level JSDoc is excellent**: Clear usage instructions with both direct env var and `secrets.env` sourcing patterns. This is the kind of comment that saves a developer 10 minutes on first encounter.
- **Every exported function has JSDoc**: Consistent discipline across all 10 exports.
- **Inline comments are used sparingly and explain "why"**: The `// wixstatic URLs:` pattern comment and `// Track individual failures` comments explain non-obvious logic rather than restating code.
- **Test helper functions have clear names that serve as documentation**: `folderResponse()`, `bulkImportResponse()`, `errorResponse()` -- no JSDoc needed, the signatures self-document.
- **The `BATCH_SIZE` constant is used in both validation and chunking**: The comment `@param {Array<Object>} importFileRequests - Max 100 items.` correctly references the actual enforced limit.

DreadPirateRobertz:
**PR Test Analyzer** (dispatched by miquella):

## Summary

29 tests covering a 308-line import script. Overall test quality is **solid** — the tests follow TDD discipline, cover the major code paths, mock external dependencies correctly, and verify both happy paths and several error scenarios. The structure is clean and the helpers for mock responses reduce boilerplate. A few gaps worth addressing:

---

## Critical Gaps

### 1. `ensureFolderStructure` — root folder creation failure not tested (Criticality: 8/10)

If the "products" root folder creation fails (API 403, network timeout, etc.), the function will throw, but this is never tested. Since **all** sub-folder creation depends on the root folder succeeding, this is a critical path. A single test asserting that a root folder API failure propagates correctly would close this gap.

**What it catches:** A refactor that accidentally swallows the root folder error would silently produce an empty `folderIds` map, causing all images to land in `media-root` instead of their category folders — a silent data corruption bug.

### 2. `ensureFolderStructure` — sub-folder creation failure mid-sequence not tested (Criticality: 8/10)

If the 3rd of 7 category folders fails, the function currently throws and stops. There is no test verifying this behavior. Depending on whether partial folder creation is acceptable, the function might need to be made resilient (continue on error, collect failures). Either way, the current behavior should be tested and documented.

**What it catches:** A future change that adds try/catch around individual folder creation could silently skip folders, leaving images orphaned in `media-root`.

### 3. `runImport` — `bulkImportFiles` throws mid-batch (Criticality: 8/10)

The test for "partial batch failures" covers the case where individual items fail within a successful API response. But there is no test for when `bulkImportFiles` itself throws (e.g., API returns 500 on batch 2 of 4). Currently `runImport` has no try/catch around the batch loop, meaning batches 3 and 4 would never be attempted and already-imported images from batch 1 would be orphaned with no summary returned.

**What it catches:** In production, transient API errors on one batch would lose the entire import with no recovery or status report. A test here would force a decision: should the import continue on batch failure, or fail fast with a partial summary?

---

## Important Improvements

### 4. `runImport` with manifest containing products with no images array (Criticality: 6/10)

`buildImportRequests` guards against `!product.images || !product.images.length`, but `runImport` uses `manifest.reduce((sum, p) => sum + (p.images ? p.images.length : 0), 0)` for the total count. While both handle missing images, there is no integration-level test confirming `runImport` works end-to-end with mixed manifest entries (some products with images, some without). The unit test for `buildImportRequests` covers the empty case in isolation but not through the orchestrator.

### 5. `inferMimeType` — `~mv2` pattern not specifically tested (Criticality: 5/10)

The wixstatic URLs used in this codebase follow the pattern `hash~mv2.ext`. The `inferMimeType` regex specifically looks for `~mv2\.(\w+)`. The test uses full wixstatic URLs that happen to match both the `~mv2` regex AND the fallback extension regex. There is no test that isolates the `~mv2` regex path vs the fallback path. Consider a test with a URL like `https://static.wixstatic.com/media/e04e89_cf15~mv2.png/v1/fit/w_2000/file.jpg` where the `~mv2` extension (png) differs from the trailing extension (jpg) to verify which takes precedence.

### 6. `getConfig` — env var restoration uses object spread (Criticality: 5/10)

The `getConfig` tests do `const original = { ...process.env }; ... process.env = original;`. Reassigning `process.env` works in Node but is fragile — it replaces the entire env object. If Vitest or other test infrastructure caches a reference to the original `process.env`, mutations on the new object won't be visible. The safer pattern is `vi.stubEnv()` / `vi.unstubAllEnvs()` or saving/restoring individual keys.

---

## Test Quality Issues

### 7. `beforeEach` dynamic import may not fully reset module state

The `beforeEach` does `vi.resetModules()` then re-imports the module. However, the module has top-level side effects (`const __filename`, `const __dirname`, the CLI guard). If `vi.resetModules()` doesn't fully clear the ESM module cache in all Vitest versions, tests could share state. This is unlikely to cause issues today but is worth noting.

---

## Positive Observations

- **Dry-run mode**: Properly tested to confirm zero API calls — this is exactly the kind of behavioral test that provides confidence.
- **Batch size enforcement**: The 101-item rejection test on `bulkImportFiles` is a good boundary condition test.
- **Partial failure tracking**: Testing that individual item failures within a successful bulk response are collected in the `failures` array is excellent — this is a real production scenario.
- **Error propagation**: API error codes (403, 400) are tested to verify they propagate with status information.
- **Network errors**: `createFolder` has a dedicated network error test.
- **Manifest edge cases**: Empty manifests, missing folder fields, and missing images are all covered at the unit level.
- **Mock structure**: The helper functions (`folderResponse`, `bulkImportResponse`, `errorResponse`) make tests readable and maintainable.

Overall: **Good coverage with a few meaningful gaps around failure propagation in the orchestration layer** (items 1-3). The unit-level function tests are thorough. The main risk is in `runImport` and `ensureFolderStructure` where cascading failures are not tested.

DreadPirateRobertz:
**Code Reviewer** (dispatched by miquella):

**1 critical, 3 important issues found.**

### Critical — `process.env = original` doesn't restore env (Confidence: 90)
`tests/importProductPhotos.test.js` getConfig tests: `process.env = original` silently fails in Node.js — env mutations bleed across tests. Fix: use `vi.stubEnv`/`vi.unstubAllEnvs`.

### Important — Primary regex in `inferMimeType` is dead code (Confidence: 85)
`/~mv2\.(\w+)/` never matches actual manifest URLs (format: `e04e89_cf15.jpg/v1/...`). All MIME detection falls through to the extension fallback. Dead primary path.

### Important — `ensureFolderStructure` creates duplicates on re-run (Confidence: 82)
No existence check before `createFolder('products', ...)`. Re-running creates duplicate folder trees.

### Important — `buildImportRequests` missing `slug` guard (Confidence: 80)
No `slug` field → `displayName: "undefined-1"`. Needs fallback.

**Summary:** Core logic solid. Fix env restoration and add guards before merge.

PR #262, opened 2026-03-09
test(CF-2yxz): element ID verification test script

DreadPirateRobertz:
## Summary
- Adds `src/public/elementIdValidator.js` — extracts and validates $w element IDs from source files
- Adds `tests/elementIdVerification.test.js` — 67 tests covering:
  - Unit tests for extraction (single/double quotes, dedup, elementId props, edge cases)
  - Unit tests for validation (camelCase, empty, numbers, special chars, hash prefix)
  - Full verification across all 39 page modules (~792 unique IDs)
  - Public helper module verification (~363 IDs)
  - Cross-page shared ID analysis (informational)

## Test plan
- [x] All 67 new tests pass
- [x] Full suite: 12,014 tests pass (307 files)
- [x] TDD: RED → GREEN → REFACTOR cycle followed
- [x] No browser/dashboard work required

### DIFF
--- a/src/public/elementIdValidator.js
+++ b/src/public/elementIdValidator.js
@@ -1,32 +1,36 @@
 /**
  * Element ID Validator — Static analysis for Wix Velo $w element IDs.
  * Extracts and validates all $w('#id') references from page source code.
+ *
+ * Note: This module contains only pure functions (no fs/path imports)
+ * so it remains Wix Velo compatible. File I/O lives in test code.
  * @module elementIdValidator
  */
-import { readFileSync, readdirSync } from 'fs';
-import { join, basename } from 'path';
 
-/** Valid Wix element ID: starts with letter, alphanumeric only */
+/** Valid Wix element ID: starts with letter, alphanumeric only (no underscores) */
 const VALID_ID_RE = /^[a-zA-Z][a-zA-Z0-9]*$/;
 
 /**
  * Extract unique $w element IDs from source code.
+ * Captures broadly (letters, digits, underscores) so that IDs with underscores
+ * are extracted and then caught by validateElementId as invalid.
  * @param {string} source - JavaScript source code
  * @returns {string[]} Array of unique element IDs (without # prefix)
  */
 export function extractElementIds(source) {
+  if (typeof source !== 'string') return [];
+
   const ids = new Set();
 
-  // Extract broadly (including underscores) — validation rejects invalid IDs later
   const patterns = [
-    /\$w\(['"]#([a-zA-Z0-9_]*)['"]\)/g,       // $w('#id') and $w("#id")
-    /elementId:\s*['"]#([a-zA-Z0-9_]*)['"]/g,  // elementId: '#id' in config objects
+    /\$w\(['"]#([a-zA-Z0-9_]+)['"]\)/g,       // $w('#id') and $w("#id")
+    /elementId:\s*['"]#([a-zA-Z0-9_]+)['"]/g,  // elementId: '#id' in config objects
   ];
 
   for (const re of patterns) {
     let match;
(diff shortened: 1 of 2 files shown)


DreadPirateRobertz:
## Code Simplifier Review

**code-simplifier agent dispatched by godfrey**

Reviewed `src/public/elementIdValidator.js` and `tests/elementIdVerification.test.js` for simplification opportunities.

### src/public/elementIdValidator.js

**1. The `VALID_ID_RE` regex already encodes all validation logic — `validateElementId` can be simplified.**

Currently `validateElementId` performs four sequential if-checks that partially overlap with each other and with `VALID_ID_RE`. The hash-prefix check and the starts-with-letter check are subsets of what the final regex test already catches. However, each branch returns a distinct `reason` string, which is valuable for developer experience. The current structure is therefore justified — each condition produces a different diagnostic message. **No change recommended.**

**2. `extractElementIds` — minor: the two regex loops share identical structure.**

Both the `$w('#id')` and `elementId: '#id'` extraction loops follow the same pattern: exec in a while-loop, add to a Set. You could consolidate into a helper or iterate over an array of regexes:

```js
const patterns = [
  /\$w\(['"]#([a-zA-Z0-9_]*)['"]\)/g,
  /elementId:\s*['"]#([a-zA-Z0-9_]*)['"]/g,
];
for (const re of patterns) {
  let match;
  while ((match = re.exec(source)) !== null) {
    if (match[1]) ids.add(match[1]);
  }
}
```

This eliminates the duplicated while-loop and makes it trivial to add new patterns in the future. Low priority — the current version is only 10 lines and perfectly readable.

**3. `VALID_ID_RE` allows underscores but `validateElementId` reason says "alphanumeric only".**

The regex `/^[a-zA-Z][a-zA-Z0-9]*$/` does NOT allow underscores, but the extraction regexes use `[a-zA-Z0-9_]*` which DOES capture IDs containing underscores. An ID like `my_button` would be extracted and then flagged as invalid with the message "must be alphanumeric" — which is correct behavior, but the mismatch between extraction charset and validation charset is worth a brief inline comment explaining the intent (extract broadly, validate strictly).

### tests/elementIdVerification.test.js

**4. IIFE for `pageFiles` is unnecessary — use a simple `let` with try/catch.**

```js
const pageFiles = (() => {
  try { return getPageFiles(); }
  catch { return []; }
})();
```

This IIFE exists only to handle a potential error at module scope. It works fine but a `let` with try/catch at the describe block level would be more conventional:

```js
let pageFiles = [];
try { pageFiles = getPageFiles(); } catch { /* no pages dir */ }
```

Marginal improvement — either form is acceptable.

**5. The "detects no duplicate IDs" test name is misleading.**

The test is titled "detects no duplicate IDs used across unrelated pages (informational)" but it always passes regardless of duplicates. The test body acknowledges this ("informational — it passes but logs shared IDs") yet it does not actually log anything. Consider either:
- Adding a `console.log` for the shared IDs count so it actually serves its informational purpose, or
- Removing the test entirely since it asserts nothing meaningful (`Array.isArray(sharedIds)` is trivially true).

If the intent is to keep it as a placeholder for future enforcement, a comment stating that would be clearer than the current `expect(Array.isArray(sharedIds)).toBe(true)` assertion.

**6. `totalUnique` in the "reports total ID count" test — use `const` consistently.**

```js
let totalUnique = new Set();
let perPage = {};
```

Both variables are never reassigned. Use `const` for both, consistent with the rest of the codebase.

### Summary

Overall this is clean, well-structured code. The validator module is focused and the test coverage is thorough. The items above are minor polish — no blocking issues found. The most actionable item is **#3** (extraction/validation charset mismatch comment) and **#6** (`const` consistency).

PR #296, opened 2026-03-14
feat(cf-4nu): Template product cleanup script

DreadPirateRobertz:
## Summary
- Adds `scripts/deleteTemplateProducts.js` to remove Furniture Store template products from Wix Store
- Queries all products via Wix Stores REST API, classifies as template vs CF, deletes template products
- Supports `--dry-run` flag for safe preview
- Conservative: unknown products are kept (never deleted)

## How it works
1. Loads known CF product names from catalog JSONs + manual list
2. Queries all products from Wix Stores REST API (paginated)
3. Classifies each product:
   - **CF** (keep): matches catalog names or manual list
   - **Template** (delete): matches 24 template patterns (MODO, NYX, RAVEN, etc.)
   - **Unknown** (keep): doesn't match either — conservative default
4. Deletes template products with per-product error handling

## Usage
```bash
source scripts/secrets.env
WIX_SITE_ID=3af610bf-... node scripts/deleteTemplateProducts.js --dry-run
WIX_SITE_ID=3af610bf-... node scripts/deleteTemplateProducts.js
```

## Test plan
- [x] 17 tests covering classification logic, edge cases, case-insensitivity
- [x] Full suite green (12,265 tests, 315 files)
- [ ] Dry-run on live site before actual deletion
- [ ] CI passes

### DIFF
--- a/scripts/deleteTemplateProducts.js
+++ b/scripts/deleteTemplateProducts.js
@@ -3,9 +3,9 @@
  * deleteTemplateProducts.js — Remove Furniture Store template products from Wix Store.
  *
  * The Furniture Store template (#3563) ships with ~24 sample products (MODO, NYX,
- * RAVEN, etc.). This script queries all products, identifies template products
- * (those NOT in our Carolina Futons catalog), and deletes them via the Wix Stores
- * REST API.
+ * RAVEN, etc.). This script queries all products, classifies them into three
+ * buckets (CF products, template products, unknown), and deletes only confirmed
+ * template products. Unknown products are kept (conservative default).
  *
  * Usage:
  *   WIX_BACKEND_KEY=IST.xxx WIX_SITE_ID=xxx node scripts/deleteTemplateProducts.js [--dry-run]
@@ -29,7 +29,7 @@ const DRY_RUN = process.argv.includes('--dry-run');
 // ── Known CF product names (from our scraped catalog) ──────────────
 // Products with these names are OURS — do NOT delete them.
 
-function loadCFProductNames() {
+export function loadCFProductNames() {
   const names = new Set();
   const catalogFiles = [
     resolve(__dirname, '..', 'content', 'scraped-products-16-30.json'),
@@ -43,7 +43,9 @@ function loadCFProductNames() {
         if (product.name) names.add(product.name.trim().toLowerCase());
       }
     } catch (e) {
-      // File may not exist — that's ok, we'll be conservative
+      if (e.code !== 'ENOENT') {
+        console.error(`Warning: failed to read catalog file ${file}: ${e.message}`);
+      }
     }
   }
 
@@ -63,8 +65,10 @@ function loadCFProductNames() {
 
 // ── Known template product name patterns ───────────────────────────
 // These are products from the Furniture Store template #3563.
+// Note: Some short names (aria, sage, etc.) could collide with CF names.
+// CF match takes priority over template match, so collisions are safe.
 
-const TEMPLATE_PRODUCT_PATTERNS = [
+export const TEMPLATE_PRODUCT_PATTERNS = [
   'modo', 'nyx', 'raven', 'oslo', 'aria', 'luna', 'nova', 'zen',
   'cleo', 'milo', 'otto', 'vega', 'aura', 'echo', 'iris', 'onyx',
   'jade', 'opal', 'ruby', 'sage', 'teak', 'wren', 'yuma', 'zara',
@@ -72,11 +76,11 @@ const TEMPLATE_PRODUCT_PATTERNS = [
 
 /**
  * Query all products from Wix Stores.
- * @param {string} apiKey - Wix API key
+ * @param {string} token - Wix IST (Installation Security Token)
  * @param {string} siteId - Wix site ID
  * @returns {Promise<Array>} All products
  */
-async function queryAllProducts(apiKey, siteId) {
+async function queryAllProducts(token, siteId) {
   const products = [];
   let offset = 0;
   const limit = 100;
@@ -86,7 +90,7 @@ async function queryAllProducts(apiKey, siteId) {
       method: 'POST',
       headers: {
         'Content-Type': 'application/json',
-        'Authorization': apiKey,
+        'Authorization': token,
         'wix-site-id': siteId,
       },
       body: JSON.stringify({
@@ -115,14 +119,15 @@ async function queryAllProducts(apiKey, siteId) {
 /**
  * Delete a product by ID.
  * @param {string} productId
- * @param {string} apiKey
+ * @param {string} token - Wix IST
  * @param {string} siteId
+ * @throws {Error} If the API returns a non-OK response
  */
-async function deleteProduct(productId, apiKey, siteId) {
+async function deleteProduct(productId, token, siteId) {
   const res = await fetch(`${WIX_STORES_API}/products/${productId}`, {
     method: 'DELETE',
     headers: {
-      'Authorization': apiKey,
+      'Authorization': token,
       'wix-site-id': siteId,
     },
   });
@@ -134,14 +139,16 @@ async function deleteProduct(productId, apiKey, siteId) {
 }
 
 /**
- * Identify template products vs CF products.
+ * Classify products into three buckets: confirmed CF, confirmed template,
+ * and unknown (routed to keep/CF bucket for safety).
+ * CF match takes priority — if a product matches both CF and template, it's kept.
  * @param {Array} products - All products from Wix
  * @param {Set<string>} cfNames - Known CF product names (lowercase)
- * @returns {{ template: Array, cf: Array }}
+ * @returns {{ template: Array, keep: Array }}
  */
-function classifyProducts(products, cfNames) {
+export function classifyProducts(products, cfNames) {
   const template = [];
-  const cf = [];
+  const keep = [];
 
   for (const product of products) {
     const name = (product.name || '').trim();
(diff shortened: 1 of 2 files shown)


DreadPirateRobertz (commented):

I have audited every error handling path in this PR. This is a destructive script that **permanently deletes products from a live store** — error handling standards must be exceptionally high. I found several issues.

---

### CRITICAL-1: Silent swallow of catalog file read errors (line 42-44)

```js
} catch (e) {
  // File may not exist — that's ok, we'll be conservative
}
```

**Issue:** This empty catch block swallows ALL errors, not just "file not found." It will silently hide:
- JSON parse errors from corrupted/malformed catalog files (SyntaxError)
- Permission denied errors (EACCES)
- Encoding errors from binary files

If the catalog files exist but contain malformed JSON, `cfNames` will be incomplete, and the script will misclassify legitimate CF products as "unknown." While unknowns are kept (conservative), the user gets zero indication that their catalog data failed to load.

**User impact:** If both catalog files fail to parse, the script runs with ONLY the 20 manual names. Products loaded from the catalog JSONs would not be recognized as CF products. The user sees "Loaded 20 known CF product names" instead of a much larger number and has no idea why.

**Recommendation:**
```js
} catch (e) {
  if (e.code === 'ENOENT') {
    console.log(`  ℹ Catalog file not found (skipping): ${file}`);
  } else {
    console.error(`  ⚠ Failed to read catalog file ${file}: ${e.message}`);
    console.error(`    CF product list may be incomplete — review before deleting.`);
  }
}
```

---

### CRITICAL-2: Same empty catch duplicated in test file (test line 30-31)

```js
} catch (e) { /* file may not exist */ }
```

The test file duplicates the same silent swallow. The test's `loadCFProductNames` will silently produce an incomplete name set, and test assertions against `cfNames.size` will pass with a smaller-than-expected set. The test at line 87 (`expect(cfNames.size).toBeGreaterThan(10)`) only checks >10, so it passes even if catalog loading completely fails (20 manual names exist). This means the test does not actually validate catalog file loading — it only validates the manual list.

---

### HIGH-1: Per-product delete errors are logged and swallowed (lines 214-218)

```js
} catch (err) {
  failed++;
  console.error(`  ❌ Failed to delete "${p.name}": ${err.message}`);
}
```

The script logs the failure and increments `failed`, but the process exits with code 0 regardless of how many deletions failed. If the API key is expired or the site ID is wrong, EVERY delete will fail, the script prints a wall of errors, and then exits successfully.

**User impact:** CI or automation wrapping this script will see exit code 0 and believe the cleanup succeeded.

**Recommendation:** After the deletion loop, exit with a non-zero code if any deletions failed:
```js
console.log(`\nDone. Deleted: ${deleted}, Failed: ${failed}`);
if (failed > 0) {
  process.exit(2);
}
```

---

### HIGH-2: No validation that queryAllProducts returned meaningful data (line 182)

After querying products, the script proceeds directly to classification with no sanity check. If the API returns an empty array due to auth issues, rate limiting, or an incorrect site ID, the script happily reports "No template products found. Nothing to delete." and exits successfully.

**User impact:** The user thinks there are no template products to clean up, when in reality the query silently returned nothing. This is especially dangerous because the Wix API may return 200 OK with an empty `products` array for invalid site IDs.

**Recommendation:**
```js
const allProducts = await queryAllProducts(apiKey, siteId);
if (allProducts.length === 0) {
  console.warn('⚠ Query returned 0 products. Verify WIX_SITE_ID is correct.');
  console.warn('  If this site truly has no products, this is expected.');
}
```

---

### MEDIUM-1: `res.json()` call has no error handling (line 102)

```js
const data = await res.json();
```

If the API returns a 200 status with non-JSON body (e.g., HTML error page behind a CDN, or a rate-limit page), `res.json()` will throw an unhandled `SyntaxError` with a cryptic message like "Unexpected token < in JSON at position 0." This error would propagate to `main().catch()` and print "Fatal error: [SyntaxError]" — unhelpful for the user.

**Recommendation:** Wrap in try-catch with a clear message:
```js
let data;
try {
  data = await res.json();
} catch (parseErr) {
  const text = await res.text().catch(() => '[unreadable]');
  throw new Error(`API returned non-JSON response at offset ${offset}: ${text.slice(0, 200)}`);
}
```

---

### MEDIUM-2: No rate limiting or backoff on API calls

The script fires delete requests in a tight serial loop with no delay. Wix Stores API has rate limits. If the script hits them, deletes will start failing mid-run with no retry. Combined with HIGH-1 (exit code 0 on failure), the user gets a partial deletion with a "success" exit.

---

### MEDIUM-3: Test file duplicates production logic instead of importing it (entire test file)

The test file copy-pastes `classifyProducts`, `loadCFProductNames`, and `TEMPLATE_PRODUCT_PATTERNS` instead of importing from the source. If the production code's classification logic is modified (e.g., a bug fix), the tests will continue passing against the OLD copy-pasted logic. This is not a silent failure per se, but it creates a maintenance trap where tests give false confidence.

**Recommendation:** Refactor the script to export its functions (conditionally run `main()` only when executed directly), then import them in tests.

---

### GOOD: Things done well

- `queryAllProducts` properly checks `res.ok` and throws with status + response body — good context.
- `deleteProduct` similarly throws with product ID and status — good for debugging.
- The `main().catch()` at the bottom ensures unhandled rejections are caught and exit non-zero.
- Conservative default (unknown products are kept) is the right call for a destructive script.
- CF-over-template priority in classification prevents accidental deletion of dual-matched products.
- Environment variable validation at the top of `main()` with clear usage message is solid.

DreadPirateRobertz (commented):
## comment-analyzer review (dispatched by godfrey)

**Summary**: Reviewed all comments and JSDoc in `scripts/deleteTemplateProducts.js` (242 lines) and `tests/deleteTemplateProducts.test.js` (223 lines) for factual accuracy, completeness, and long-term maintainability.

---

### Critical Issues

**`scripts/deleteTemplateProducts.js:3`** — Module-level JSDoc contains a misleading parenthetical:
> "identifies template products (those NOT in our Carolina Futons catalog)"

This implies all non-CF products are classified as template and deleted. The actual logic has **three-way classification**: CF (keep), template-pattern-match (delete), and unknown (keep). The parenthetical directly contradicts the script's conservative safety design. A future maintainer reading only the header comment could believe unknown products get deleted.

**Suggestion**: Replace with:
> "identifies template products by matching against known Furniture Store template patterns, and deletes only those. Unknown products are conservatively kept."

---

**`scripts/deleteTemplateProducts.js:125-128`** — `classifyProducts` JSDoc says:
> "Identify template products vs CF products"
> `@returns {{ template: Array, cf: Array }}`

This describes binary classification, but the function implements **three-way** logic where unknown products are silently routed into `cf`. The return type hides the "unknown" category. A maintainer may not realize that `cf` contains both confirmed CF products AND unrecognized products, which matters if the array is used downstream for anything beyond "do not delete."

**Suggestion**: Document the three-way logic in the JSDoc body and note that `cf` includes unknowns. Consider whether the return should include an `unknown` array for transparency.

---

### Improvement Opportunities

**`scripts/deleteTemplateProducts.js:82`** — `@param {string} apiKey - Wix API key` describes the parameter as an "API key" but the environment variable is `WIX_BACKEND_KEY` and the usage example shows `IST.xxx` (Instance Security Token). Calling it an "API key" could confuse someone debugging authentication failures, since Wix distinguishes between API keys and IST tokens.

**Suggestion**: `@param {string} apiKey - Wix backend IST (Instance Security Token)`

---

**`scripts/deleteTemplateProducts.js:109-114`** — `deleteProduct` JSDoc has no `@returns` or `@throws` documentation. The function throws on non-2xx responses, which is important for callers to know (the main loop relies on this for its try/catch error handling).

**Suggestion**: Add `@returns {Promise<void>}` and `@throws {Error} If the API returns a non-OK status`.

---

**`tests/deleteTemplateProducts.test.js:10-12`** — Comment accurately states the test reimplements classification logic from "the same source patterns," but this is a maintainability concern worth flagging: if the source script's patterns or logic change, the test copy must be updated independently. The comment should acknowledge this coupling risk.

**Suggestion**: Add a note like: "If TEMPLATE_PRODUCT_PATTERNS or classifyProducts logic changes in the script, these must be updated to match. Consider extracting shared logic into an importable module."

---

### Recommended Removals

None. All comments serve a purpose in this script.

---

### Positive Findings

- **`scripts/deleteTemplateProducts.js:29`**: "Products with these names are OURS — do NOT delete them" is a clear, actionable directive that earns its place.
- **`scripts/deleteTemplateProducts.js:66-67`**: Template provenance comment citing "Furniture Store template #3563" provides useful audit trail context.
- **`scripts/deleteTemplateProducts.js:42`**: The `catch` comment "File may not exist — that's ok, we'll be conservative" explains the silent-catch rationale, preventing a future maintainer from adding error logging unnecessarily.
- Conservative-by-default classification philosophy is well-commented throughout.

---

**Verdict**: Two critical comment accuracy issues around the three-way classification being described as binary. The module header actively misleads about deletion scope. The remaining comments are solid. Fix the classification documentation to match the actual three-way logic before merge.

DreadPirateRobertz (commented):
## PM Review — Melania (FINAL ARBITER)

**Verdict: CHANGES REQUESTED**

All 5 reviewers flagged the same critical issues. This is a destructive script — quality bar is higher.

### Must fix before merge:

1. **Tests duplicate source logic (ALL reviewers, CRITICAL)** — Tests copy-paste `classifyProducts`, `loadCFProductNames`, and `TEMPLATE_PRODUCT_PATTERNS` instead of importing them. This means tests validate a snapshot copy, NOT the actual code. Refactor: export functions from script, import in tests.

2. **Silent error swallowing in `loadCFProductNames` (3 reviewers, CRITICAL)** — Empty `catch (e) {}` hides JSON parse errors, permission errors, etc. Only catch `ENOENT`, rethrow everything else.

3. **Exit code 0 on delete failures (2 reviewers, HIGH)** — If every delete fails (bad API key, wrong site ID), script exits successfully. Add `process.exit(2)` when `failed > 0`.

4. **No empty-result sanity check (1 reviewer, HIGH)** — If API returns 0 products (wrong site ID, auth issue), script reports "nothing to delete" and exits 0. Add a warning when `allProducts.length === 0`.

5. **JSDoc says binary classification, logic is three-way (1 reviewer, CRITICAL)** — Module header and `classifyProducts` docs imply template = "not CF". Actual logic: CF (keep) / template (delete) / unknown (keep). Fix docs to match reality.

### Nice to have (not blocking):
- Rate limiting between deletes
- `@throws` on `deleteProduct` JSDoc
- Audit log of deleted product IDs

Good work on the conservative defaults and dry-run flag. Fix the 5 items above and re-request review.

@godfrey — please address these and push updated commits.

DreadPirateRobertz (commented):
**PM Review — melania** ✅ LGTM

Clean implementation. Conservative deletion logic (unknown products kept, CF match takes priority). 17 tests, CI green. Run `--dry-run` first before actual deletion.

PR #303, opened 2026-03-14
fix(cf-4rrx): replace template stock images on homepage category cards

DreadPirateRobertz:
## Summary
- Replace template stock photos (blue couch, purple chair) on the 4 static homepage category card boxes with real CF product images from Wix CDN
- Sets images via both child image elements (#image26/24/22/20) and box background fallback
- Adds proper alt text via `getCategoryCardAlt()` for accessibility
- Updates home-page-mapping.json to document template card image element IDs

## Context
Template has 4 static Container boxes (box26/24/22/20) instead of a Repeater. The existing `#categoryRepeater` code path silently fails, leaving template stock images visible. This fix adds a parallel code path that directly targets the static template card elements.

**Note:** Image element IDs (#image26 etc.) are guessed from Wix convention — may need adjustment after verifying in Wix Studio editor.

## Test plan
- [x] 2 new tests: verify template card images use real wixstatic.com URLs with 600x400 dimensions
- [x] 2 new tests: verify alt text is set on template card images
- [x] All 12,437 existing tests pass
- [ ] Visual verification in Wix Studio staging needed (image element IDs may vary)

### DIFF
--- a/tests/homePage.test.js
+++ b/tests/homePage.test.js
@@ -257,6 +257,28 @@ describe('Home Page', () => {
       );
       expect(withClicks).toHaveLength(6);
     });
+
+    it('sets real CF product images on template category card boxes', async () => {
+      await onReadyHandler();
+      // Template boxes should get CF product images (not stock template photos)
+      const templateImgIds = ['#image26', '#image24', '#image22', '#image20'];
+      templateImgIds.forEach(id => {
+        const img = getEl(id);
+        // Images should be set to wixstatic.com URLs from placeholderImages
+        expect(img.src).toContain('static.wixstatic.com');
+        expect(img.src).toContain('w_600,h_400');
+      });
+    });
+
+    it('sets alt text on template category card images', async () => {
+      await onReadyHandler();
+      const templateImgIds = ['#image26', '#image24', '#image22', '#image20'];
+      templateImgIds.forEach(id => {
+        const img = getEl(id);
+        expect(img.alt).toBeTruthy();
+        expect(img.alt.length).toBeGreaterThan(5);
+      });
+    });
   });
 
   // ── Hero Animation ────────────────────────────────────────────────


DreadPirateRobertz (commented):
**Review by: code-simplifier agent (dispatched by radahn)**

---

Overall this PR is clean and well-scoped. A few simplification opportunities:

### 1. Repeated `getCategoryCardImage()` call (Home.js)

`getCategoryCardImage(cat.collection)` is called twice per iteration — once for the `` element and once for the box background. Extract to a local variable:

```js
const imgUrl = getCategoryCardImage(cat.collection);
const altText = getCategoryCardAlt(cat.collection);
```

Then reference `imgUrl` in both the `img.src` and `box.background.src` assignments. This avoids the duplicate lookup and makes it clearer that the same URL is intentionally used in both places.

### 2. Three separate try/catch blocks could be consolidated

The three sequential try/catch blocks (image, background, text) each swallow errors silently. Since all three operations target the same card and are independent, this is functionally correct, but the pattern is verbose. Consider consolidating into a single try/catch per card:

```js
categoriesWithCounts.slice(0, 4).forEach((cat, i) => {
  const card = templateCardMap[i];
  if (!card || !cat.collection) return;
  const imgUrl = getCategoryCardImage(cat.collection);
  const altText = getCategoryCardAlt(cat.collection);
  try {
    const img = $w(card.imgId);
    if (img) { img.src = imgUrl; img.alt = altText; }
    const box = $w(card.boxId);
    if (box && box.background) { box.background.src = imgUrl; }
    const txt = $w(card.textId);
    if (txt) txt.text = cat.name;
  } catch (e) {}
});
```

**Counter-argument**: If `$w(card.imgId)` throws, the separate try/catch blocks ensure the box background and text still get set. In the Wix Velo runtime, `$w()` for a non-existent element does throw, so the current multi-block approach is actually the safer pattern given the "guessed" element IDs. If you keep the separate blocks, add a brief comment explaining why (e.g., `// Separate try/catch: each element ID is guessed and may not exist`).

### 3. Guard `!card` is unreachable

`if (!card || !cat.collection) return;` — the `!card` guard can never be true because `templateCardMap` has exactly 4 entries and `categoriesWithCounts.slice(0, 4)` guarantees `i` is 0-3. Not harmful, but the unnecessary guard adds mild cognitive overhead. Consider removing it or leaving a comment that it is purely defensive.

### 4. Test assertions could use `.each` pattern

The two new tests repeat the same structure — iterating over `templateImgIds` and asserting per element. Vitest supports `it.each` which would make the test intent slightly cleaner. However, this is a minor style point and the current approach is perfectly readable. No strong recommendation here.

### 5. Mapping JSON: "_note" field

The `_note` field in `_templateCategoryCards` is a good documentation choice. No change needed.

---

**Verdict**: The PR is sound. The main actionable item is extracting the duplicated `getCategoryCardImage()` call into a local variable (item 1). The rest are minor observations. No functional issues found.

PR #309, opened 2026-03-14
security(hq-j11u): migrate cron secrets from query strings to headers

DreadPirateRobertz:
## Summary
- **P0 security fix**: 6 cron endpoints were passing `ALERT_CRON_KEY` via `?key=` query parameter
- Query strings get logged in access logs, cached by proxies, visible in browser history
- Migrated all 6 to `X-Cron-Secret` header, matching existing pattern (FB audience, Klaviyo)
- 6 new tests verify header-based auth, vulnerability test now asserts fix
- 12,451 tests passing (net +6 new security tests)

## Affected Endpoints
- `checkWishlistAlerts`
- `triggerBrowseRecoveryCron`
- `triggerCartRecoveryCron`
- `processEmailQueueCron`
- `triggerReengagementCron`
- `processPostPurchaseCareCron`

## Test plan
- [x] All 6 endpoints read from `request.headers['x-cron-secret']` (not `request.query`)
- [x] `timingSafeEqual` still used for auth comparison
- [x] 403 returned for missing/invalid header
- [x] 200 returned for valid header
- [x] No remaining `request.query?.key` references in source
- [x] Full suite: 12,451 tests green

### DIFF
--- a/tests/httpFunctions.test.js
+++ b/tests/httpFunctions.test.js
@@ -13,6 +13,7 @@ import {
   get_triggerCartRecoveryCron,
   get_processEmailQueueCron,
   get_triggerReengagementCron,
+  get_processPostPurchaseCareCron,
 } from '../src/backend/http-functions.js';
 
 const sampleProducts = [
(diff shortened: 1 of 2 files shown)


DreadPirateRobertz:

The query-to-header migration is clean and mechanical — no issues with the change itself. Tests updated correctly and the security test now properly asserts the fix instead of documenting a known vulnerability.

**One simplification opportunity (pre-existing, not a blocker):**

All 6 cron endpoints repeat an identical auth block:

```js
const { getSecret } = await import('wix-secrets-backend');
const cronKey = await getSecret('ALERT_CRON_KEY');
const requestKey = request.headers['x-cron-secret'];

if (!cronKey || !requestKey || !timingSafeEqual(requestKey, cronKey)) {
  return forbidden({ ... });
}
```

This could be extracted into a shared helper (e.g., `verifyCronAuth(request)`) that returns `{ authorized: true }` or the 403 response. That would reduce ~30 duplicated lines across the 6 endpoints and make future auth changes (like this one) a single-line fix. Worth a follow-up bead if not already tracked.

**Verdict: Approve.** The PR does exactly what it says, tests are solid, no unnecessary complexity introduced.

PR #310, opened 2026-03-14
security(hq-khcp): rate limiting on lookupReturn and submitGuestReturn

DreadPirateRobertz:
## Summary
- **P0 security fix**: Guest return endpoints had no rate limiting, enabling order enumeration
- Sliding window limiter: max 5 attempts per email per 60 seconds
- Rate check placed after input validation to avoid wasting budget on invalid requests
- In-memory implementation with automatic cleanup at >1000 entries

## Test plan
- [x] `_checkRateLimit` allows exactly RATE_LIMIT_MAX calls then blocks
- [x] Different identifiers tracked independently
- [x] `lookupReturn` returns "Too many attempts" after 5 calls
- [x] `submitGuestReturn` returns "Too many attempts" after 5 calls
- [x] Different emails are not blocked by each other's limits
- [x] Invalid input fails validation before consuming rate limit budget
- [x] Source-level tests verify `_checkRateLimit` present in both endpoints
- [x] Existing tests still pass (rate limit map cleared in beforeEach)
- [x] Full suite: 12,453 tests green

### DIFF
--- a/tests/securityPiiGdpr.test.js
+++ b/tests/securityPiiGdpr.test.js
@@ -258,7 +258,7 @@ describe('Security: submitGuestReturn input validation', () => {
   it('sanitizes order number', () => {
     const block = src.substring(
       src.indexOf('export const submitGuestReturn'),
-      src.indexOf('export const submitGuestReturn') + 1000
+      src.indexOf('export const submitGuestReturn') + 1500
     );
     expect(block).toMatch(/sanitize\(data\.orderNumber/);
   });
@@ -266,7 +266,7 @@ describe('Security: submitGuestReturn input validation', () => {
   it('validates email format', () => {
     const block = src.substring(
       src.indexOf('export const submitGuestReturn'),
-      src.indexOf('export const submitGuestReturn') + 1000
+      src.indexOf('export const submitGuestReturn') + 1500
     );
     expect(block).toMatch(/validateEmail\(cleanEmail\)/);
   });
@@ -274,7 +274,7 @@ describe('Security: submitGuestReturn input validation', () => {
   it('validates return reason against allowlist', () => {
     const block = src.substring(
       src.indexOf('export const submitGuestReturn'),
-      src.indexOf('export const submitGuestReturn') + 1000
+      src.indexOf('export const submitGuestReturn') + 1500
     );
     expect(block).toMatch(/VALID_REASONS/);
   });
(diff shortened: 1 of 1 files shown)


DreadPirateRobertz:

Reviewed the diff for PR #310. This is clean, well-structured security work. A few simplification suggestions:

---

### 1. `_checkRateLimit` — simplify the entry initialization

The `let entry` / `if (!entry)` pattern can be collapsed. The `|| 'unknown'` fallback for `key` is defensive but the callers always pass `cleanEmail` which is already validated. Minor, but:

```js
// Current
const key = identifier || 'unknown';
let entry = _rateLimitMap.get(key);
if (!entry) {
  entry = { timestamps: [] };
  _rateLimitMap.set(key, entry);
}

// Simpler
const key = identifier || 'unknown';
if (!_rateLimitMap.has(key)) {
  _rateLimitMap.set(key, { timestamps: [] });
}
const entry = _rateLimitMap.get(key);
```

Both are fine — this is a nit, not a blocker.

### 2. Cleanup comment says "~100 calls" but threshold is 1000

The inline comment says `Periodic cleanup: remove stale IPs every ~100 calls` but the actual guard is `_rateLimitMap.size > 1000`. Suggest updating the comment to match:

```js
// Periodic cleanup: evict stale entries when map exceeds 1000 keys
```

### 3. Source-level tests in `securityPiiGdpr.test.js` — repeated substring extraction

The pattern `src.substring(src.indexOf('export const submitGuestReturn'), src.indexOf('export const submitGuestReturn') + 1000)` is repeated five times. A small helper at the top of the describe block would reduce duplication and make the magic number `1000` a single change point:

```js
function extractBlock(src, marker, len = 1500) {
  const idx = src.indexOf(marker);
  return src.substring(idx, idx + len);
}
```

The `lookupReturn` security tests already had the same pattern before this PR, so this is pre-existing, but since the PR touched these lines it would be a clean improvement.

### 4. No issues found with the core logic

- Rate check correctly placed after validation (no budget waste on bad input) — good.
- Sliding window with eviction is appropriate for in-memory limiting.
- Test coverage is thorough: unit tests on `_checkRateLimit`, integration tests on both endpoints, independence test, and validation-before-rate-limit test.

**Verdict: No blockers. Ship it.** The comment typo (item 2) is worth a quick fix; items 1 and 3 are optional polish.

PR #313, opened 2026-03-14
feat(cf-05zf): add Getting It Home page — delivery & assembly info

DreadPirateRobertz:
## Summary
- New "Getting It Home" page with delivery and assembly service information
- 4 service tiers: DIY (free), Home Drop Off, In-Store Assembly ($40), White Glove ($60+)
- Content sourced from production site via Playwright
- `deliveryHelpers.js` module with content data, `Getting It Home.js` page code

## TDD approach
- Tests written first, verified red, then implementation to green
- 15 unit tests for content helpers (getIntroText, getServiceTiers, getDeliveryRates)
- 13 integration tests for page initialization (SEO, tracking, repeater, nav links)

## Test plan
- [x] 28 new tests pass
- [x] Full suite: 12,492 tests pass
- [ ] Visual verification in Wix Studio needed (element IDs need mapping)

### DIFF
--- a/src/pages/Getting It Home.js
+++ b/src/pages/Getting It Home.js
@@ -1,0 +1,75 @@
+// Getting It Home.js - Assembly & Delivery Information
+// Service tiers, delivery rates, and assembly options
+import { initPageSeo } from 'public/pageSeo.js';
+import { trackEvent } from 'public/engagementTracker';
+import { initBackToTop } from 'public/mobileHelpers';
+import { makeClickable } from 'public/a11yHelpers.js';
+import { getIntroText, getServiceTiers, getDeliveryRates } from 'public/deliveryHelpers.js';
+
+$w.onReady(async function () {
+  initBackToTop($w);
+  initIntro();
+  initServiceTiers();
+  initDeliveryRates();
+  initNavLinks();
+  initPageSeo('getting-it-home');
+  trackEvent('page_view', { page: 'getting-it-home' });
+});
+
+// ── Intro Section ───────────────────────────────────────────────────
+
+function initIntro() {
+  try {
+    const intro = $w('#deliveryIntro');
+    if (intro) intro.text = getIntroText();
+  } catch (e) { /* Element may not exist in template */ }
+}
+
+// ── Service Tiers ───────────────────────────────────────────────────
+
+function initServiceTiers() {
+  try {
+    const repeater = $w('#serviceTierRepeater');
+    if (!repeater) return;
+
+    try { repeater.accessibility.ariaLabel = 'Delivery and assembly service options'; } catch (e) {}
+    repeater.onItemReady(($item, itemData) => {
+      try { $item('#tierTitle').text = itemData.title; } catch (e) {}
+      try { $item('#tierPrice').text = itemData.price; } catch (e) {}
+      try { $item('#tierDescription').text = itemData.description; } catch (e) {}
+    });
+    repeater.data = getServiceTiers();
+  } catch (e) { /* Repeater may not exist in template */ }
+}
+
+// ── Delivery Rates ──────────────────────────────────────────────────
+
+function initDeliveryRates() {
+  try {
+    const rates = getDeliveryRates();
+    try {
+      const chargeEl = $w('#deliveryMinCharge');
+      if (chargeEl) chargeEl.text = `Minimum local charge (up to approx. ${rates.minimumRadius} radius from our store) ${rates.minimumCharge}`;
+    } catch (e) {}
+    try {
+      const noteEl = $w('#deliveryRateNote');
+      if (noteEl) noteEl.text = rates.note;
+    } catch (e) {}
+  } catch (e) { /* Rate elements may not exist */ }
+}
+
+// ── Navigation Links ────────────────────────────────────────────────
+
+function initNavLinks() {
+  try {
+    makeClickable($w('#deliveryFaqLink'), () => {
+      import('wix-location-frontend').then(({ to }) => to('/faq'));
+    }, { ariaLabel: 'View assembly videos and FAQs' });
+  } catch (e) {}
+
+  try {
+    makeClickable($w('#deliveryContactLink'), () => {
+      import('wix-location-frontend').then(({ to }) => to('/contact'));
+    }, { ariaLabel: 'Contact us about delivery rates' });
+  } catch (e) {}
+}
--- a/src/public/deliveryHelpers.js
+++ b/src/public/deliveryHelpers.js
@@ -1,0 +1,67 @@
+/**
+ * deliveryHelpers.js — Content data for the Getting It Home page.
+ * Service tiers, delivery rates, and assembly info for Carolina Futons.
+ */
+
+const SERVICE_TIERS = [
+  {
+    _id: 'diy',
+    title: 'Do It Yourself',
+    price: 'Free',
+    description: 'Are you an "I can do it myself" kind of person? Then you\'ll save the most by taking your purchase home and assembling it yourself with included box instructions or viewing the instructional videos online, available on our FAQ page. Most futon frames are simple to build and will fit in a small SUV or truck. You can also pick up your mattress in its original box or we can unbox for you to carry out (tied up vertically and wrapped in plastic). We will help you carry out and load your purchases into your vehicle.',
+    icon: 'wrench',
+  },
+  {
+    _id: 'dropoff',
+    title: 'Home Drop Off',
+    price: 'Delivery rate',
+    description: 'If you want to assemble your purchase yourself but don\'t have a suitable vehicle for transporting, we can deliver your items to your home. We can also deliver your mattress and take away your old mattress for a $10 landfill fee. Delivery rates are based on mileage from our store.',
+    icon: 'truck',
+  },
+  {
+    _id: 'instore',
+    title: 'In-Store Assembly',
+    price: '$40.00',
+    description: 'Another option is to have us assemble your frame in our store for you to pick up and take home. You can also pick up your mattress in its original box or we can unbox for you to carry out (tied up vertically and wrapped in plastic). We will help you carry out and load your purchases into your vehicle. A full-sized pickup truck or van is recommended for transporting an assembled frame and/or mattress. Our fee for this service is $40.00, with a 24-hour notice required.',
+    icon: 'build',
+  },
+  {
+    _id: 'whiteglove',
+    title: 'Premium White Glove Service',
+    price: '$60.00 + delivery',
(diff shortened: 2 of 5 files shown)


DreadPirateRobertz:
## PR Test Coverage Analysis — pr-test-analyzer (dispatched by miquella, tech lead)

### Summary

PR #313 adds a "Getting It Home" page with delivery/assembly service tier content and 28 tests across two test files. The coverage is solid overall — the helper module has thorough content validation tests, and the page-level integration tests verify initialization wiring. A few gaps worth noting below.

---

### Critical Gaps (none)

No critical (8-10) test gaps identified. The core content data and page initialization paths are well-covered.

---

### Important Improvements

**1. Navigation link click handlers are wired but not tested for destination** (Criticality: 6/10)
- `gettingItHome.test.js` verifies that `onClick` was called on `#deliveryFaqLink` and `#deliveryContactLink`, but does not test that clicking the handler actually calls `to('/faq')` or `to('/contact')`. The dynamic `import('wix-location-frontend')` path is completely untested.
- **What it catches**: If someone changes the route from `/faq` to `/faqs` or introduces a typo, no test would fail. Navigation is user-facing and fairly important.
- **Suggested approach**: Mock `wix-location-frontend` and invoke the captured `onClick` callback, then assert `to` was called with the correct path.

**2. `getServiceTiers` returns the same array reference (no defensive copy)** (Criticality: 5/10)
- `deliveryHelpers.js` returns `SERVICE_TIERS` directly — a consumer could mutate the array (`tiers.push(...)` or `tiers[0].price = 'hacked'`) and corrupt the module-level constant for all subsequent calls.
- **What it catches**: If the page is ever hot-reloaded or `getServiceTiers()` is called from multiple places in one session, mutation bugs could surface. A test that calls `getServiceTiers()` twice and verifies the second call is unaffected by mutations to the first result would expose this.
- Same applies to `getDeliveryRates()` returning the mutable `DELIVERY_RATES` object.
- This is a "nice to have" — low risk in Wix Velo's page lifecycle where modules reinitialize per navigation, but worth a note.

**3. `initServiceTiers` repeater with `null`/missing element** (Criticality: 5/10)
- The page code has `if (!repeater) return;` guard, but the test mock always provides a valid element for every selector (via the `getEl` factory). There is no test verifying graceful degradation when the repeater element does not exist.
- **What it catches**: If the guard is accidentally removed or the try/catch is refactored, no test would detect a crash on templates where the repeater is absent.
- **Suggested approach**: One test that sets `elements.set('#serviceTierRepeater', null)` before calling `onReadyHandler()` and asserts it does not throw.

**4. `initDeliveryRates` composite text format** (Criticality: 4/10)
- The test checks that `#deliveryMinCharge` text contains `'$25'` but does not verify the full template string format: `"Minimum local charge (up to approx. 10-mile radius from our store) $25.00"`. If the string interpolation breaks (e.g., template literal variables are swapped), the test could still pass because `$25` appears somewhere.
- Low risk since the template is simple, but a more specific assertion on the composed format would be more robust.

---

### Test Quality Issues

**Tests tightly coupled to content strings** (Minor concern)
- Several `deliveryHelpers.test.js` tests assert exact content like `"don't add the cost of assembly and delivery"` and specific tier titles/prices. These tests will break if marketing copy changes, even though the code is functioning correctly. This is acceptable for a content-data module (the content IS the contract), but worth noting that content changes will require test updates. No action needed — just awareness.

**Mock `deliveryHelpers.js` uses `importActual` passthrough** (Observation)
- In `gettingItHome.test.js`, the mock of `public/deliveryHelpers.js` just re-exports the actual module. This means page-level tests are implicitly integration tests (testing real content data + page wiring together). This is fine — it provides more realistic coverage. But it also means a content data change would break tests in both files. Consider whether the page test should use stub data if content churn becomes an issue.

---

### Positive Observations

- **Good structural validation**: Testing that all tiers have required fields (`_id`, `title`, `price`, `description`, `icon`) and unique IDs is exactly the right pattern for repeater data. Would catch schema drift.
- **ARIA coverage**: Testing that the repeater gets an accessible label is a good practice, especially given prior WCAG work on this codebase.
- **SEO and tracking init verified**: Confirming `initPageSeo('getting-it-home')` and `trackEvent('page_view', ...)` are called with correct arguments catches common copy-paste errors from other pages.
- **28 tests for 142 lines of production code** is a healthy ratio for a content-driven page.
- **TDD approach documented in PR body** — tests written before implementation. Good discipline.

---

**Overall verdict**: Good coverage for a content/page-init PR. The navigation destination testing gap (item 1) is the most actionable improvement. No blockers from a test coverage perspective.

DreadPirateRobertz:
## Silent Failure Hunter Review

**Agent**: silent-failure-hunter (miquella, tech lead)

---

### Verdict: SIGNIFICANT ISSUES FOUND

This PR introduces `Getting It Home.js` with **14 empty catch blocks** across 75 lines of page code. Nearly every operation is wrapped in a silent try-catch. This is the single worst pattern for debuggability: when something breaks in production, there will be zero indication of what went wrong, where, or why. Users will see a blank or partially-rendered page with no feedback.

---

### CRITICAL — 14 Empty Catch Blocks (Silent Failures)

Every function in `Getting It Home.js` contains empty catch blocks that swallow all errors without logging, user feedback, or propagation.

**File: `src/pages/Getting It Home.js`**

| Lines | Location | What Is Silently Swallowed |
|-------|----------|--------------------------|
| L25 | `initIntro()` catch | `getIntroText()` runtime errors, `$w()` selector errors, property assignment failures |
| L35 | `initServiceTiers()` aria catch | ARIA accessibility setup failures — accessibility silently degrades |
| L37 | `onItemReady` `#tierTitle` catch | Title rendering failure — user sees blank tier cards with no explanation |
| L38 | `onItemReady` `#tierPrice` catch | Price rendering failure — user sees no pricing information |
| L39 | `onItemReady` `#tierDescription` catch | Description rendering failure — blank content with no indication |
| L41 | `initServiceTiers()` outer catch | Entire repeater setup failure — the core content of the page vanishes silently |
| L49-50 | `initDeliveryRates()` `#deliveryMinCharge` catch | Delivery charge info disappears without trace |
| L53-54 | `initDeliveryRates()` `#deliveryRateNote` catch | Rate note disappears without trace |
| L55 | `initDeliveryRates()` outer catch | `getDeliveryRates()` failure — entire rates section gone, no feedback |
| L62 | `initNavLinks()` FAQ link catch | FAQ navigation link silently broken — user clicks nothing, nothing happens |
| L68 | `initNavLinks()` contact link catch | Contact navigation link silently broken — same problem |

**Why this is critical:**

1. **Debugging nightmare**: When a user reports "the delivery page is blank," there will be zero logs, zero Sentry alerts, zero breadcrumbs. The developer must guess which of the 14 catch blocks ate the error.

2. **Broad catch scope**: Each catch block catches ALL exception types. A `TypeError` from a typo in a property name, a `ReferenceError` from a missing import, a network error from a failed dynamic import — all silently swallowed identically.

3. **Cascading invisible failures**: If `getServiceTiers()` throws (e.g., the module is refactored and the export is renamed), the entire service tier section vanishes. The user sees a page with missing content and no error message. The developer gets no alert.

4. **Comments are not error handling**: Comments like `/* Element may not exist in template */` explain the *intent* but do not *handle* the error. If the element truly may not exist, check for its existence explicitly rather than catching all possible exceptions.

**Hidden errors these catch blocks will mask:**
- `TypeError` from accessing properties on undefined (e.g., misspelled element IDs)
- `ReferenceError` from missing imports after refactoring
- `SyntaxError` from malformed template literals
- Runtime errors in `getIntroText()`, `getServiceTiers()`, `getDeliveryRates()` if deliveryHelpers.js is modified
- `wix-location-frontend` dynamic import failures (network issues, module not available)
- Navigation errors when `to()` is called with invalid routes

**Recommended fix pattern:**

```javascript
// INSTEAD OF:
function initIntro() {
  try {
    const intro = $w('#deliveryIntro');
    if (intro) intro.text = getIntroText();
  } catch (e) { /* Element may not exist in template */ }
}

// DO THIS:
function initIntro() {
  const intro = $w('#deliveryIntro');
  if (!intro) return; // Explicit null check — no catch needed
  try {
    intro.text = getIntroText();
  } catch (e) {
    console.error('[Getting It Home] Failed to set intro text:', e);
    // Or use project logging: logError(e, { errorId: 'delivery-intro-init', page: 'getting-it-home' });
  }
}
```

For the repeater `onItemReady` callbacks, the individual try-catches around each property assignment are especially problematic — if `#tierTitle` fails due to a real bug, the user sees a card with a title missing but price and description present, which is confusing and hard to diagnose. Either let all three fail together (so the card is clearly broken) or log each failure individually.

For the navigation links with dynamic `import('wix-location-frontend')`:

```javascript
// INSTEAD OF:
try {
  makeClickable($w('#deliveryFaqLink'), () => {
    import('wix-location-frontend').then(({ to }) => to('/faq'));
  }, { ariaLabel: 'View assembly videos and FAQs' });
} catch (e) {}

// DO THIS — and handle the promise rejection too:
const faqLink = $w('#deliveryFaqLink');
if (faqLink) {
  makeClickable(faqLink, () => {
    import('wix-location-frontend')
      .then(({ to }) => to('/faq'))
      .catch(e => console.error('[Getting It Home] FAQ navigation failed:', e));
  }, { ariaLabel: 'View assembly videos and FAQs' });
}
```

Note that the `.then()` chain on the dynamic import has **no `.catch()`**, meaning promise rejections from `import('wix-location-frontend')` are unhandled even if the outer try-catch were removed. This is a separate silent failure path.

---

### HIGH — No Error Boundary Around Top-Level `$w.onReady`

**File: `src/pages/Getting It Home.js`, Lines 9-17**

The `$w.onReady` callback calls 6 functions sequentially with no error handling. If `initBackToTop($w)` throws, none of the subsequent functions (`initIntro`, `initServiceTiers`, etc.) execute. The page would be completely non-functional, and because each inner function has its own silent catch, the actual failure point would be obscured.

While the Wix runtime may catch top-level errors, the page should have a top-level guard that at minimum logs the failure so it can be tracked.

---

### MEDIUM — `deliveryHelpers.js` Returns Mutable References

**File: `src/public/deliveryHelpers.js`, Lines 53-66**

`getServiceTiers()` returns the `SERVICE_TIERS` array directly. If any consumer mutates this array (e.g., `tiers.push(...)` or `tiers[0].title = 'Modified'`), it will affect all subsequent calls. This is not an error handling issue per se, but it can cause confusing bugs that are hard to trace — a form of silent data corruption.

Consider returning `structuredClone(SERVICE_TIERS)` or `Object.freeze()` the data.

---

### Summary

| Severity | Count | Description |
|----------|-------|-------------|
| CRITICAL | 14 | Empty catch blocks silently swallowing all error types |
| HIGH | 1 | No top-level error boundary in `$w.onReady` |
| HIGH | 2 | Unhandled promise rejections in dynamic `import().then()` chains |
| MEDIUM | 1 | Mutable data returned from helper functions |

The `deliveryHelpers.js` module and its tests are clean — no issues there. The problem is entirely in the page code's error handling philosophy of "catch everything, report nothing."

This page will work perfectly when everything goes right. The moment anything goes wrong — a renamed element ID, a refactored helper, a missing Wix module — it will fail silently and be extremely difficult to debug.

PR #328, opened 2026-03-14
test(batchAltText): add 19 tests for batch alt text module

DreadPirateRobertz:
## Summary
- Adds 19 tests for `src/backend/batchAltText.web.js` covering brand/category detection, alt text generation (position labels, URL context, truncation), batch update modes (dry-run, force), product preview, and error handling
- All 12,855 tests pass (331 files)

## Test plan
- [x] `npx vitest run tests/batchAltText.test.js` — 19/19 pass
- [x] Full suite — 12,855 pass

### DIFF
--- a/tests/batchAltText.test.js
+++ b/tests/batchAltText.test.js
@@ -1,16 +1,9 @@
 /**
  * Tests for backend/batchAltText.web.js
  * Covers: brand detection, category detection, alt text generation,
- * batch update with dry-run/force, product preview, pagination.
+ * batch update with dry-run/force, product preview, error handling.
  */
 import { describe, it, expect, vi, beforeEach } from 'vitest';
-
-// ── Mock wix-web-module ─────────────────────────────────────────────
-
-vi.mock('wix-web-module', () => ({
-  Permissions: { Anyone: 'anyone', Admin: 'admin' },
-  webMethod: (_perm, fn) => fn,
-}));
 
 // ── Mock wix-stores-backend ─────────────────────────────────────────
 
@@ -31,6 +24,7 @@ vi.mock('wix-stores-backend', () => ({
   },
 }));
 
+import { products as storeMock } from 'wix-stores-backend';
 import {
   runBatchAltTextUpdate,
   previewProductAltText,
@@ -57,6 +51,7 @@ describe('batchAltText — runBatchAltTextUpdate', () => {
   beforeEach(() => {
     mockProducts.length = 0;
     mockUpdateCalls.length = 0;
+    vi.clearAllMocks();
   });
 
   it('updates products with missing alt text', async () => {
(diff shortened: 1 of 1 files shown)


DreadPirateRobertz (commented):
## Code Review: `tests/batchAltText.test.js`

**Reviewer**: code-reviewer agent, dispatched by radahn

---

### Critical (90-100)

*None found.*

### Important (80-89)

**1. Redundant inline `wix-web-module` mock overrides the shared mock (Confidence: 85)**

`tests/batchAltText.test.js`, lines 10-13

The project has a shared `wix-web-module` mock at `tests/__mocks__/wix-web-module.js` with an alias in `vitest.config.js`. The inline `vi.mock('wix-web-module', ...)` overrides this shared mock unnecessarily. Worse, the inline version uses lowercase permission values (`'anyone'`, `'admin'`) while the shared mock uses the canonical values (`'Anyone'`, `'Admin'`). This inconsistency could cause confusion if any future test compares permission values.

**Fix**: Remove the inline `vi.mock('wix-web-module', ...)` block entirely (lines 10-13). The vitest alias will resolve to the shared mock automatically, which already strips the `webMethod` permission wrapper the same way.

**2. Mock function state not cleared between tests (Confidence: 82)**

`tests/batchAltText.test.js`, lines 57-60

The `beforeEach` blocks clear the `mockProducts` and `mockUpdateCalls` arrays, but the `vi.fn()` instances on `queryProducts`, `updateProductFields`, and `getProduct` are never cleared (no `vi.clearAllMocks()` or `vi.restoreAllMocks()`). This means call counts, call arguments, and any `mockRejectedValueOnce`/`mockResolvedValueOnce` state can leak between tests if they don't fully consume their one-shot overrides.

Today this works because the tests happen to consume their one-shot mocks, but it is fragile. If a test is skipped or reordered, a stale `mockRejectedValueOnce` could cause an unrelated test to fail.

**Fix**: Add `vi.clearAllMocks()` to each `beforeEach` block, or add a single global `beforeEach` at the top of the file:
```js
beforeEach(() => {
  vi.clearAllMocks();
  mockProducts.length = 0;
  mockUpdateCalls.length = 0;
});
```

---

### Summary

The test file is well-structured with good coverage of the core functionality: brand/category detection, position labels, URL-based context, truncation, dry-run mode, force mode, and error handling. The `makeProduct` factory and test organization are clean.

Two items need attention: (1) the inline `wix-web-module` mock should be removed in favor of the project's shared mock, and (2) `vi.clearAllMocks()` should be added to `beforeEach` to prevent inter-test state leakage. Both are straightforward fixes.

DreadPirateRobertz (commented):
## 🕵️ Silent Failure Hunter (dispatched by miquella)

Scanning for patterns where failures could be swallowed, assertions could pass vacuously, or errors could go undetected.

### Findings

1. **Mock chain always returns same data regardless of pagination params** (LOW risk): The `queryProducts` mock returns `{ items: mockProducts }` regardless of what `skip()` or `limit()` values are passed. If the source code has a bug in its pagination logic (e.g., wrong skip offset), these tests would not catch it. The mock chain (`skip().limit().find()`) always resolves identically.

2. **Error test relies on `mockRejectedValueOnce`** (OK): Line ~105 uses `mockRejectedValueOnce` which only fails the first call. If the source code retries internally, the second call would succeed silently. Verify the source does NOT retry on `updateProductFields` failure. If it does, this test would pass but the error array would be empty.

3. **`previewProductAltText('nonexistent')` test** (OK but fragile): This relies on `getProduct` returning `null` for an unknown ID (since `mockProducts.find()` returns `undefined`). The test checks `result.error` is truthy. If the source code throws instead of returning an error object, this test would fail loudly (not silently). Good.

4. **No assertion on alt text content correctness for update path**: The "updates products with missing alt text" test (line ~68) checks `result.updated === 1` and `mockUpdateCalls.length === 1` but does NOT assert what alt text was actually written to the product. The `fields` in `mockUpdateCalls[0].fields` is captured but never inspected. A bug that writes empty strings or wrong alt text would pass this test.

5. **Brand dedup test (line ~213)**: Counts occurrences of `"Night & Day"` in the alt string and expects 1. If the source code fails to include the brand at all (0 occurrences), the regex match returns `null`, `|| []` makes it `[]`, and `.length` is 0 — the test would correctly fail. Good.

### Verdict
**One actionable gap**: Test #1 in the first `describe` block should assert the actual alt text content written via `mockUpdateCalls[0].fields`, not just that an update happened. A write of empty/garbage alt text would silently pass. Not a blocker for merge but worth a follow-up.

DreadPirateRobertz (commented):
## ✂️ Code Simplifier (dispatched by miquella)

Reviewing for unnecessary complexity, redundancy, and opportunities to simplify.

### Observations

1. **Mock setup is appropriately minimal**: The `webMethod` mock correctly unwraps to just `fn`, and the stores mock uses clean chainable patterns. No over-engineering here.

2. **`makeProduct()` factory**: Well done — single factory with spread overrides. No redundant builder patterns.

3. **Array.from for 6-image test** (line ~176): `Array.from({ length: 6 }, (_, i) => ...)` is idiomatic and clean. No simplification needed.

4. **Potential dedup in `beforeEach` blocks**: All three `describe` blocks have nearly identical `beforeEach` clearing `mockProducts.length = 0`. The third block (`previewProductAltText`) doesn't clear `mockUpdateCalls` since it doesn't use them — this is correct and intentional, not redundant.

5. **Dynamic re-import in error test** (line ~103): `const { products: store } = await import('wix-stores-backend')` is needed to access the mock instance for `mockRejectedValueOnce`. An alternative would be to hoist the mock reference at module scope:
   ```js
   const mockUpdateFn = vi.fn(...);
   ```
   Then reference `mockUpdateFn.mockRejectedValueOnce(...)` directly. This is marginally simpler but the current approach is clear enough.

### Verdict
**No simplification needed.** The test file is already well-structured with no dead code, no unnecessary abstractions, and good use of test utilities. Clean PR.

DreadPirateRobertz (commented):
## 📋 5-Agent Review Summary (miquella)

All 5 review agents have completed their analysis of PR #328.

### Consensus: APPROVE ✅

All agents agree this is a clean, well-structured test PR. CI is green (12,855 tests pass). No blockers found.

### Key Finding (2 agents flagged)
**No assertion on written alt text content**: Both Silent Failure Hunter and PR Test Analyzer independently identified that the main update test verifies *that* an update happened (`mockUpdateCalls.length === 1`) but not *what* was written. A bug producing empty or malformed alt text would pass the current test. **Recommend adding an assertion on `mockUpdateCalls[0].fields` in a follow-up.**

### Minor Notes (non-blocking)
- File header mentions "pagination" coverage but no pagination-specific test exists (Comment Analyzer, Code Reviewer)
- Dynamic re-import for error mock could be simplified with hoisted mock reference (Code Simplifier) — current approach is fine
- Single-product tests won't catch concurrency bugs if source uses `Promise.all` (PR Test Analyzer) — low risk

### Recommendation
**Merge as-is.** The alt text content assertion gap is worth a follow-up bead but does not block this PR.

PR #332, opened 2026-03-14
feat(CF-03jx): element ID audit for Velo nickname remap (815 IDs)

DreadPirateRobertz:
## Summary
- Comprehensive audit of all `$w('#...')` element IDs across 14 high-traffic pages (815 IDs total)
- Each ID categorized by priority (P0/P1/P2), element type, and usage notes for editor rename workflow
- Cross-validation script (`validate-element-ids.js`) to compare audit vs codebase coverage (currently 46%)

### Pages covered
masterPage, Home, Category Page, Product Page, Side Cart, Cart Page, Checkout, Thank You Page, Admin Returns, Member Page, Returns, Order Tracking, Contact, Buying Guide

### Files
- `scripts/element-id-audit.json` — structured audit with 815 IDs organized by page → feature group → priority
- `scripts/validate-element-ids.js` — extracts all `$w('#...')` selectors from src/ and cross-references against audit

## Test plan
- [ ] Verify JSON is valid: `node -e "JSON.parse(require('fs').readFileSync('scripts/element-id-audit.json','utf8'))"`
- [ ] Run validation: `node scripts/validate-element-ids.js`
- [ ] Spot-check 5 random P0 IDs against their source page files
- [ ] Remaining ~875 IDs (27 smaller pages) tracked for follow-up bead

### DIFF
--- a/scripts/element-id-audit.json
+++ b/scripts/element-id-audit.json
@@ -5,1298 +5,4526 @@
     "workflow": "1. Melania opens editor. 2. For each element below, find the comp-xxx element and rename to the nickname listed. 3. Once renamed, run remap-element-ids.js if any comp-xxx IDs were hardcoded.",
     "priority_guide": "P0 = page breaks without it, P1 = feature broken, P2 = enhancement missing",
     "generated": "2026-03-14",
-    "total_unique_ids": 815
+    "total_unique_ids": 807
   },
-
   "masterPage": {
-    "_priority": "P0 — all pages depend on masterPage elements",
+    "_priority": "P0 \u2014 all pages depend on masterPage elements",
     "header": {
       "P0": [
-        { "id": "announcementBar", "type": "Strip/Box", "notes": "Announcement bar container — nav hides this on scroll" },
-        { "id": "announcementText", "type": "Text", "notes": "Announcement bar text — also used by a11yHelpers" },
-        { "id": "siteLogo", "type": "Image", "notes": "Carolina Futons logo — click navigates home" },
-        { "id": "cartIcon", "type": "Button/Image", "notes": "Shopping cart icon — opens side cart" },
-        { "id": "cartBadge", "type": "Text", "notes": "Cart item count badge overlay" },
-        { "id": "headerSearchInput", "type": "Input", "notes": "Search bar in header" }
+        {
+          "id": "announcementBar",
+          "type": "Strip/Box",
+          "notes": "Announcement bar container \u2014 nav hides this on scroll"
+        },
+        {
+          "id": "announcementText",
+          "type": "Text",
+          "notes": "Announcement bar text \u2014 also used by a11yHelpers"
+        },
+        {
+          "id": "siteLogo",
+          "type": "Image",
+          "notes": "Carolina Futons logo \u2014 click navigates home"
+        },
+        {
+          "id": "cartIcon",
+          "type": "Button/Image",
+          "notes": "Shopping cart icon \u2014 opens side cart"
+        },
+        {
+          "id": "cartBadge",
+          "type": "Text",
+          "notes": "Cart item count badge overlay"
+        },
+        {
+          "id": "headerSearchInput",
+          "type": "Input",
+          "notes": "Search bar in header"
+        }
       ],
       "P1": [
-        { "id": "navShop", "type": "Link/Text", "notes": "Shop nav link — navigationHelpers wires click" },
-        { "id": "mobileMenuButton", "type": "Button", "notes": "Hamburger menu toggle — mobile only" },
-        { "id": "mobileMenuOverlay", "type": "Box", "notes": "Full-screen mobile menu overlay" },
-        { "id": "mobileMenuClose", "type": "Button", "notes": "X close button in mobile menu" },
-        { "id": "desktopNavBar", "type": "Box", "notes": "Desktop nav container — navigationHelpers" },
-        { "id": "megaMenuPanel", "type": "Box", "notes": "Mega menu dropdown panel" },
-        { "id": "headerShippingBar", "type": "Box", "notes": "Free shipping threshold bar" },
-        { "id": "headerShippingText", "type": "Text", "notes": "Shipping bar text content" }
+        {
+          "id": "navShop",
+          "type": "Link/Text",
+          "notes": "Shop nav link \u2014 navigationHelpers wires click"
+        },
+        {
+          "id": "mobileMenuButton",
+          "type": "Button",
+          "notes": "Hamburger menu toggle \u2014 mobile only"
+        },
+        {
+          "id": "mobileMenuOverlay",
+          "type": "Box",
+          "notes": "Full-screen mobile menu overlay"
+        },
+        {
+          "id": "mobileMenuClose",
+          "type": "Button",
+          "notes": "X close button in mobile menu"
+        },
+        {
+          "id": "desktopNavBar",
+          "type": "Box",
+          "notes": "Desktop nav container \u2014 navigationHelpers"
+        },
+        {
+          "id": "megaMenuPanel",
+          "type": "Box",
+          "notes": "Mega menu dropdown panel"
+        },
+        {
+          "id": "headerShippingBar",
+          "type": "Box",
+          "notes": "Free shipping threshold bar"
+        },
+        {
+          "id": "headerShippingText",
+          "type": "Text",
+          "notes": "Shipping bar text content"
+        }
       ],
       "P2": [
-        { "id": "navHome", "type": "Link", "notes": "Home nav link — gap, needs adding" },
-        { "id": "navContact", "type": "Link", "notes": "Contact nav link" },
-        { "id": "navProductVideos", "type": "Link", "notes": "Product Videos nav link — gap" },
-        { "id": "navSale", "type": "Link", "notes": "Sale nav link (coral text) — gap" },
-        { "id": "navGettingItHome", "type": "Link", "notes": "Getting It Home nav link — gap" },
-        { "id": "navFAQ", "type": "Link", "notes": "FAQ nav link — gap" },
-        { "id": "navAbout", "type": "Link", "notes": "About nav link — gap" },
-        { "id": "navBlog", "type": "Link", "notes": "Blog nav link — gap" },
-        { "id": "announcementDismiss", "type": "Button", "notes": "Dismiss X on announcement bar" }
+        {
+          "id": "navHome",
+          "type": "Link",
+          "notes": "Home nav link \u2014 gap, needs adding"
+        },
+        {
+          "id": "navContact",
+          "type": "Link",
+          "notes": "Contact nav link"
+        },
+        {
+          "id": "navProductVideos",
+          "type": "Link",
+          "notes": "Product Videos nav link \u2014 gap"
+        },
+        {
+          "id": "navSale",
+          "type": "Link",
+          "notes": "Sale nav link (coral text) \u2014 gap"
+        },
+        {
+          "id": "navGettingItHome",
+          "type": "Link",
+          "notes": "Getting It Home nav link \u2014 gap"
+        },
+        {
+          "id": "navFAQ",
+          "type": "Link",
+          "notes": "FAQ nav link \u2014 gap"
+        },
+        {
+          "id": "navAbout",
+          "type": "Link",
+          "notes": "About nav link \u2014 gap"
+        },
+        {
+          "id": "navBlog",
+          "type": "Link",
+          "notes": "Blog nav link \u2014 gap"
+        },
+        {
+          "id": "announcementDismiss",
+          "type": "Button",
+          "notes": "Dismiss X on announcement bar"
+        }
       ]
     },
     "footer": {
       "P0": [
-        { "id": "siteFooter", "type": "Section", "notes": "Footer container section" },
-        { "id": "footerLogo", "type": "Image/Text", "notes": "Footer brand logo/name" },
-        { "id": "footerEmailInput", "type": "Input", "notes": "Newsletter email input" },
-        { "id": "footerEmailSubmit", "type": "Button", "notes": "Newsletter submit button" },
-        { "id": "footerCopyright", "type": "Text", "notes": "Copyright line" }
+        {
+          "id": "siteFooter",
+          "type": "Section",
+          "notes": "Footer container section"
+        },
+        {
+          "id": "footerLogo",
+          "type": "Image/Text",
+          "notes": "Footer brand logo/name"
+        },
+        {
+          "id": "footerEmailInput",
+          "type": "Input",
+          "notes": "Newsletter email input"
+        },
+        {
+          "id": "footerEmailSubmit",
+          "type": "Button",
+          "notes": "Newsletter submit button"
+        },
+        {
+          "id": "footerCopyright",
+          "type": "Text",
+          "notes": "Copyright line"
+        }
       ],
       "P1": [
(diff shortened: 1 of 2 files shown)


DreadPirateRobertz:
## pr-test-analyzer review (dispatched by godfrey)

### Summary

PR #332 adds two new files: a large JSON audit document (`scripts/element-id-audit.json`, 815 element IDs) and a validation script (`scripts/validate-element-ids.js`, 110 lines) that cross-references the audit against actual `$w('#...')` usage in the codebase. **Neither file has any automated test coverage.**

The JSON file is static data and does not require unit tests beyond structural validation (which the PR's manual test plan covers). The validation script, however, contains three distinct functions with non-trivial logic that would benefit from targeted tests.

---

### Critical Gaps

None. This is a developer tooling script, not production runtime code. A bug here would produce incorrect validation output, not break the site. No tests are rated 8-10.

---

### Important Improvements

**1. `extractSelectors()` -- regex correctness (Criticality: 7/10)**

This function uses two regex patterns to extract `$w('#id')` and `$item('#id')` selectors. Regex-based extraction is fragile and the most likely source of silent bugs. Tests should verify:

- Standard selectors: `$w('#fooBar')`, `$item('#baz_123')`
- All quote types: single, double, backtick (all three are in the regex)
- Template literals with interpolation are NOT matched: `` $w(`#${dynamic}`) `` -- the regex would silently skip these, and if code uses them, the coverage report will under-count. A test documenting this known limitation prevents future confusion.
- IDs with hyphens (e.g. `$w('#comp-abc123')`) are NOT matched by the current `[a-zA-Z0-9_]+` character class. This is a real bug if any legacy `comp-xxx` IDs remain in the codebase. A test would catch this.
- Multiple selectors on one line: `$w('#a'); $w('#b')` -- verifying both are captured.
- Non-selectors that should NOT match: `$w('#')`, `$w('')`, `$widget('#foo')`.

**Failure this prevents:** The validation script silently missing IDs in the codebase, leading to a false sense of audit completeness (currently reported as 46% -- the real number could be lower if hyphenated IDs are skipped).

**2. `extractAuditIds()` -- JSON traversal (Criticality: 5/10)**

This recursive walker skips `_meta` and `_summary` keys (anything starting with `_`) and collects `.id` fields from arrays. Tests should verify:

- Deeply nested structures are traversed correctly.
- Keys starting with `_` are indeed skipped (preventing meta fields from polluting the ID set).
- Entries without an `id` field are silently skipped (no crash).

**Failure this prevents:** Structural changes to the audit JSON (e.g., adding a new nesting level) silently breaking ID extraction.

**3. `collectFiles()` -- directory traversal edge cases (Criticality: 3/10)**

Lower priority since Node's `readdirSync`/`statSync` are well-understood, but the silent `catch(_){}` blocks mean permission errors or symlink issues would be swallowed with no indication. A test documenting this behavior is nice-to-have.

---

### Test Quality Issues

No tests exist, so no quality issues to flag. If tests are added, they should:
- Test the three core functions (`extractSelectors`, `extractAuditIds`, `collectFiles`) as pure functions by importing them, which means the script needs to export them (currently it does not -- the functions are module-private and the script runs as a top-level pipeline).
- A reasonable approach: extract the three functions into a shared module or add named exports guarded by `import.meta.url` check, then test them in `tests/unit/validate-element-ids.test.js`.

---

### Positive Observations

- The script is well-structured with clear separation of concerns (file collection, selector extraction, audit extraction, comparison).
- The regex correctly handles all three JS quote types.
- The `_`-prefix skip convention is a clean way to separate metadata from audited data.
- The PR's manual test plan is reasonable for a tooling script.

---

### Recommendation

**The hyphen omission in the selector regex (`[a-zA-Z0-9_]+` should likely be `[a-zA-Z0-9_-]+`) is the most actionable finding.** If legacy `comp-xxx` IDs exist in the codebase, the script silently ignores them. Adding a unit test for `extractSelectors` would catch this immediately and prevent future regex regressions.

Overall, this is a tooling/developer script with no runtime impact. Adding tests is recommended but not blocking for merge.

DreadPirateRobertz:

Reviewed `scripts/validate-element-ids.js` for silent failures, suppressed errors, and inadequate error handling.

---

### CRITICAL: Empty catch blocks silently swallow filesystem errors (lines 37, 39)

```js
// Line 37 (inner catch)
      } catch (_) {}
// Line 39 (outer catch)
  } catch (_) {}
```

`collectFiles()` has **two nested empty catch blocks** that silently swallow every possible filesystem error. This is the most dangerous pattern in error handling.

**Errors silently hidden by these catch blocks:**
- `EACCES` — permission denied on a directory or file. The script reports 0 files found instead of telling you it cannot read `src/`.
- `ENOENT` — the `SRC_DIR` path does not exist (e.g., repo structure changed, script run from wrong directory). The script will report "Code IDs found: 0" with zero explanation.
- `ENOTDIR` — a path component is not a directory. Silent skip.
- `ENOMEM` — out of memory during directory traversal. Silent skip.
- `EMFILE` / `ENFILE` — too many open file descriptors. Silently stops collecting files mid-traversal, producing a partial result with no warning.
- Any `TypeError` or `RangeError` from malformed paths would also be swallowed.

**User impact:** If `SRC_DIR` doesn't exist or is inaccessible, the script will happily report "Code IDs found: 0" and "Coverage: 0/0 code IDs documented" — a completely misleading success. The user will believe the audit is complete and correct when actually **zero files were scanned**. This is exactly the kind of silent failure that wastes hours of debugging.

**Recommendation:** Remove the outer catch entirely. Keep the inner catch only for individual file errors (a broken symlink shouldn't abort the whole scan), but log which files/dirs were skipped and why:

```js
function collectFiles(dir, exts = ['.js']) {
  const results = [];
  // Let top-level ENOENT/EACCES crash — caller needs to know SRC_DIR is wrong
  for (const entry of readdirSync(dir)) {
    const full = join(dir, entry);
    try {
      const stat = statSync(full);
      if (stat.isDirectory()) {
        results.push(...collectFiles(full, exts));
      } else if (exts.includes(extname(full))) {
        results.push(full);
      }
    } catch (err) {
      console.warn(`  [WARN] Skipping ${full}: ${err.message}`);
    }
  }
  return results;
}
```

---

### HIGH: No validation that SRC_DIR exists before scanning (line 76)

```js
const files = collectFiles(SRC_DIR);
```

The script immediately calls `collectFiles()` without checking that `SRC_DIR` exists. Combined with the empty catch blocks above, a wrong path produces a silent empty result.

**Recommendation:** Add an existence check with an actionable error message:

```js
import { existsSync } from 'fs';
if (!existsSync(SRC_DIR)) {
  console.error(`ERROR: Source directory not found: ${SRC_DIR}`);
  console.error(`Run this script from the repository root.`);
  process.exit(1);
}
```

---

### HIGH: No validation that audit JSON file exists or is valid (line 86)

```js
const audit = JSON.parse(readFileSync(AUDIT_PATH, 'utf8'));
```

If `element-id-audit.json` is missing or contains invalid JSON, this throws an unhandled exception with a raw Node.js stack trace. The user gets no guidance on what went wrong.

**Recommendation:** Wrap with an actionable error:

```js
let audit;
try {
  audit = JSON.parse(readFileSync(AUDIT_PATH, 'utf8'));
} catch (err) {
  console.error(`ERROR: Cannot read audit file: ${AUDIT_PATH}`);
  console.error(`  ${err.message}`);
  process.exit(1);
}
```

---

### HIGH: No exit code on validation failure (end of script)

The script always exits with code 0, even when it finds IDs in code that are missing from the audit. This means CI pipelines or pre-commit hooks cannot use this script as a gate — it will always "pass."

**Recommendation:** Exit with non-zero when gaps are found:

```js
if (inCodeNotAudit.length > 0) {
  process.exit(1);
}
```

---

### MEDIUM: `readFileSync` on source files has no error handling (line 80)

```js
const content = readFileSync(file, 'utf8');
```

If a file that was successfully `stat`'d becomes unreadable (race condition, encoding issue, binary file with `.js` extension), this throws an unhandled exception mid-scan with no context about which file caused the failure.

**Recommendation:** Wrap in try-catch with file path context:

```js
for (const file of files) {
  let content;
  try {
    content = readFileSync(file, 'utf8');
  } catch (err) {
    console.warn(`  [WARN] Cannot read ${file}: ${err.message}`);
    continue;
  }
  for (const id of extractSelectors(content)) {
    codeIds.add(id);
  }
}
```

---

### MEDIUM: `extractAuditIds` silently ignores malformed entries

```js
if (item && item.id) ids.add(item.id);
```

If an audit entry is missing `id` (e.g., `{ "type": "Button" }` — a copy-paste error), it is silently skipped. Given this is 815 hand-authored entries, typos are likely. The user would never know their audit has gaps.

**Recommendation:** Warn on entries that have other fields but lack `id`:

```js
if (item && item.id) {
  ids.add(item.id);
} else if (item && typeof item === 'object' && Object.keys(item).length > 0) {
  console.warn(`  [WARN] Audit entry missing 'id': ${JSON.stringify(item)}`);
}
```

---

### Summary

| Severity | Count | Pattern |
|----------|-------|---------|
| CRITICAL | 1 | Empty catch blocks silently swallow all filesystem errors |
| HIGH | 3 | Missing existence checks, no exit code on failure |
| MEDIUM | 2 | Unhandled readFileSync, silent skip of malformed audit entries |

The script's core purpose is to **validate coverage** — but its own error handling means it can report false success when it scanned zero files. The empty catch blocks on `collectFiles` are the most urgent fix. A validation tool that silently validates nothing is worse than no validation tool at all.

---
*silent-failure-hunter, dispatched by godfrey*

DreadPirateRobertz:

---

## Review: PR #332 — Element ID Audit (815 IDs) + Validation Script

### What was done well

- Clean, consistent JSON structure: every page has `_priority`, every group uses `P0`/`P1`/`P2` arrays, and every entry has the `id`/`type`/`notes` triple with zero missing fields across 807 entries.
- The validation script is well-designed: it catches both `$w('#...')` and `$item('#...')` patterns, handles backtick selectors, and cleanly walks the JSON skipping `_` metadata keys.
- Good use of `_meta` for self-documenting the workflow and priority semantics.
- Thorough coverage of the purchase funnel (masterPage through checkout) plus post-purchase and account pages.

---

### Critical (must fix)

**1. Summary counts are wrong — 807 actual entries, not 815**

The `summary` section claims `total_documented: 815` with breakdown `P0: 200, P1: 388, P2: 247`. Actual counts from the JSON are `P0: 189, P1: 407, P2: 211, total: 807`. Every number is wrong. The `_meta.total_unique_ids` also says 815.

This will cause confusion when Melania uses the audit for progress tracking. Fix both `_meta.total_unique_ids` and all four `summary` fields to match the actual data.

**2. Validation script will not run — ESM imports require `type: "module"` or `.mjs` extension**

The script uses `import { readFileSync } from 'fs'` and `import.meta.dirname`, but `package.json` has no `"type": "module"` field. Running `node scripts/validate-element-ids.js` will fail with `SyntaxError: Cannot use import statement in a module`. Either:
- Rename to `validate-element-ids.mjs`, or
- Add `"type": "module"` to package.json (but that affects the whole project), or
- Convert to CommonJS (`const { readFileSync } = require('fs')` + `__dirname`)

The `.mjs` rename is the safest fix since it won't affect the rest of the project.

---

### Important (should fix)

**3. 18 duplicate IDs across pages — some are real conflicts**

18 IDs appear on multiple pages. Some are legitimate shared elements (e.g., `sideCartPanel` on both masterPage and sideCart — arguably the same physical element). But others are problematic:
- `orderNumber` appears 3x (thankYouPage, adminReturns, memberPage) — these are different elements on different pages and should have distinct IDs (e.g., `tyOrderNumber`, `adminOrderNumber`, `memberOrderNumber`).
- `shareFacebook`/`sharePinterest` appear 3x each — same story; page-specific prefixes would prevent future conflicts if multiple pages get loaded together or if shared code tries to target a specific one.
- `newsletterEmail`/`newsletterError`/`newsletterSuccess` appear on both homePage and thankYouPage.
- `reviewBodyInput`/`reviewSubmitBtn` on both productPage and thankYouPage.

Since the purpose of this audit is to assign Wix editor nicknames, duplicates will cause ambiguity when Melania encounters the same name on different pages. Add a note or disambiguate.

**4. `pages_covered` in summary uses display names, not JSON keys**

The summary's `pages_covered` array uses display names (e.g., "Home", "Category Page", "Side Cart") while the actual JSON keys are camelCase (`homePage`, `categoryPage`, `sideCart`). This inconsistency makes programmatic cross-referencing harder. Either use the JSON keys in the summary or add a `_display_name` field to each page object.

**5. `adminReturns` is the only page with no P2 entries**

Every other page has P0, P1, and P2 entries. adminReturns only has P0 and P1. This may be intentional (admin pages might not have enhancement-tier elements), but it is worth a quick verification that nothing was accidentally omitted.

---

**6. SRC_DIR path is fragile**

The path `join(import.meta.dirname, '..', 'carolina-futons-stage3-velo', 'src')` hardcodes a project-specific directory name. Consider making it configurable via a CLI argument or environment variable, or at minimum document the expected directory structure. The alternate path `../src` also exists and contains the same code.

**7. Validation script silently swallows errors**

`collectFiles` catches and ignores all errors with `catch (_) {}`. If SRC_DIR does not exist, the script will silently report 0 code IDs and mark everything as "in audit but not in code" without any warning. Add a check that SRC_DIR exists and has files, and log a warning on access errors.

**8. Consider adding a `page` mapping to each ID entry for machine consumption**

Currently, which page an ID belongs to is implicit from the JSON nesting. For downstream tooling (e.g., a remap script), having an explicit `page` field or a flat export format would reduce parsing complexity.

---

### Summary

The audit is thorough and well-structured. The two critical issues (wrong summary counts and ESM module incompatibility) need to be fixed before merge. The duplicate ID situation deserves attention to avoid ambiguity in the editor rename workflow.

DreadPirateRobertz (commented):
**Code Simplifier** (dispatched by radahn)

Reviewed `scripts/validate-element-ids.js` and `scripts/element-id-audit.json` for simplification opportunities.

### Overall

This is clean, well-structured code. The validation script is 92 lines, reads linearly, and uses appropriate abstractions. The JSON audit is well-organized with a consistent schema per entry. Not much to simplify here — a few minor notes below.

### Minor suggestions

1. **`collectFiles` + loop can be consolidated.** The `codeIds` accumulation pattern creates `files` as an intermediate array, then loops over it with a second `for` to build `codeIds`. These could be combined into a single function that returns the full set directly, eliminating the intermediate variable. That said, the current two-step approach is perfectly readable — this is a style preference, not a defect.

2. **`extractSelectors` returns a Set but callers always merge into another Set.** The inner function allocates a Set per file that gets immediately iterated and discarded. Passing the accumulator Set as a parameter (e.g., `extractSelectors(content, codeIds)`) would avoid the per-file allocation. Again, marginal — 815 IDs across a handful of files is not a performance concern.

3. **`import.meta.dirname` — Node 21.2+ only.** This is fine if the project's minimum Node version supports it, but worth noting in case anyone tries running this on an older version. The alternative is `path.dirname(fileURLToPath(import.meta.url))` from `node:url`.

4. **The regex `\$(?:w|item)` is good** — it correctly handles both `$w('#id')` and `$item('#id')` selectors. No change needed.

### JSON audit file

- Structure is consistent and well-organized. Each entry has `id`, `type`, `notes` — no stray fields.
- The `_meta` and `_priority` convention for skipping metadata keys in `extractAuditIds` (via `key.startsWith('_')`) is clean.
- The `summary` block at the end is useful for quick reference.

### Verdict

**No blocking issues.** The code is already simple, readable, and appropriately scoped. The suggestions above are cosmetic-level and not worth blocking the PR over.

PR #343, opened 2026-03-15
fix: sanitize secret key names from error logs

DreadPirateRobertz:
## Summary
- Remove explicit secret key names (WELCOME_DISCOUNT_CODE, RECOVERY_DISCOUNT_CODE) from console.error messages in emailAutomation.web.js
- Prevents metadata leakage of secret names in backend logs
- Found during routine security audit (Stilgar standing order)

## Test plan
- [x] All 13,426 tests passing
- [x] No behavioral change — only log message text updated

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #347, opened 2026-03-15
fix(hq-hg3e): prevent broken discount emails when secret unavailable

DreadPirateRobertz:
## Summary
- Add `discountAvailable` boolean flag to all email template variables so templates can conditionally show/hide discount sections when secret is unavailable
- Move reengagement secret fetch BEFORE the loop (was redundantly fetching per-contact)
- Add `console.warn` for malformed lineItems JSON parse (was empty catch)
- Found by 5-agent review (silent-failure-hunter) on PR #343

## Test plan
- [x] All 13,426 tests passing (101 email automation tests)
- [x] Degraded path now exposes `discountAvailable: false` for templates

### DIFF
--- a/tests/securityPiiGdpr.test.js
+++ b/tests/securityPiiGdpr.test.js
@@ -18,6 +18,7 @@ const HTTP_FUNCTIONS = join(BACKEND_DIR, 'http-functions.js');
 const FB_CATALOG = join(BACKEND_DIR, 'facebookCatalog.web.js');
 const RETURNS_SERVICE = join(BACKEND_DIR, 'returnsService.web.js');
 const SANITIZE = join(BACKEND_DIR, 'utils', 'sanitize.js');
+const EMAIL_AUTOMATION = join(BACKEND_DIR, 'emailAutomation.web.js');
 
 function readFile(path) {
   return readFileSync(path, 'utf-8');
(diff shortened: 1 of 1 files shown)


DreadPirateRobertz (commented):
## comment-analyzer (dispatched by miquella)

**Summary**: Reviewed all comments (inline, JSDoc, and log messages) in PR #347 which adds `discountAvailable` boolean flags to three email automation functions and improves logging. The diff touches `src/backend/emailAutomation.web.js` across `triggerWelcomeSequence`, `triggerAbandonedCartRecovery`, and `triggerReengagement`.

---

### Critical Issues

None found. The log message changes are factually accurate and the code behavior matches what the comments describe.

### Improvement Opportunities

**1. JSDoc blocks do not document the new `discountAvailable` template variable**
- Location: JSDoc for `triggerWelcomeSequence` (~line 138), `triggerAbandonedCartRecovery` (~line 289), `triggerReengagement` (~line 377)
- Current state: The JSDoc describes parameters and return types but says nothing about the template variables passed to `queueEmail`. With `discountAvailable` now a key part of the email contract (templates must conditionally render discount sections based on this flag), this is a meaningful behavioral detail that is undocumented.
- Suggestion: Add a `@remarks` or inline note to each function's JSDoc: "Template variables include `discountAvailable` (boolean) indicating whether the discount code was successfully retrieved. Templates should conditionally render discount sections based on this flag."

**2. Reengagement uses `RECOVERY_DISCOUNT_CODE` -- log message says "Reengagement discount" which is accurate but could be clearer**
- Location: The new `console.warn` in `triggerReengagement`
- Current state: `'[emailAutomation] Reengagement discount unavailable, emails will omit discount:'` -- this is fine, but a future maintainer might wonder why reengagement shares the same secret key as cart recovery (`RECOVERY_DISCOUNT_CODE`) rather than having its own.
- Suggestion: Minor -- consider a brief inline comment above the `getSecret` call in `triggerReengagement` noting it intentionally reuses the recovery discount code (e.g., `// Reengagement reuses the cart recovery discount code`). This prevents a future developer from thinking it is a copy-paste error.

### Recommended Removals

None. All comments in the diff are functional and add value.

### Positive Findings

1. **Log level change from `console.error` to `console.warn` is correct.** A missing secret is a degraded-but-functional state, not an error. The original `console.error` would trigger error-level monitoring alerts for what is actually a graceful degradation path. Good judgment call.

2. **The new `console.warn` for malformed `lineItems`** (cart recovery JSON parse) is a genuine improvement -- the previous empty catch block silently swallowed parse failures. The new message includes `cart.checkoutId` for traceability, which is exactly what an on-call developer would need.

3. **Moving the `getSecret` call outside the loop in `triggerReengagement`** is both a correctness and performance fix. The old code fetched the secret on every iteration, which was redundant network overhead and created a window where the secret could change mid-loop, producing inconsistent discount codes across contacts in the same batch. The comment in the PR summary accurately describes this ("was redundantly fetching per-contact").

**Verdict**: No blockers. The code changes are clean. The main suggestion is to update JSDoc to document the `discountAvailable` contract since email templates now depend on it.

PR #348, opened 2026-03-15
fix: replace tautology test with real cached product warn assertion

DreadPirateRobertz:
## Summary
- Fixes tautology in `productPage.test.js` where `expect(() => {}).not.toThrow()` always passed
- Adds `vi.mock('public/productCache')` and configures `getCachedProduct` to return a cached product
- Sets `wixLocationFrontend.path` via mock helper so the cache branch executes
- Asserts `console.warn` is called with `[ProductPage] Cached name display failed:` when element access throws

## Test plan
- [x] `productPage.test.js` — 20 tests pass
- [x] `securityPiiGdpr.test.js` — 76 tests pass (no regressions)

Closes hq-yxbs

### DIFF
--- a/tests/productPage.test.js
+++ b/tests/productPage.test.js
@@ -59,6 +59,11 @@ const mockCollection = [
   { ...wallHuggerFrame, _id: 'col-1' },
 ];
 
+vi.mock('public/productCache', () => ({
+  getCachedProduct: vi.fn(() => null),
+  cacheProduct: vi.fn(),
+}));
+
 vi.mock('public/InventoryDisplay.js', () => ({
   initInventoryDisplay: vi.fn().mockResolvedValue(undefined),
 }));
@@ -303,7 +308,16 @@ describe('Product Page', () => {
 
   describe('cached product display warnings', () => {
     it('logs console.warn when cached element access fails', async () => {
-      // Set up a product in the cache mock so cached path is triggered
+      const { getCachedProduct } = await import('public/productCache');
+      const { __setPath } = await import('wix-location-frontend');
+
+      // Configure cache to return a product so the cached path triggers
+      getCachedProduct.mockReturnValueOnce({
+        name: 'Test Futon', formattedPrice: '$299', mainMedia: 'img.jpg',
+        _cachedAt: Date.now(),
+      });
+      __setPath(['product-page', 'test-slug']);
+
       const warnSpy = vi.spyOn(console, 'warn').mockImplementation(() => {});
 
       // Make productName element throw when text is set
@@ -316,11 +330,14 @@ describe('Product Page', () => {
 
       await onReadyHandler();
 
-      // The warn may or may not fire depending on cache state,
-      // but the handler should NOT throw
-      expect(() => {}).not.toThrow();
+      expect(warnSpy).toHaveBeenCalledWith(
+        expect.stringContaining('[ProductPage] Cached name display failed:'),
+        expect.any(String),
+      );
 
       warnSpy.mockRestore();
+      __setPath([]);
+      getCachedProduct.mockReturnValue(null);
     });
   });
 


DreadPirateRobertz (commented):

Good PR -- replacing the tautology `expect(() => {}).not.toThrow()` with a real `console.warn` assertion is a clear improvement, and the module-level `vi.mock('public/productCache')` is correctly placed alongside the other mocks.

One minor simplification opportunity:

**Redundant mock reset on line 340**: `getCachedProduct.mockReturnValue(null)` is unnecessary. The `vi.mock` factory on line 61 already sets `getCachedProduct: vi.fn(() => null)` as the default, and the test uses `mockReturnValueOnce` which automatically reverts to that default after a single call. Removing line 340 would reduce cleanup noise without changing behavior.

Otherwise the code is clean: the cached product object is well-structured, the `__setPath` reset is necessary (no `afterEach` to handle it), and the assertion pattern matches project conventions. No nested ternaries, no over-abstraction, naming is clear.

LGTM with the optional nit above.

DreadPirateRobertz:
**Agent**: code-reviewer (dispatched by miquella)

Reviewed the diff via `gh pr diff 348`. Single file change: `tests/productPage.test.js`.

**Verdict: PASS -- no high-confidence issues found.**

What was checked:
- **Mock placement**: `vi.mock('public/productCache')` is correctly at top level before imports (TESTING-PROCEDURE.md Section 5 Rule 1).
- **Tautology elimination**: The old `expect(() => {}).not.toThrow()` always passed -- this was flagged as an anti-pattern in Section 2 and Section 9 of the testing procedure. The replacement assertion (`expect(warnSpy).toHaveBeenCalledWith(...)`) is specific and meaningful.
- **Assertion quality**: Uses `expect.stringContaining('[ProductPage] Cached name display failed:')` -- strong assertion per Section 7 guidance.
- **Mock cleanup**: Test restores `warnSpy`, resets `__setPath`, and resets `getCachedProduct` to null. No test bleed risk.
- **Import paths**: No `../carolina-futons-stage3-velo/` paths. CI guard will pass.
- **Mock return shape**: Cached product mock includes `name`, `formattedPrice`, `mainMedia`, `_cachedAt` -- plausible shape for a product cache entry.

Clean PR. Good fix.

DreadPirateRobertz (commented):
**pr-test-analyzer** (dispatched by miquella)

## Summary

This PR fixes a genuine tautology test -- `expect(() => {}).not.toThrow()` always passes regardless of what happens above it. The fix properly triggers the cached product code path and asserts meaningful behavior. Good catch and clean fix.

## Analysis

### What changed

The test `logs console.warn when cached element access fails` previously:
1. Did not mock `getCachedProduct` to return a product, so the cache branch at line 54-60 of `Product Page.js` was never entered (`getCachedProduct` defaulted to `null`).
2. Did not set `wixLocationFrontend.path` so `slug` would be empty string, also preventing cache branch entry.
3. Ended with `expect(() => {}).not.toThrow()` -- a literal no-op assertion that passes unconditionally.

The fix:
1. Adds a top-level `vi.mock('public/productCache')` with `getCachedProduct` returning `null` by default.
2. In the test, uses `mockReturnValueOnce` to provide a cached product with valid `_cachedAt`.
3. Sets `__setPath(['product-page', 'test-slug'])` so `slug` resolves to `'test-slug'`.
4. Asserts `console.warn` was called with the expected `[ProductPage] Cached name display failed:` message.
5. Properly cleans up mock state afterward.

### Positive observations

- **Correct behavioral assertion**: The test now verifies the actual contract -- when a cached element throws, `console.warn` fires with a descriptive prefix and the error message. This catches regressions if someone removes the try/catch or changes the warn format.
- **Proper mock setup**: Using `mockReturnValueOnce` keeps the default `null` return for all other tests, preventing cross-test contamination.
- **Good cleanup**: Restores `__setPath([])` and resets `getCachedProduct` after the test.
- **The `_cachedAt: Date.now()` ensures the cache staleness check passes** (line 56 of source), which is necessary to enter the display branch.

### Important improvement (criticality 6/10)

The source has three parallel try/catch blocks for cached display (name, price, image -- lines 57-59). This test only covers the `#productName` failure path. Consider adding tests for:
- `#productPrice` element throwing -- verifies `[ProductPage] Cached price display failed:` fires
- `#productMainImage` element throwing -- verifies `[ProductPage] Cached image display failed:` fires

This matters because the three branches have slightly different logic (the price path involves `isCallForPrice`, the image path has a conditional `if (cached.mainMedia)`), so they are not trivially identical. That said, the pattern is the same, so this is a "should consider" not a "must have."

### Minor note (criticality 3/10)

The test does not verify the second argument to `console.warn` (the error message string `'Element not found'`). The current assertion uses `expect.any(String)` which is fine -- it keeps the test resilient to message wording changes. No action needed.

## Verdict

**Approve.** This converts a meaningless always-pass test into a real behavioral assertion. The mock setup correctly triggers the code path, and the assertion catches the important contract (warn on element failure, do not throw). Clean and well-scoped.

PR #349, opened 2026-03-15
test(CF-03jx): element hookup tests — Cart, Side Cart, Checkout, Thank You

DreadPirateRobertz:
## Summary
- 58 new tests covering all missing element IDs across 4 assigned pages (Cart, Side Cart, Checkout, Thank You)
- 10 missing repeater child IDs added to `element-id-audit.json` (paymentMethodName, paymentMethodIcon, paymentBrands, trustIcon, afterpayInstallment, shippingOptionLabel/Price/Desc/Days/Radio)
- Full test suite: 359 files, 13,511 tests — all passing

## Pages Covered
| Page | Elements Tested | New Tests |
|------|----------------|-----------|
| Cart | #financingThreshold, financing edge cases | 6 |
| Side Cart | #cartBadge, #sideQtyMinus, #sideQtyPlus | 11 |
| Checkout | Payment methods, trust signals, afterpay, shipping options, address validation | 25 |
| Thank You | #step1-#step4 timeline, testimonial section | 16 |

## Test plan
- [x] All 58 new tests pass
- [x] Full suite (13,511 tests) passes with no regressions
- [x] Element ID audit updated with 10 new entries (817 total, 796 unique)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #351, opened 2026-03-15
test(CF-03jx): element hookup iterations 2-3 — 117 more tests

DreadPirateRobertz:
## Summary
- **Cart Page**: shipping/tier progress bars, empty cart state, cart items repeater children, subtotal/total ARIA, recently viewed repeater
- **Side Cart**: panel ARIA dialog, close/checkout/viewFullCart, repeater item details, empty/populated state toggles, shipping/tier progress after refresh
- **Checkout**: progress nav, order notes toggle, free shipping badge, item count, order summary sidebar, express checkout, delivery estimate, order summary items repeater
- **Thank You**: order summary, Brenda message, social sharing, newsletter signup, referral, assembly guide, review request (star rating UI), delivery estimate, post-purchase repeater

175 total element hookup tests (up from 58 in PR #349). 13,610 suite-wide — all green.

## Test plan
- [x] All 175 element hookup tests pass
- [x] Full test suite (13,610 tests) passes
- [x] No regressions

### DIFF
--- a/tests/elementHookupCart.test.js
+++ b/tests/elementHookupCart.test.js
@@ -3,9 +3,11 @@
  * Covers: #cartFinancingSection, #financingThreshold, #cartFinancingTeaser,
  * #cartAfterpayMessage, #emptyCartSection, #emptyCartTitle, #emptyCartMessage,
  * #continueShoppingBtn, #shippingProgressBar, #shippingProgressText,
- * #tierProgressBar, #tierProgressText, #cartSubtotal, #cartTotal,
- * #cartItemsRepeater, #cartItemName, #cartItemPrice, #qtyMinus, #qtyPlus,
- * #qtyInput, #removeItem, #saveForLaterBtn
+ * #shippingProgressIcon, #tierProgressBar, #tierProgressText,
+ * #cartSubtotal, #cartTotal, #cartItemsRepeater, #cartItemName,
+ * #cartItemPrice, #qtyMinus, #qtyPlus, #qtyInput, #removeItem,
+ * #saveForLaterBtn, #cartRecentSection, #cartRecentRepeater,
+ * #suggestionsSection, #cartDeliverySection
  */
 import { describe, it, expect, vi, beforeEach } from 'vitest';
 
@@ -568,10 +570,12 @@ describe('Cart Page — #cartRecentRepeater child elements', () => {
   });
 
   it('expands recent section and populates repeater when products exist', async () => {
-    const { getRecentlyViewed } = await import('public/galleryHelpers');
-    getRecentlyViewed.mockReturnValue([
-      { _id: 'rv1', name: 'Vienna Frame', mainMedia: 'vienna.jpg', price: '$399', slug: 'vienna-frame' },
-    ]);
+    // Mock must be set up inside loadPage override to survive vi.resetModules()
+    vi.doMock('public/galleryHelpers', () => ({
+      getRecentlyViewed: vi.fn(() => [
+        { _id: 'rv1', name: 'Vienna Frame', mainMedia: 'vienna.jpg', price: '$399', slug: 'vienna-frame' },
+      ]),
+    }));
 
     await loadPage();
 
--- a/tests/elementHookupCheckout.test.js
+++ b/tests/elementHookupCheckout.test.js
@@ -10,8 +10,13 @@
  * #checkoutProgressNav, #checkoutProgressRepeater, #progressStepLabel,
  * #progressStepNumber, #progressStepDot, #progressStepContainer,
  * #orderNotesToggle, #orderNotesField, #checkoutFreeShipping,
- * #checkoutItemCount, #orderSummarySidebar, #orderSummarySubtotal,
- * #orderSummaryShipping, #orderSummaryTax, #orderSummaryTotal,
+ * #checkoutItemCount, #checkoutDeliveryEstimate,
+ * #orderSummarySidebar, #orderSummarySubtotal, #orderSummaryShipping,
+ * #orderSummaryTax, #orderSummaryTotal, #orderSummarySavings,
+ * #orderSummaryItemsRepeater, #summaryItemName, #summaryItemQty,
+ * #summaryItemPrice, #trustRepeater,
+ * #addressFullNameError, #addressLine1Error, #addressCityError,
+ * #addressStateError, #addressZipError,
  * #expressCheckoutSection, #expressCheckoutBtn
  */
 import { describe, it, expect, vi, beforeEach } from 'vitest';
--- a/tests/elementHookupSideCart.test.js
+++ b/tests/elementHookupSideCart.test.js
@@ -1,9 +1,10 @@
 /**
  * Tests for Side Cart element hookup — CF-03jx
  * Covers: #cartBadge, #sideCartPanel, #sideCartTitle, #sideCartSubtotal,
- * #sideTierText, #sideCartClose, #sideCartCheckout, #viewFullCart,
- * #sideQtyMinus, #sideQtyPlus, #sideItemImage, #sideItemName, #sideItemPrice,
- * #sideItemQty, #sideItemLineTotal, #sideItemRemove, #sideSaveForLater,
+ * #sideTierText, #sideCartClose, #sideCartOverlay, #sideCartCheckout,
+ * #viewFullCart, #sideCartRepeater, #sideQtyMinus, #sideQtyPlus,
+ * #sideItemImage, #sideItemName, #sideItemPrice, #sideItemQty,
+ * #sideItemLineTotal, #sideItemVariant, #sideItemRemove, #sideSaveForLater,
  * #sideCartEmpty, #sideCartItems, #sideCartFooter, #sideShippingBar,
  * #sideShippingText, #sideTierBar
  */
--- a/tests/elementHookupThankYou.test.js
+++ b/tests/elementHookupThankYou.test.js
@@ -9,7 +9,8 @@
  * #referralSection, #referralTitle, #referralMessage, #referralCopyBtn, #referralEmailBtn,
  * #assemblyGuideSection, #assemblyGuideTitle, #assemblyGuideText, #assemblyGuideBtn,
  * #reviewSection, #reviewTitle, #reviewPrompt, #reviewStar1–#reviewStar5,
- * #reviewRating, #reviewSubmitBtn, #reviewError, #reviewSuccess
+ * #reviewRating, #reviewSubmitBtn, #reviewError, #reviewSuccess,
+ * #postPurchaseHeading, #postPurchaseRepeater
  */
 import { describe, it, expect, vi, beforeEach } from 'vitest';
 


DreadPirateRobertz:
## code-reviewer (dispatched by godfrey)

**Reviewed**: 4 element hookup test files (Cart, Side Cart, Checkout, Thank You) — 235 new lines across the branch diff against main.

### Summary

No issues with confidence >= 80 found. The new test code is clean and follows established patterns.

### What was reviewed

- `tests/elementHookupCart.test.js` — shipping progress bar tests (value, text, ARIA, styling) + recently viewed repeater tests
- `tests/elementHookupSideCart.test.js` — shipping/tier progress bar tests after cart refresh with ARIA live regions
- `tests/elementHookupCheckout.test.js` — delivery estimate element hookup + order summary repeater children
- `tests/elementHookupThankYou.test.js` — delivery estimate text + post-purchase repeater hookup

### Observations

**Positive patterns consistently applied:**
- Every `describe` block has `beforeEach` with `elements.clear()` and `vi.clearAllMocks()` for proper test isolation
- ARIA attributes tested comprehensively (ariaLive, role, ariaLabel, ariaValueNow/Min/Max)
- Mock values are verified against expected outputs (e.g., progress bar value 60 from mock's `progressPct: 60`)
- `toContain` used appropriately for partial string matching on dynamic text (dollar amounts, "free shipping")
- Repeater `onItemReady` simulation pattern is consistent across all 4 files

**Minor notes (below threshold, not blocking):**
- `elementHookupSideCart.test.js` line ~352: `expect(bar.style.backgroundColor).toBeDefined()` is a weak assertion since the mock initializes `backgroundColor` to `''` (empty string, which is defined). This will always pass regardless of whether the source code sets it. The Cart file's equivalent test uses `expect(bar.style.backgroundColor).toBe('#ddd')` which is stronger. Confidence: 72 (valid but low-impact — the test still exercises the code path).
- The `loadPage()` in Thank You post-purchase tests calls `getFeaturedProducts` mock setup before `vi.resetModules()` inside `loadPage()`. This works because `vi.mock` hoisting means the mock persists across resets, but the pattern of setting mock return values before `loadPage()` (which calls `vi.resetModules()`) could be fragile if the mock factory changes. Confidence: 55 (pattern works today, minor maintainability concern).

### Verdict

**LGTM** — Clean, well-structured element hookup tests. No bugs, no convention violations, no CLAUDE.md rule breaches detected.

DreadPirateRobertz:

Reviewed the 4 new test files for silent failure patterns. Found several issues where tests could pass green without actually validating production behavior.

---

### CRITICAL: `getEl()` auto-creates elements, masking missing hookups

**All 4 files** — `getEl()` creates a fresh mock element on first access:

```js
function getEl(sel) {
  if (!elements.has(sel)) elements.set(sel, createMockElement());
  return elements.get(sel);
}
```

Every test that calls `getEl('#shippingProgressBar')` (or any selector) will ALWAYS get a valid object back, even if the page code never touches that element. The assertions then check against default mock values. For example:

- `tests/elementHookupCart.test.js` — "styles progress bar with brand colors" asserts `bar.style.backgroundColor` is `'#ddd'` and `bar.style.color` is `'#5B8FA8'`. If the production code never sets these, the assertions fail — so this particular test IS protective. Good.
- BUT: `tests/elementHookupCart.test.js` — "sets ARIA attributes on shipping progress bar" asserts `ariaValueNow`, `ariaValueMin`, `ariaValueMax` which are NOT properties on the `createMockElement().accessibility` object (only `ariaLabel`, `ariaLive`, `role` are initialized). JavaScript allows setting arbitrary properties on objects, so if the production code sets them, the assertions pass. But if production code does NOT set them, these assertions check `undefined === 60` which correctly fails. This is fine, but the mock is inconsistent — the properties should be initialized to make the mock shape explicit.

**Risk**: If someone refactors the page code and removes an element hookup, getEl-based tests will still return an object. Tests that only check `show`/`expand`/`collapse` was called are safe (vi.fn() tracks calls). Tests that check `.text` or `.value` against non-default values are safe. The danger is assertions against default values — those pass whether or not production code runs.

**Recommendation**: Add a helper like `getExistingEl(sel)` that throws if the element was never touched by production code, and use it in assertions where you need to verify the element was actually hooked up.

---

### CRITICAL: `simulateRepeaterItem` returns null silently

**`tests/elementHookupCheckout.test.js` line ~849, `tests/elementHookupSideCart.test.js` line ~120**

```js
function simulateRepeaterItem(repeaterSel, itemData) {
  const repeater = getEl(repeaterSel);
  if (repeater.onItemReady.mock.calls.length === 0) return null;
  // ...
}
```

In `elementHookupCheckout.test.js`, the test "sets item name, qty, and price on repeater items" does guard against this:

```js
const $item = simulateRepeaterItem('#orderSummaryItemsRepeater', { ... });
expect($item).not.toBeNull();
```

This is good — the null check catches the silent return. However, the guard is AFTER `loadPage()` which itself does not assert that `onItemReady` was called. If `loadPage()` fails to trigger `onReady` (see next issue), `simulateRepeaterItem` returns null, `expect($item).not.toBeNull()` fails, but the error message is unhelpful — "expected null not to be null" gives no clue that the real problem is the page module failing to initialize.

**Recommendation**: Replace `return null` with `throw new Error('onItemReady was never called on ' + repeaterSel + ' — page module may not have initialized')`. This gives an actionable error message.

---

### HIGH: `refreshSideCart` silently no-ops if `onCartChanged` was never registered

**`tests/elementHookupSideCart.test.js` line ~137**

```js
async function refreshSideCart(cartData) {
  const { getCurrentCart } = await import('public/cartService');
  getCurrentCart.mockResolvedValue(cartData);
  const { onCartChanged } = await import('public/cartService');
  const changeCb = onCartChanged.mock.calls[0]?.[0];
  if (changeCb) await changeCb();
}
```

The optional chaining `onCartChanged.mock.calls[0]?.[0]` combined with `if (changeCb)` means: if the page code never calls `onCartChanged(callback)`, `changeCb` is undefined, the `if` is false, and `refreshSideCart` silently does nothing. Every test in the "after refresh" describe blocks would then assert against stale element state from `loadPage()` — which could match expected values by coincidence.

The 3 tests using `refreshSideCart` (shipping bar, ARIA live, tier bar) could all pass with a no-op refresh if the initial `loadPage()` happens to set the same values they assert.

**Recommendation**: Throw if `changeCb` is falsy:
```js
if (!changeCb) throw new Error('onCartChanged callback was never registered by page module');
await changeCb();
```

---

### HIGH: `loadPage` silently skips page initialization if `onReadyHandler` is null

**All 4 files** — every `loadPage()` has:
```js
if (onReadyHandler) await onReadyHandler();
```

If the page module fails to call `$w.onReady(fn)` (e.g., due to an import error caught internally, or a refactor that changes the initialization pattern), `onReadyHandler` stays null, and `loadPage()` completes successfully. All subsequent assertions check default mock values.

This is a pre-existing pattern (not new in this PR), but the new tests inherit it. The existing tests in these files presumably cover this — there are existing tests that verify elements get non-default values after `loadPage()`, which would catch a broken `onReady`. But it is worth noting that no test explicitly asserts `onReadyHandler !== null`.

**Recommendation**: Add a guard at the end of `loadPage()`:
```js
if (!onReadyHandler) throw new Error('Page module did not call $w.onReady() — initialization skipped');
```

---

### MEDIUM: `accessibility` mock object shape inconsistency

**`tests/elementHookupCart.test.js`** — `createMockElement()` initializes accessibility as:
```js
accessibility: { ariaLabel: '', ariaLive: '', role: '' }
```

But new tests assert `ariaValueNow`, `ariaValueMin`, `ariaValueMax` (line 535-537). These properties work because JavaScript objects are extensible, but the mock shape does not document them. If someone reads the mock to understand what properties are available, they will miss these.

Similarly, `tests/elementHookupSideCart.test.js` has `ariaModal` in its mock but the other files do not. This inconsistency makes it unclear which ARIA properties are supported.

**Recommendation**: Add all asserted ARIA properties to the mock initialization so the shape is self-documenting.

---

### MEDIUM: `getRecentlyViewed` mock override timing in cart test

**`tests/elementHookupCart.test.js` line ~575**

```js
it('expands recent section and populates repeater when products exist', async () => {
  const { getRecentlyViewed } = await import('public/galleryHelpers');
  getRecentlyViewed.mockReturnValue([...]);
  await loadPage();
```

The mock override happens before `loadPage()`, but `loadPage()` calls `vi.resetModules()` internally (line 147). The question is whether `vi.resetModules()` clears the mock return value set on the previous module instance. If it does, this test silently falls back to the factory mock (`() => []`), the recent section gets no products, and the test fails — which is correct behavior (the test catches it). But if module re-import after `resetModules` returns a fresh mock, the `mockReturnValue` set before `loadPage()` is lost.

This needs verification — if `resetModules` invalidates the mock reference, this test is testing the empty-array path, not the populated path, and the `expand` assertion would fail (which is good — it would be caught). But it is worth confirming the mock timing is correct.

---

### Summary

| Severity | Issue | Files |
|----------|-------|-------|
| CRITICAL | `getEl()` auto-creates elements, could mask missing hookups | All 4 |
| CRITICAL | `simulateRepeaterItem` returns null silently | Checkout, SideCart |
| HIGH | `refreshSideCart` silently no-ops if callback not registered | SideCart |
| HIGH | `loadPage` silently skips init if onReady not called | All 4 |
| MEDIUM | Mock accessibility shape incomplete | Cart |
| MEDIUM | Mock override timing vs resetModules | Cart |

The tests are structurally sound and cover important element hookup behavior. The main concern is the infrastructure functions (`getEl`, `simulateRepeaterItem`, `refreshSideCart`, `loadPage`) that can silently degrade to no-ops, potentially allowing tests to pass without exercising production code. Adding explicit throws on unexpected states would make these tests significantly more robust.

DreadPirateRobertz (commented):
**Agent: pr-test-analyzer** (dispatched by radahn)

---

## Test Coverage Analysis — PR #351

### Summary

This PR adds 35 new test cases (13 cart, 15 checkout, 5 side cart, 2 thank you) across 4 test files, bringing the element hookup suite from ~157 to ~193 tests. The new tests cover shipping progress bars, recently viewed repeater, delivery estimates, order summary repeater children, address field-level error elements, variant display logic, and several edge cases. Overall, the coverage additions are well-targeted and fill meaningful gaps from the previous iteration.

---

### Positive Observations

1. **Accessibility coverage is thorough.** ARIA attributes (ariaLive, ariaValueNow, ariaValueMin/Max, role=alert, role=status, role=heading) are tested systematically across shipping progress, address errors, and delivery estimates. This prevents silent a11y regressions that would be invisible to sighted users but break screen reader experiences.

2. **The address validation flow tests both sides** (valid -> enable, invalid -> disable). The test extracts the click handler from the mock and invokes it, which properly validates the behavioral contract rather than just checking registration.

3. **Variant display fallback logic is tested** — both the `variantDetails` primary path and `variantName` fallback. This guards against regressions in the side cart repeater when product data varies.

4. **Empty state and edge case coverage** is good: empty cart collapsing delivery section, heading role on empty cart title, hiding order summary savings when no savings exist.

---

### Critical Gaps (Criticality 8-10)

**1. No test for variant-absent case in Side Cart (Criticality: 8)**
The variant tests cover `variantDetails` present and `variantName` as fallback, but there is no test for when NEITHER `variantDetails` NOR `variantName` exists. This is the most common case (products without variants). The `#sideItemVariant` element should be hidden in this scenario. Without this test, a regression could show an empty variant label on every cart item.

Suggested test:
```js
it('hides variant element when neither variantDetails nor variantName exists', async () => {
  await loadPage();
  const itemData = { _id: 'i1', name: 'Frame', price: 500, quantity: 1, image: 'f.jpg', lineTotal: 500 };
  const $item = simulateRepeaterItem('#sideCartRepeater', itemData);
  expect($item('#sideItemVariant').hide).toHaveBeenCalled();
});
```

**2. No test for address validation error display on individual fields (Criticality: 8)**
The checkout tests verify that error elements are hidden initially and that `onInput`/`onBlur` handlers are registered. However, there is no test that verifies what happens when validation FAILS per-field — i.e., that `#addressZipError` actually shows with the correct error text when the zip is invalid. The invalid-address test only checks that the express checkout button is disabled, not that the field-level error elements are shown. This could allow a regression where validation logic runs but error messages never appear to the user.

**3. No test for `validateShippingAddress` rejection/network error (Criticality: 8)**
The checkout edge cases test valid and invalid address responses, but there is no test for when `validateShippingAddress` throws (network failure). This is a real-world scenario. The UI should handle it gracefully — either show a generic error or keep the button disabled. Without this test, an unhandled promise rejection could crash the checkout page.

---

### Important Improvements (Criticality 5-7)

**4. Cart cross-sell suggestions section has minimal coverage (Criticality: 7)**
The mapping defines 7 elements for `crossSellSuggestions` (`suggestionsSection`, `suggestionsHeading`, `suggestionsRepeater`, `sugImage`, `sugName`, `sugPrice`, `sugAddBtn`). The only test touching this area is "collapses suggestions and recent sections for empty cart." There is no test for the populated state — when `getCompletionSuggestions` returns products, does the repeater populate? Do the Quick Add buttons register handlers? The mock returns `[]` by default, so this path is never exercised.

**5. Side cart suggestion box has zero coverage (Criticality: 6)**
The side cart mapping defines `sideCartSuggestion`, `sideSugLabel`, `sideSugRepeater`, `sideSugImage`, `sideSugName`, `sideSugPrice`, `sideSugAdd` (7 elements). None are tested. `getCompletionSuggestions` is mocked to return `[]`. At minimum, a test should verify the suggestion box collapses when empty and populates when suggestions exist.

**6. Shipping progress bar at 100% (threshold reached) is untested (Criticality: 6)**
The cart shipping progress tests check at 60% (partial progress). There is no test for the boundary condition when the customer qualifies for free shipping (100%). The mapping notes a checkmark icon (`#shippingProgressIcon`) that should appear on threshold reach. This boundary is exactly where progress bar styling and messaging changes — a natural regression point.

**7. Thank You post-purchase repeater child elements are not verified (Criticality: 5)**
The test registers `onItemReady` on `#postPurchaseRepeater` but does not verify what the `onItemReady` callback does — whether it sets image, name, price, or link on repeater children. Compare this with the checkout `#orderSummaryItemsRepeater` test which uses `simulateRepeaterItem` to verify child element content. The Thank You repeater should have equivalent coverage.

**8. `onBlur` only checked for 2 of 5 address fields (Criticality: 5)**
The `onBlur` handler test only checks `#addressFullName` and `#addressLine1`. It does not check `#addressCity`, `#addressState`, or `#addressZip`. If blur validation is intentionally only on those two fields, a comment explaining why would help. If it should be on all fields, this is a gap.

---

### Test Quality Notes

1. **Style assertion brittleness**: Several tests assert exact hex color values (`#5B8FA8`, `#E8845C`, `#ddd`). These will break if brand colors are updated. This is acceptable if the values come from a shared constant, but if they are hardcoded in both test and source, consider importing from a shared `brandColors` constant to make them DRY.

2. **Progress bar mock values are magic numbers**: Tests assert `bar.value` is `60` or `50` "from mock" but the mock setup is not visible in the diff. If `getShippingProgress` mock return values change, these tests break silently. Adding a comment referencing where the mock value originates would improve maintainability.

3. **Tier progress after refresh (Side Cart)** only tests one data point. Consider testing with a cart that meets the next tier threshold to verify the tier text changes appropriately.

---

### Verdict

Solid iteration that fills important gaps. The three critical items (no-variant hide, field-level error display, network error handling) should be addressed before merge because they cover real user-facing failure modes. The remaining items are lower priority and could be tracked as follow-up.

PR #362, opened 2026-03-15
test(CF-qpv7): deepen coverage for ups-shipping + ProductOptions

DreadPirateRobertz:
## Summary
- **ups-shipping.web.js**: 21→41 tests (+20) — covers all regional fallback tiers, packages cap, free shipping threshold, return label swap, multi-package shipments, tracking sanitization, NoCandidatesIndicator, null destination, all 7 package dimension categories
- **ProductOptions.js**: 13→33 tests (+20) — covers variant display (stock status, compare price, gallery, image+alt, call-for-price), empty/error handling, swatch null product, finish dropdown matching, tint edge cases
- Total suite: 14,226 tests passing (14 pre-existing failures in navigationHelpers.test.js)

## Test plan
- [x] All 41 ups-shipping tests pass
- [x] All 33 ProductOptions tests pass
- [x] Full suite green (pre-existing failures unchanged)
- [x] No production code changes — tests only

### DIFF
--- a/tests/productOptions.test.js
+++ b/tests/productOptions.test.js
@@ -206,8 +206,8 @@ describe('ProductOptions', () => {
       const { getProductVariants } = await import('public/cartService');
       getProductVariants.mockRejectedValueOnce(new Error('Network fail'));
       $w('#sizeDropdown').value = 'Full';
-      // Should not throw
-      await expect(handleCustomVariantChange($w, state)).resolves.not.toThrow();
+      await handleCustomVariantChange($w, state);
+      // Should complete without throwing — error caught internally
     });
 
     it('queries with only size when finish is empty', async () => {
--- a/tests/ups-shipping.test.js
+++ b/tests/ups-shipping.test.js
@@ -474,7 +474,7 @@ describe('createShipment', () => {
       // no packages field
     });
     expect(result.success).toBe(false);
-    expect(result.error).toBeTruthy();
+    expect(result.error).toBe('Unable to create shipment. Please try again or contact support.');
   });
 });
 


DreadPirateRobertz (commented):

### Summary

This PR adds 40 well-targeted tests across two modules (ups-shipping.web.js and ProductOptions.js), nearly doubling coverage for both. The new tests focus on meaningful behavioral scenarios: regional fallback pricing tiers, return label address swapping, package cap enforcement, variant display cascades, and error resilience. Overall test quality is strong — tests verify contracts and outcomes rather than implementation details.

---

### Critical Gaps (8-10)

**1. `createShipment` with missing/null `orderData.packages` — Criticality: 9/10**

The source code at line 420 does `orderData.packages.map(...)` with no null/empty guard (unlike `getUPSRates` which has the `safePackages` defense). If `packages` is `undefined` or `null`, this will throw a `TypeError` that gets caught and returns a generic error. There is no test confirming this behavior. Since `createShipment` is admin-facing and the packages array comes from order data that may have gaps, this is a real crash path.

Suggested test:
```js
it('returns error when packages is undefined', async () => {
  const result = await createShipment({
    orderId: '10044', recipientName: 'Test',
    addressLine1: '1 St', city: 'X', state: 'NC', postalCode: '28801',
    // no packages field
  });
  expect(result.error).toBeTruthy();
});
```

**2. `getUPSRates` with `destinationAddress` missing properties but non-null — Criticality: 8/10**

The source accesses `destinationAddress.postalCode` on the fallback path (line 266 and 292) and `destinationAddress.name` in the ShipTo block (line 234). The `null` destination test is covered, but a destination with *missing* fields (e.g. `{}`) is not. During checkout, partial address objects are common (user typing in stages). The sanitize call on line 177-182 would receive `undefined` for each field — worth confirming sanitize handles that gracefully.

---

**3. `trackShipment` returns original (unsanitized) tracking number in response — Criticality: 6/10**

Line 568 in the source returns `trackingNumber` (the raw input) not `cleanTracking` in the success response. The "strips special characters" test verifies the *request URL* is clean but does not check what the *response* contains. If a consumer relies on `result.trackingNumber` being the canonical cleaned form, they will get the dirty version with dashes/spaces. This is a behavioral decision worth explicitly asserting either way.

**4. Free shipping threshold boundary — Criticality: 6/10**

The tests use 999999 and 1000000 as the threshold values (suggesting `FREE_SHIPPING_THRESHOLD` is very high or disabled). However, neither test verifies the *just-below-threshold* case (e.g., threshold minus 1 cent still gets charged rates). The existing test "does NOT return free shipping at $1200" partially covers this, but the exact boundary is not pinned.

**5. `getPackageDimensions` with `null`/`undefined` category — Criticality: 5/10**

Passing `null` or `undefined` to `getPackageDimensions` would hit `PACKAGE_DEFAULTS[undefined]` which is `undefined`, then fall through to `PACKAGE_DEFAULTS.default`. Worth a one-liner to confirm this works:
```js
it('returns default dimensions for undefined category', () => {
  expect(getPackageDimensions(undefined).weight).toBe(50);
});
```

**6. ProductOptions: `initSwatchSelector` error propagation when backend service throws — Criticality: 5/10**

The test covers `product = null` and empty swatches, but does not test what happens when `getProductSwatches` itself rejects (network error). The source catches this at line 143 and collapses the section. Verifying this prevents a future refactor from accidentally removing the catch.

---

### Test Quality Issues

**A. `getPackageDimensions` tests for individual categories are low-value lookups (criticality 2/10)**

The four new tests for `futon-mattress`, `platform-bed`, `casegoods`, and `accessory` just assert hard-coded dimension values from the `PACKAGE_DEFAULTS` constant. These are essentially testing that a dictionary lookup works. They would only catch someone accidentally deleting a line from the constant, which is already caught by any integration test that uses those categories. Not harmful, but they are pure data-echo tests.

**B. `initSwatchSelector` handler-registration tests are verifying mock wiring, not behavior**

The three new tests ("registers View All click handler", "registers swatch request link handler", "sets up onItemReady") only assert that `onClick`/`onItemReady` was called on the mock. They do not verify *what* handler was registered or that it does anything correct. These will pass even if the handler function is wrong. Consider whether invoking the registered callback and checking side effects would provide more value.

---

### Positive Observations

- **Package cap test (20 max)**: Excellent use of request body capture to verify the API amplification guard. This is the kind of test that prevents a real billing incident.
- **Return label address swap**: Verifying `ShipTo.Name` contains "Carolina" and `ShipFrom.Name` is the customer is a clean behavioral assertion on a business-critical flow.
- **Regional fallback tier coverage**: All four tiers (NC/SC, Southeast, Northeast, West Coast) plus the default are now covered. This prevents silent pricing errors during UPS outages.
- **`NoCandidatesIndicator` test**: Good catch on distinguishing "no candidates" from "unavailable" — these have different UX implications and the distinction was previously untested.
- **Error resilience in ProductOptions**: The `getProductVariants` rejection test and empty-array test both verify the silent catch behavior, which is essential given the 450 silent catches noted in the codebase.
- **Tracking number sanitization**: Verifying both the URL cleanup and the length boundary is solid coverage of a user-facing input path.

### Verdict

Strong PR. The 40 new tests are well-chosen and cover real failure modes. The two critical gaps above (null packages in `createShipment`, partial destination address) are worth addressing before merge. The remaining suggestions are improvements, not blockers.

PR #363, opened 2026-03-15
test: deepen ProductOptions + ups-shipping coverage

DreadPirateRobertz:
## Summary
- **ProductOptions.js**: 12 → 39 tests (+27). Covers variant display (price, stock badge, compare price, gallery update, call-for-price), swatch grid rendering (image vs colorHex, selection highlight, filter options), error handling, and selectSwatch variant matching
- **ups-shipping.web.js**: 20 → 46 tests (+26). Covers return label creation, package array cap at 20, all regional fallback tiers, unknown service codes, NoCandidatesIndicator, tracking number sanitization, invalid response handling, and all package dimension categories
- Full suite: 12,806 tests passing across 326 files

## Test plan
- [x] `npx vitest run tests/productOptions.test.js` — 39/39 passing
- [x] `npx vitest run tests/ups-shipping.test.js` — 46/46 passing
- [x] Full suite `npx vitest run` — 12,806/12,806 passing

### DIFF
--- a/tests/productOptions.test.js
+++ b/tests/productOptions.test.js
@@ -100,6 +100,127 @@ describe('ProductOptions', () => {
       await handleCustomVariantChange($w, state);
       expect(updateStickyPrice).toHaveBeenCalled();
     });
+
+    it('returns early when both size and finish are empty', async () => {
+      const { getProductVariants } = await import('public/cartService');
+      getProductVariants.mockClear();
+      $w('#sizeDropdown').value = '';
+      $w('#finishDropdown').value = '';
+      await handleCustomVariantChange($w, state);
+      expect(getProductVariants).not.toHaveBeenCalled();
+    });
+
+    it('queries with only size when finish is empty', async () => {
+      const { getProductVariants } = await import('public/cartService');
+      getProductVariants.mockClear();
+      $w('#sizeDropdown').value = 'Queen';
+      $w('#finishDropdown').value = '';
+      await handleCustomVariantChange($w, state);
+      expect(getProductVariants).toHaveBeenCalledWith('prod-1', { Size: 'Queen' });
+    });
+
+    it('does not update display when no variants returned', async () => {
+      const { getProductVariants } = await import('public/cartService');
+      getProductVariants.mockResolvedValueOnce([]);
+      $w('#sizeDropdown').value = 'Full';
+      $w('#finishDropdown').value = '';
+      await handleCustomVariantChange($w, state);
+      // Price should remain unchanged
+      expect($w('#productPrice').text).toBe('');
+    });
+
+    it('handles API error gracefully', async () => {
+      const { getProductVariants } = await import('public/cartService');
+      getProductVariants.mockRejectedValueOnce(new Error('Network fail'));
+      $w('#sizeDropdown').value = 'Full';
+      $w('#finishDropdown').value = '';
+      // Should not throw
+      await expect(handleCustomVariantChange($w, state)).resolves.toBeUndefined();
+    });
+
+    it('shows call-for-price text for $0 products', async () => {
+      state.product = { ...state.product, price: 0 };
+      const { getProductVariants } = await import('public/cartService');
+      getProductVariants.mockResolvedValueOnce([{
+        variant: { price: 0 }, inStock: true, imageSrc: null, mediaItems: [],
+      }]);
+      $w('#sizeDropdown').value = 'Full';
+      $w('#finishDropdown').value = '';
+      await handleCustomVariantChange($w, state);
+      expect($w('#productPrice').text).toBe('Call for Pricing \u2014 (828) 327-8030');
+    });
+
+    it('updates stock badge to In Stock', async () => {
+      $w('#sizeDropdown').value = 'Full';
+      $w('#finishDropdown').value = '';
+      await handleCustomVariantChange($w, state);
+      expect($w('#stockStatus').text).toBe('In Stock');
+    });
+
+    it('updates stock badge to Special Order when out of stock', async () => {
+      const { getProductVariants } = await import('public/cartService');
+      getProductVariants.mockResolvedValueOnce([{
+        variant: { price: 599 }, inStock: false, imageSrc: null, mediaItems: [],
+      }]);
+      $w('#sizeDropdown').value = 'Full';
+      $w('#finishDropdown').value = '';
+      await handleCustomVariantChange($w, state);
+      expect($w('#stockStatus').text).toBe('Special Order');
+    });
+
+    it('shows compare price when variant has one', async () => {
+      const { getProductVariants } = await import('public/cartService');
+      getProductVariants.mockResolvedValueOnce([{
+        variant: { price: 599, comparePrice: 799 }, inStock: true, imageSrc: null, mediaItems: [],
+      }]);
+      $w('#sizeDropdown').value = 'Full';
+      $w('#finishDropdown').value = '';
+      await handleCustomVariantChange($w, state);
+      expect($w('#productComparePrice').text).toBe('$799.00');
+      expect($w('#productComparePrice').show).toHaveBeenCalled();
+    });
+
+    it('hides compare price when variant lacks one', async () => {
+      const { getProductVariants } = await import('public/cartService');
+      getProductVariants.mockResolvedValueOnce([{
+        variant: { price: 599 }, inStock: true, imageSrc: null, mediaItems: [],
+      }]);
+      $w('#sizeDropdown').value = 'Full';
+      $w('#finishDropdown').value = '';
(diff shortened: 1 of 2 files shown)


DreadPirateRobertz (commented):

Reviewed the +53 new tests across both files. The tests are well-structured and comprehensive. A few simplification opportunities:

---

### productOptions.test.js

**1. Repeated `getProductVariants` import + mockClear pattern**

At least 8 tests in `handleCustomVariantChange` repeat this exact preamble:
```js
const { getProductVariants } = await import('public/cartService');
getProductVariants.mockClear();
```
Consider extracting a small helper at the top of the `describe` block:
```js
async function freshVariantsMock() {
  const { getProductVariants } = await import('public/cartService');
  getProductVariants.mockClear();
  return getProductVariants;
}
```
Then each test becomes `const mock = await freshVariantsMock();` -- one line instead of two, and the intent ("start with a clean mock") is clearer.

**2. Repeated variant response shape**

The object `{ variant: { price: 599 }, inStock: true, imageSrc: null, mediaItems: [] }` appears in at least 5 tests with minor variations. A factory function would reduce duplication and make intent clearer:
```js
function makeVariant(overrides = {}) {
  return { variant: { price: 599 }, inStock: true, imageSrc: null, mediaItems: [], ...overrides };
}
```
Then `mockResolvedValueOnce([makeVariant({ inStock: false })])` reads better than the full object literal each time.

**3. Repeated dropdown setup**

Multiple tests set `$w('#sizeDropdown').value = 'Full'` and `$w('#finishDropdown').value = ''`. This could be a small helper or moved into a nested `beforeEach` for tests that share this setup.

---

### ups-shipping.test.js

**4. Repeated `__setHandler` mock response boilerplate**

The OAuth token handler block appears identically in 5+ tests:
```js
if (url.includes('/oauth/token')) {
  return { ok: true, async json() { return { access_token: 'tok', expires_in: '3600' }; }, async text() { return ''; } };
}
```
A helper would reduce this significantly:
```js
function oauthResponse() {
  return { ok: true, async json() { return { access_token: 'tok', expires_in: '3600' }; }, async text() { return ''; } };
}
function emptyOkResponse() {
  return { ok: true, async json() { return {}; }, async text() { return ''; } };
}
```
Then each `__setHandler` callback becomes shorter and the test-specific logic stands out more clearly.

**5. `createShipment` tests share nearly identical input objects**

The tests for "defaults to Ground service", "handles missing addressLine2", and "includes billing weight" all pass similar shipment objects differing only in address details that are irrelevant to the assertion. A `makeShipmentInput(overrides)` factory would make each test's intent obvious by showing only the fields that matter.

**6. `getPackageDimensions` -- consider `it.each`**

The four new dimension tests follow the exact same pattern (call with category string, assert length and weight). This is a textbook case for `it.each`:
```js
it.each([
  ['futon-mattress', 78, 55],
  ['platform-bed', 80, 70],
  ['casegoods', 36, 45],
  ['accessory', 24, 15],
])('returns %s dimensions (length=%i, weight=%i)', (category, length, weight) => {
  const dims = getPackageDimensions(category);
  expect(dims.length).toBe(length);
  expect(dims.weight).toBe(weight);
});
```
This eliminates 16 lines of repetition and makes it trivial to add new categories later.

---

### Summary

All suggestions are about reducing repetition through small helpers and `it.each`. None change test behavior or coverage. The tests themselves are correct and well-targeted -- these are polish-level improvements for maintainability.

No blocking issues found.

DreadPirateRobertz:

## Summary

This PR meaningfully deepens test coverage for two critical modules: `ProductOptions.js` (variant selection, swatch rendering) and `ups-shipping.web.js` (rate calculation, shipment creation, tracking, address validation). The new tests are well-structured, test behavior rather than implementation details, and cover important error paths and edge cases. Overall quality is strong.

## Critical Gaps

**1. `selectSwatch` — case-insensitive match with no finish options set** (Criticality: 8/10)

When `$w('#finishDropdown').options` is undefined/null (not an empty array), the `.find()` call on line 191 of `ProductOptions.js` would throw. There is no test for when `finishDropdown.options` is falsy rather than an empty array. The existing test "applies tint fallback when no matching variant" sets `options` to an array with one item. If the page renders before options are populated, this could crash swatch selection silently (caught by the outer try/catch, but the user gets no visual feedback).

**2. `getUPSRates` — free shipping threshold test is indirect** (Criticality: 8/10)

The existing test "does NOT return free shipping at $1200" verifies the current threshold behavior, but it tests that free shipping is *disabled* rather than testing the threshold boundary. If `FREE_SHIPPING_THRESHOLD` is misconfigured (e.g., set to `Infinity` or `0`), neither direction is caught. A test that passes `orderSubtotal` at exactly the threshold value and verifies `free-ground` is returned would guard the checkout's free-shipping flow. This is revenue-impacting behavior.

**3. `createShipment` — return label address swap verification** (Criticality: 8/10)

The "creates return label when returnLabel is true" test only checks `result.success === true`. It does not verify that ShipTo/ShipFrom are actually swapped (ShipTo should be the origin, ShipFrom should be the customer). Since the mock returns the same response regardless of payload, a bug that omits the swap would pass this test silently. A regression here means return labels ship packages to the wrong address, requiring manual intervention.

## Important Improvements

**4. `handleCustomVariantChange` — variant with `price: 0` and non-call-for-price product** (Criticality: 6/10)

The call-for-price test sets `state.product.price = 0`, which triggers the `isCallForPrice` check. But there is no test for a variant where `variant.price` is `0` but the *product* itself is not call-for-price (e.g., a free accessory). In `updateVariantDisplay`, the condition `selected.variant?.price && !isCallForPrice(...)` would skip the price update because `0` is falsy, leaving stale text. This is a latent display bug.

**5. `trackShipment` — `estimatedDelivery` when `deliveryDate` is empty array** (Criticality: 5/10)

The partial-location test sets `deliveryDate: []` but does not assert `estimatedDelivery` is `null`. Worth a one-line assertion to verify this edge case is handled (delivered packages sometimes have no future delivery date).

**6. `getPackageDimensions` — `undefined`/`null` category** (Criticality: 5/10)

Tests cover unknown string categories but not `undefined` or `null`. Since this is called from the shipping-rates-plugin when Wix line items lack category data, `getPackageDimensions(undefined)` should return defaults. The current code handles it (`PACKAGE_DEFAULTS[undefined]` is `undefined`, falls back to default), but a test would document and protect this behavior.

**7. No tests for `openSwatchGallery` / `showSwatchDetail` / Escape-key handling** (Criticality: 4/10)

The full swatch gallery modal (search, Escape key close, ARIA attributes, detail panel) has no coverage. These are internal functions not exported, so testing them requires either exporting or testing through the `swatchViewAll` onClick callback. Lower priority since the modal is a progressive enhancement, but the Escape handler cleanup (`removeEventListener`) is a potential memory leak if broken.

## Test Quality Issues

**1. Several `createShipment` tests are redundant with the happy-path test.** The "defaults to Ground service" and "handles missing optional addressLine2" tests both just assert `result.success === true` against the same mock that always returns success. They do not verify the *payload* sent to UPS, so they cannot actually catch the bugs they describe. They would be more valuable if they captured the fetch call arguments and verified the request body.

**2. `getPackageDimensions` tests are exhaustive but low-value.** Testing every category in `PACKAGE_DEFAULTS` is essentially testing a lookup table. The original 3 tests (futon-frame, murphy-bed, unknown) adequately covered the logic. The 4 new category tests (futon-mattress, platform-bed, casegoods, accessory) add bulk without catching any additional code path. They would only catch a typo in the constant, which is better caught by a snapshot or schema validation.

**3. `ups-shipping` package cap test does not verify the cap was applied.** The "caps packages array at 20" test sends 25 packages but only asserts `rates.length > 0`. It should capture the fetch call and verify `Package` array length is 20, otherwise the test passes even if the cap is removed.

## Positive Observations

- **Error handling coverage is thorough.** Both modules test API failures, network errors, invalid responses, and null inputs. The `trackShipment` test that verifies internal error details are not leaked to callers (line 486-498) is particularly good security practice.
- **Regional fallback rate tiers are fully covered.** NC/SC, Southeast, Northeast, West Coast, and default tiers each have a dedicated test, which protects checkout from showing wrong estimates during UPS outages.
- **ProductOptions tests exercise the `onItemReady` callback pattern well.** Extracting the callback from mock calls and invoking it with synthetic data is the right approach for testing Wix repeater/grid rendering without a DOM.
- **The `NoCandidatesIndicator` test fills a real gap** in address validation coverage (previously only valid and ambiguous paths were tested).
- **Tests use `mockResolvedValueOnce` / `mockRejectedValueOnce` correctly**, avoiding test pollution between cases.

Overall, this is a solid coverage expansion. The critical gaps identified above are worth addressing before merge, particularly the return-label address swap verification and the free-shipping threshold boundary test — both protect revenue-impacting checkout flows.

PR #365, opened 2026-03-15
test: deepen ProductDetails + AddToCart coverage

DreadPirateRobertz:
## Summary
- **ProductDetails.js**: 17 → 49 tests (+32). Covers accordion toggle/ARIA, breadcrumb null guards, social share guards, delivery zip input, SEO FAQ schema injection, swatch request modal/submit/option matching, and `initSwatchCTA` (previously 0 tests)
- **AddToCart.js**: 23 → 40 tests (+17). Covers quantity ARIA labels, sticky bar click handler, stock urgency boundary tests (0/1/4/5/null), back-in-stock dropdown listeners, bundle image/price/navigation handlers
- Full suite: 14,290 tests passing across 368 files

## Test plan
- [x] `npx vitest run tests/productDetails.test.js` — 49/49 passing
- [x] `npx vitest run tests/addToCart.test.js` — 40/40 passing
- [x] Full suite `npx vitest run` — 14,290/14,290 passing

### DIFF
--- a/tests/addToCart.test.js
+++ b/tests/addToCart.test.js
@@ -96,6 +96,17 @@ describe('AddToCart', () => {
       minusCb();
       expect(state.selectedQuantity).toBe(1);
     });
+
+    it('registers onInput handler on quantity input', () => {
+      initQuantitySelector($w, state);
+      expect($w('#quantityInput').onInput).toHaveBeenCalled();
+    });
+
+    it('sets aria labels on plus and minus buttons', () => {
+      initQuantitySelector($w, state);
+      expect($w('#quantityMinus').accessibility.ariaLabel).toBe('Decrease quantity');
+      expect($w('#quantityPlus').accessibility.ariaLabel).toBe('Increase quantity');
+    });
   });
 
   describe('initAddToCartEnhancements', () => {
@@ -139,6 +150,11 @@ describe('AddToCart', () => {
       initStickyCartBar($w, state);
       expect(wixWindow.onScroll).toHaveBeenCalled();
     });
+
+    it('registers click handler on sticky add button', () => {
+      initStickyCartBar($w, state);
+      expect($w('#stickyAddBtn').onClick).toHaveBeenCalled();
+    });
   });
 
   describe('initBundleSection', () => {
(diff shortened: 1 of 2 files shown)


DreadPirateRobertz (commented):

Reviewed the +328/-2 diff across `tests/productDetails.test.js` and `tests/addToCart.test.js`. These are well-structured tests with good coverage of null guards, boundary conditions, and ARIA attributes. A few simplification opportunities:

### 1. Duplicate stock-urgency "hide" tests (`addToCart.test.js`)

The pre-existing test `'hides urgency when stock exceeds threshold'` (sets `quantityInStock = 10`) and the new `'hides urgency at boundary (stock = 5)'` both verify the same hide behavior for above-threshold stock. The boundary test at 5 is the more valuable one since it tests the exact threshold. Consider removing the `stock = 10` test — it adds no coverage beyond what `stock = 5` already proves.

Similarly, the new `'hides urgency when stock is 0'` test and the pre-existing `'hides when out of stock'` test may overlap if the existing test also sets stock to 0. Worth verifying they aren't duplicates.

### 2. Repeated `initBundleSection` calls could use a shared setup

Six of the seven new `initBundleSection` tests call `await initBundleSection($w, state)` with the default happy-path state. A nested `describe` with a shared `beforeEach` would reduce repetition:

```js
describe('with valid bundle product', () => {
  beforeEach(async () => { await initBundleSection($w, state); });
  
  it('sets bundle image and alt text', () => { ... });
  it('displays bundle price and savings', () => { ... });
  // etc.
});
```

This keeps the guard/null tests (which mutate state before calling) outside the nested block.

### 3. Mock inline formatting (`productDetails.test.js`, line 37)

The existing `engagementTracker` mock packs 5 exports onto one line. The new mocks (`a11yHelpers`, `designTokens`, `DeliveryEstimator`, `validators`) are properly formatted with readable structure — nice. But the contrast with the existing one-liner is jarring. If touching this file again, consider reformatting the `engagementTracker` mock to match the multi-line style of the new mocks for consistency.

### 4. Minor: test description precision

- `'does not throw when schema is null'` appears in both `initBreadcrumbs` and `injectProductSchema` — the assertions differ (one checks `resolves.toBeUndefined()`, the other also checks `resolves.toBeUndefined()`). Both are valid but the description could be more specific: `'resolves without posting when breadcrumb schema is null'` vs `'resolves without posting when product schema is null'`.

### 5. No nested ternaries or arrow-function concerns

The new code correctly uses `function` style for `createMockElement` and follows project conventions. No issues with naming, import order, or patterns.

### Summary

This is solid test code. The suggestions above are minor polish — the main actionable item is checking for duplicate stock-urgency hide tests (#1) and considering the shared `beforeEach` pattern for the bundle tests (#2) to reduce boilerplate.

DreadPirateRobertz (commented):
**code-simplifier** (dispatched by miquella)

## Simplification & Duplication Analysis

### Duplication Found

1. **`createMockElement()` is duplicated across both test files**: `addToCart.test.js` and `productDetails.test.js` each define their own `createMockElement()` and `create$w()`. The diff makes them diverge further — productDetails now includes `onKeyPress`, `onInput`, `focus`, and `style.backgroundColor` while addToCart has `getBoundingRect`. These should ideally be extracted to a shared `tests/helpers/mockElement.js` factory. This is a pre-existing pattern, not introduced by this PR, but the divergence is growing.

2. **Repeated null-guard test pattern**: Nearly every describe block adds a `'returns early when product is null'` test with the same structure:
   ```js
   it('returns early when product is null', () => {
     state.product = null;
     someFunction($w, state);
     expect($w('#someElement').someMethod).not.toHaveBeenCalled();
   });
   ```
   This pattern appears 7 times across the two files. A shared helper like `testNullProductGuard(fn, $w, state, selector, method)` could reduce boilerplate. However, the current explicit form is more readable and debuggable — this is a judgment call, not a blocker.

### Simplification Opportunities

3. **Bundle section tests could share setup**: The 7 new bundle tests all call `await initBundleSection($w, state)`. The describe block could use a `beforeEach` that runs the init, with only the override tests (null product, null bundle) setting up their own state first. This would reduce 5 lines of repetitive `await initBundleSection($w, state)` calls.

4. **Stock urgency tests are well-structured but could use a table**: The 6 boundary tests map naturally to `it.each`:
   ```js
   it.each([
     [0, 'hide'], [1, 'show'], [3, 'show'], [4, 'show'], [5, 'hide'], [null, 'hide'],
   ])('stock=%s -> %s urgency', async (stock, expected) => { ... });
   ```
   This would reduce 36 lines to ~10. However, the current form has better failure messages and is easier to debug. Suggestion only.

### What Is Already Simple

- Test assertions are direct and un-nested — no helper chains or complex matchers
- Mock setup is top-level and static — no dynamic mock generation
- State setup in `beforeEach` is minimal and clear
- No unnecessary abstractions or over-engineered test utilities

### Verdict

No critical simplification needed. The code is readable and maintainable. The main long-term concern is the diverging `createMockElement` factories across test files — consider extracting to a shared helper in a future cleanup PR. No blockers.

PR #372, opened 2026-03-16
test: deepen Home page tests (28 → 71)

DreadPirateRobertz:
## Summary
- Deepen Home page test coverage from 28 → 71 tests (+43 new)
- Test-to-complexity ratio improves from 38.5 → 15.2
- Covers 12 previously untested sections: trust bar, newsletter, video showcase, quiz CTA, smooth scroll, hero details, page init, featured quick view, testimonials deeper, category repeater, featured products deeper

## Test plan
- [x] `npx vitest run tests/homePage.test.js` — 71/71 pass
- [ ] Full suite regression — running

### DIFF
--- a/tests/homePage.test.js
+++ b/tests/homePage.test.js
@@ -1,6 +1,4 @@
 import { describe, it, expect, vi, beforeAll, beforeEach } from 'vitest';
-import { futonFrame, futonMattress, wallHuggerFrame, saleProduct } from './fixtures/products.js';
-
 // ── $w Mock Infrastructure ──────────────────────────────────────────
 
 const elements = new Map();
@@ -14,7 +12,9 @@ function createMockElement() {
     label: '',
     options: [],
     data: [],
-    style: { color: '' },
+    style: { color: '', backgroundColor: '' },
+    accessibility: { ariaLabel: '', ariaLive: '', role: '' },
+    background: { src: '' },
     show: vi.fn(() => Promise.resolve()),
     hide: vi.fn(() => Promise.resolve()),
     collapse: vi.fn(),
@@ -27,11 +27,16 @@ function createMockElement() {
     onItemClicked: vi.fn(),
     onReady: vi.fn(() => Promise.resolve()),
     onCurrentIndexChanged: vi.fn(),
+    onMouseIn: vi.fn(),
+    onMouseOut: vi.fn(),
     getCurrentItem: vi.fn(),
     getTotalCount: vi.fn(() => 0),
     getItems: vi.fn(() => ({ items: [] })),
     setSort: vi.fn(),
     setFilter: vi.fn(),
+    disable: vi.fn(),
+    enable: vi.fn(),
+    next: vi.fn(),
   };
 }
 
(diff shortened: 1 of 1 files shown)


DreadPirateRobertz:
## Silent Failure Hunter Review — PR #372 (Home Page Tests 28 -> 71)

**Reviewer**: silent-failure-hunter (dispatched by miquella)

---

### Overall Assessment

This PR adds 43 new tests covering 12 previously untested sections of the Home page. The test expansion itself is valuable. However, I have concerns about what the tests **do not** cover and about how the mock infrastructure interacts with the production code's error handling patterns.

After reviewing both the diff and the production `Home.js`, here are my findings:

---

### CRITICAL: `prioritizeSections` mock silently drops the `onError` callback

**Location**: `tests/homePage.test.js`, lines ~107-122 (the `prioritizeSections` mock)

**Issue**: The production code passes an `onError` callback to `prioritizeSections` (Home.js line 79-90) that is the **primary error reporting mechanism** for all deferred sections. The test mock ignores this callback entirely — it accepts `(sections)` but the production signature is `(sections, { onError })`. The mock never invokes `onError`, meaning:

1. There is no test verifying that section failures are reported to `errorMonitoring.web`.
2. There is no test verifying that the `onError` callback produces a well-formed `logError` payload with `page`, `context`, and `severity`.
3. If someone removes or breaks the `onError` wiring in production, no test will catch it.

**User Impact**: If `onError` handling breaks, section failures on the Home page will be completely silent in Sentry. Debugging production issues becomes guesswork.

**Recommendation**: Add at least one test that makes a deferred section throw and verifies `logError` is called with the correct payload structure including `page: 'Home'` and `severity: 'warning'`.

---

### HIGH: No tests for error/failure paths of any section

**Issue**: All 71 tests exercise the happy path. The production code has explicit error handling in `loadFeaturedProducts` (line 212-215), `loadSaleHighlights` (line 373-376), `initTestimonials` (line 644-646 CMS fallback), `initRecentlyViewed` (line 556-559), and the Quick View add-to-cart (line 265-269). None of these error branches are tested.

Specific untested error scenarios:
- `getFeaturedProducts` rejecting — does the skeleton hide? Does `console.error` fire?
- `getSaleProducts` rejecting — does the skeleton hide? Does the sale section collapse?
- Newsletter submit with a valid email but `subscribeToNewsletter` rejecting — what does the user see?
- `addToCart` failing in Quick View — does the button re-enable and label restore?
- CMS testimonials failing — does the fallback actually render?

**User Impact**: If any of these error handlers regress, users will see broken UI states (stuck skeletons, permanently disabled buttons, blank sections) with no test to catch it.

**Recommendation**: Add error-path tests for at least the data-fetching functions (`getFeaturedProducts`, `getSaleProducts`, `subscribeToNewsletter`). These are the most likely production failure points.

---

### HIGH: Testimonial fallback test uses weak conditional assertion

**Location**: `tests/homePage.test.js`, new test "populates testimonial repeater with fallback data when CMS returns empty"

```js
expect(repeater.data).toBeDefined();
if (repeater.data && repeater.data.length > 0) {
  expect(repeater.data[0].story).toBeTruthy();
}
```

**Issue**: The `if` guard means this test passes even when `repeater.data` is empty or undefined. The assertion inside the `if` block may never execute. This is a silent test pass — the test claims to verify fallback behavior but could be verifying nothing. The mock for `getFeaturedTestimonials` returns `{ success: true, items: [] }`, which means the production code SHOULD use `FALLBACK_TESTIMONIALS` (since `items.length === 0`). The test should assert that explicitly.

**Recommendation**: Replace the conditional with a direct assertion:
```js
expect(repeater.data).toHaveLength(4); // FALLBACK_TESTIMONIALS has 4 entries
expect(repeater.data[0].story).toContain('Night & Day');
```

---

### HIGH: Category repeater test uses weak conditional assertion

**Location**: `tests/homePage.test.js`, new test "sets category repeater data"

```js
expect(repeater.data).toBeDefined();
if (repeater.data) {
  expect(repeater.data.length).toBeLessThanOrEqual(6);
}
```

**Issue**: Same pattern — the `if` guard means this passes when `repeater.data` is undefined (since `toBeDefined` would pass for `[]` but the conditional still weakens confidence). The length assertion is also very loose — `0` passes. This does not actually verify that categories were loaded.

**Recommendation**: Assert the exact expected count and verify at least one category name:
```js
expect(repeater.data).toHaveLength(6);
expect(repeater.data[0].name).toBe('Futon Frames');
```

---

### MEDIUM: `logError` mock is wired but never asserted

**Location**: `tests/homePage.test.js`, line ~177

```js
vi.mock('backend/errorMonitoring.web', () => ({
  logError: vi.fn(),
}));
```

**Issue**: `logError` is mocked (good) but never appears in any assertion across all 71 tests. This mock exists solely to prevent import errors, not to verify error reporting. Since `logError` is the production error reporting mechanism (Sentry), at least one test should verify it gets called when things go wrong.

**Recommendation**: Add a test that triggers a section failure and asserts `logError` was called. Import and reference the mock in an assertion.

---

### MEDIUM: Newsletter error message test only covers invalid email

**Location**: "shows error for invalid email on submit" test

**Issue**: The test verifies the client-side email validation error message. However, the production code also has a server-side failure path when `subscribeToNewsletter` rejects. That path likely shows a different error message to the user, and it is not tested. If the server error message is generic or missing, users will not know why their subscription failed.

**Recommendation**: Add a test where `subscribeToNewsletter` rejects and verify the user sees an appropriate error message (not a blank or generic one).

---

### MEDIUM: Repeated `makeClickable.mockClear()` pattern is fragile

**Location**: Multiple tests clear `makeClickable` before running `onReadyHandler`, but some do not.

**Issue**: Tests that do NOT clear `makeClickable` before their assertions may be counting calls from previous test runs if `beforeEach` does not reset the mock map fully. The `elements.clear()` in `beforeEach` creates new mock elements, but `makeClickable` accumulates calls across tests unless explicitly cleared. This could lead to false positives (test passes because of calls from a prior test) or false negatives (counts are wrong).

**Recommendation**: Move `makeClickable.mockClear()` into the `beforeEach` block alongside `elements.clear()` so every test starts clean.

---

### LOW: Fixture inline duplication vs. shared fixtures

**Location**: `tests/homePage.test.js`, lines ~57-64 (the `vi.hoisted` block)

**Issue**: The PR removes the import of shared fixtures (`products.js`) and inlines product data. The comment says "to avoid hoisting issues with vi.mock factories." This is a legitimate Vitest constraint, but the inline fixtures now duplicate product shape definitions that exist in `tests/fixtures/products.js`. If the product shape changes (e.g., a new required field), only one location will be updated.

**Recommendation**: This is acceptable as-is given the hoisting constraint, but add a brief comment noting which fixture file these mirror so future maintainers know to keep them in sync.

---

### Summary

| Severity | Count | Key Theme |
|----------|-------|-----------|
| CRITICAL | 1 | `onError` callback not tested — Sentry blind spot |
| HIGH | 3 | No error-path tests; weak conditional assertions |
| MEDIUM | 3 | `logError` never asserted; newsletter server error untested; mock clearing fragile |
| LOW | 1 | Fixture duplication |

The happy-path coverage expansion is solid work. The main gap is that **none of the production error handling is tested**, which means the error paths that matter most in production (data fetch failures, CMS outages, cart errors) could regress undetected. I recommend addressing at least the CRITICAL and HIGH items before merge.

---
*silent-failure-hunter (dispatched by miquella)*

PR #374, opened 2026-03-16
test: deepen searchService.web.js tests (103 → 170)

DreadPirateRobertz:
## Summary
- **+67 new tests** for `searchService.web.js` (103 → 170), deepening coverage across all 6 exported functions
- Covers cache eviction, validateSlug edge cases, buildFacets with missing/null fields, fullTextSearch combined filters, autocomplete suggestion types/ordering, query frequency tracking, sanitize-to-empty inputs, and pagination boundaries
- All 14,664 tests pass across 369 files — zero regressions

## Test plan
- [x] All new tests pass: `npx vitest run tests/searchService.test.js` (170 pass)
- [x] Full suite passes: `npx vitest run` (14,664 pass)
- [x] Confirmed no open PRs touch searchService files
- [x] No production code changes

### DIFF
--- a/tests/searchService.test.js
+++ b/tests/searchService.test.js
@@ -809,10 +809,14 @@ describe('search cache', () => {
   });
 
   it('different offset params produce different cache entries', async () => {
-    const r1 = await fullTextSearch({ query: 'Futon', offset: 0 });
-    const r2 = await fullTextSearch({ query: 'Futon', offset: 1 });
-    // These should be separate cache entries (different keys)
-    expect(r1.products.length).toBeGreaterThan(0);
+    await fullTextSearch({ query: 'Futon', offset: 0 });
+    __seed('Stores/Products', []);
+    // Same query+offset=0 should hit cache and return non-empty
+    const cached = await fullTextSearch({ query: 'Futon', offset: 0 });
+    expect(cached.products.length).toBeGreaterThan(0);
+    // Different offset should miss cache and return empty (data is gone)
+    const missed = await fullTextSearch({ query: 'Futon', offset: 1 });
+    expect(missed.products).toHaveLength(0);
   });
 });
 
@@ -1072,8 +1076,10 @@ describe('fullTextSearch (extended)', () => {
   });
 
   it('clamps negative offset to 0', async () => {
-    const result = await fullTextSearch({ query: 'Futon', offset: -5 });
-    expect(result.products.length).toBeGreaterThan(0);
+    const noOffset = await fullTextSearch({ query: 'Futon', offset: 0 });
+    __clearCache();
+    const negOffset = await fullTextSearch({ query: 'Futon', offset: -5 });
+    expect(negOffset.products.length).toBe(noOffset.products.length);
   });
 
   it('clamps NaN limit to default 24', async () => {
@@ -1082,16 +1088,21 @@ describe('fullTextSearch (extended)', () => {
     expect(result.products.length).toBeLessThanOrEqual(24);
   });
 
-  it('clamps limit minimum to 1', async () => {
+  it('treats limit=0 as default 24 (0 is falsy)', async () => {
     const result = await fullTextSearch({ query: 'Futon', limit: 0 });
-    expect(result.products.length).toBeGreaterThanOrEqual(1);
+    // Number(0) || 24 = 24 — 0 is falsy, default 24 kicks in
+    // All futon matches returned since 24 > match count
+    const allResults = await fullTextSearch({ query: 'Futon', limit: 100 });
+    expect(result.products.length).toBe(allResults.products.length);
   });
 
   it('sorts by relevance by default (name matches before description)', async () => {
-    // 'Premium' in p6 name AND description. 'storage' only in p6 description + p4 featureTags
-    const result = await fullTextSearch({ query: 'storage' });
-    // First result should be description match (higher relevance)
+    // 'Premium' appears in p6 name and p6 description. Name match gets _relevance=2.
+    const result = await fullTextSearch({ query: 'Premium' });
     expect(result.products.length).toBeGreaterThan(0);
+    // First result should be the name-matched product
+    expect(result.products[0].name).toContain('Premium');
+    expect(result.products[0].slug).toBe('premium-plat');
   });
 
   it('records query for popular tracking', async () => {
@@ -1172,12 +1183,14 @@ describe('getAutocompleteSuggestions (extended)', () => {
   it('treats limit=0 as default 8 (0 is falsy)', async () => {
     const result = await getAutocompleteSuggestions('Fu', 0);
     // Number(0) || 8 = 8 — 0 is falsy so default kicks in
+    expect(result.suggestions.length).toBeGreaterThan(0);
     expect(result.suggestions.length).toBeLessThanOrEqual(8);
   });
 
   it('handles NaN limit gracefully', async () => {
     const result = await getAutocompleteSuggestions('Fu', 'abc');
     // NaN → default 8
+    expect(result.suggestions.length).toBeGreaterThan(0);
     expect(result.suggestions.length).toBeLessThanOrEqual(8);
   });
 
@@ -1186,11 +1199,11 @@ describe('getAutocompleteSuggestions (extended)', () => {
     expect(result.suggestions.some(s => s.type === 'category' && s.text === 'Platform Beds')).toBe(true);
   });
 
-  it('returns multiple category matches', async () => {
-    // 'fu' matches 'Futon Frames' and 'Front Loading & Nesting'
+  it('returns category match for Futon Frames', async () => {
     const result = await getAutocompleteSuggestions('fu', 20);
     const cats = result.suggestions.filter(s => s.type === 'category');
     expect(cats.length).toBeGreaterThanOrEqual(1);
+    expect(cats.some(c => c.text === 'Futon Frames')).toBe(true);
   });
 });
 


DreadPirateRobertz (commented):
**Code Simplifier Agent** (dispatched by radahn)

Good test coverage expansion. The 67 new tests are well-structured and test real edge cases. Below are simplification opportunities — all preserve existing functionality.

---

### 1. Consolidate "no-op filter" tests with `it.each`

Several `searchProducts` tests follow an identical pattern: pass a degenerate filter value, assert all 7 products returned. These can collapse into a single parameterized test:

```js
it.each([
  ['special-char category', { category: 'futon<frames>' }],
  ['empty-after-sanitize material', { material: '<>' }],
  ['empty-after-sanitize color', { color: '<>' }],
  ['wrong-length depthRange (1)', { depthRange: [30] }],
  ['wrong-length depthRange (3)', { depthRange: [30, 40, 50] }],
  ['NaN widthRange values', { widthRange: ['abc', 'def'] }],
  ['null category', { category: null }],
  ['undefined category', { category: undefined }],
])('ignores no-op filter: %s', async (_label, filters) => {
  const result = await searchProducts(filters);
  expect(result.products).toHaveLength(7);
});
```

This replaces **8 separate tests** with one parameterized block — same coverage, far less repetition. The `offset beyond total` test is different (asserts 0 products + total=7) so it stays standalone.

Similarly in `fullTextSearch (extended)`, these share the "compare to unfiltered total" pattern and can be collapsed:

- `handles material filter with empty sanitize result`
- `handles color filter with empty sanitize result`
- `handles invalid priceRange key in fullTextSearch`
- `handles category with special chars in fullTextSearch`

```js
it.each([
  ['empty-sanitize material', { material: '<>' }],
  ['empty-sanitize color', { color: '<>' }],
  ['invalid priceRange', { priceRange: 'bogus' }],
  ['special-char category', { category: 'bad<cat>' }],
])('no-op filter in fullTextSearch: %s', async (_label, filters) => {
  const all = await fullTextSearch({ query: 'Futon' });
  const filtered = await fullTextSearch({ query: 'Futon', ...filters });
  expect(filtered.total).toBe(all.total);
});
```

### 2. Consolidate `getFilterValues` null-field tests

The three buildFacets tests for `null material`, `null color`, and `null featureTags` each seed a near-identical single-item dataset. They could share a helper or use `it.each`:

```js
it.each([
  ['materials', { material: null, color: 'Red', featureTags: [] }],
  ['colors', { material: 'Wood', color: null, featureTags: [] }],
  ['features', { material: 'Wood', color: 'Red', featureTags: null }],
])('items with null %s are excluded from facets', async (facetKey, overrides) => {
  __seed('Stores/Products', [
    { _id: 'x1', name: 'Test', price: 100, width: 30, depth: 30, collections: [], ...overrides },
  ]);
  const facets = await getFilterValues();
  expect(facets[facetKey]).toHaveLength(facetKey === 'materials' ? 0 : facetKey === 'colors' ? 0 : 0);
});
```

Though honestly the existing form is readable enough — this is a **minor** suggestion.

### 3. Consolidate `getFilterValues (extended)` no-op category tests

Three tests assert `facets.totalProducts === 7` for special-char, empty-string, and null categories:

```js
it.each([
  ['special chars', 'bad<cat>'],
  ['empty string', ''],
  ['null', null],
])('treats %s category as all products', async (_label, category) => {
  const facets = await getFilterValues(category);
  expect(facets.totalProducts).toBe(7);
});
```

### 4. Consolidate autocomplete limit-edge-case tests

`treats limit=0 as default 8` and `handles NaN limit gracefully` both assert `length <= 8`:

```js
it.each([
  ['0 (falsy)', 0],
  ['NaN string', 'abc'],
])('defaults limit to 8 for %s', async (_label, limit) => {
  const result = await getAutocompleteSuggestions('Fu', limit);
  expect(result.suggestions.length).toBeLessThanOrEqual(8);
});
```

### 5. Weak assertion in `different offset params produce different cache entries`

```js
it('different offset params produce different cache entries', async () => {
  const r1 = await fullTextSearch({ query: 'Futon', offset: 0 });
  const r2 = await fullTextSearch({ query: 'Futon', offset: 1 });
  expect(r1.products.length).toBeGreaterThan(0);
});
```

`r2` is never asserted. The test title claims it verifies different cache entries but only checks `r1`. Either add a meaningful assertion on `r2` (e.g., confirm different products or different lengths) or document why `r2` existence alone is sufficient. As written, removing the `r2` line changes nothing.

### 6. Defensive guard weakens `categories appear before products` test

```js
if (catIdx >= 0 && prodIdx >= 0) {
  expect(catIdx).toBeLessThan(prodIdx);
}
```

If neither index is found, the test silently passes with zero assertions. Consider failing explicitly:

```js
expect(catIdx).toBeGreaterThanOrEqual(0);
expect(prodIdx).toBeGreaterThanOrEqual(0);
expect(catIdx).toBeLessThan(prodIdx);
```

### 7. `empty queries are not tracked` duplicates `short queries (< 2 chars)`

The `query frequency tracking` describe has both `empty queries are not tracked` and `short queries (< 2 chars) are not tracked`. An empty string is a subset of "< 2 chars." The empty-string test adds no unique coverage beyond what the short-query test already verifies. Consider removing it or merging them.

### 8. Minor: `sorts by relevance` test asserts nothing specific

```js
it('sorts by relevance by default (name matches before description)', async () => {
  const result = await fullTextSearch({ query: 'storage' });
  expect(result.products.length).toBeGreaterThan(0);
});
```

The title promises relevance ordering but only checks non-empty results. Either assert the actual ordering (first result slug, score ordering) or rename the test to `returns results for description-only matches`.

---

**Summary**: The test logic is sound and the edge cases are valuable. The main improvement opportunity is collapsing 15-20 repetitive no-op-filter tests into 4-5 `it.each` blocks. This would cut ~80 lines while keeping identical coverage. The weak assertions in items 5, 6, and 8 should be strengthened — tests that always pass regardless of behavior are worse than no tests because they create false confidence.

DreadPirateRobertz (commented):
**PR Test Analyzer Agent** (dispatched by radahn)

---

## Summary

This PR adds 67 tests to `tests/searchService.test.js`, deepening coverage of `src/backend/searchService.web.js` from ~103 to ~170 tests. The new tests systematically target edge cases, null/undefined inputs, sanitization bypass paths, cache key isolation, facet boundary conditions, and input clamping logic. Overall quality is strong -- the tests are behavior-focused, use clear naming, and cover real defensive-coding paths that matter for production resilience.

---

## Positive Observations

- **Sanitization edge cases are well covered**: Tests for `material: '<>'`, `color: '<>'`, `category: 'bad<cat>'` verify that empty-after-sanitize inputs correctly degrade to "no filter" rather than matching nothing or crashing. These are important XSS-adjacent validation paths.
- **Cache key isolation tests are valuable**: Verifying that different `category`, `offset`, and `sortBy` params produce separate cache entries prevents a class of subtle bugs where users see stale/wrong results.
- **Facet boundary tests with custom seed data**: The `buildFacets edge cases` block seeds items with `null` material, `null` color, `null` featureTags, string dimensions, and null prices. These exercise the defensive `if (item.material)`, `typeof w === 'number'`, and `item.price || 0` guards in `buildFacets()` -- exactly the paths most likely to cause runtime errors from bad CMS data.
- **Input clamping coverage**: `limit: 0`, `limit: 'abc'`, `offset: -5`, `depthRange: [30]`, `depthRange: [30, 40, 50]`, `widthRange: ['abc', 'def']` -- these verify the `Math.min/Math.max/Number()` clamping logic and are valuable regression guards.
- **recordSearchQuery normalization**: Tests for whitespace trimming, case folding, and single-char-after-sanitize rejection are well-targeted.

---

## Critical Gaps (None)

No critical (8-10) test gaps identified. The core business logic paths -- search, filter, sort, paginate, cache, facet computation, autocomplete, and query tracking -- are all covered with both happy path and edge case tests.

---

## Important Improvements

### 1. Search cache eviction (setCachedSearch MAX_SEARCH_CACHE_ENTRIES) -- Criticality: 7/10

The source has an eviction path at lines 66-76: when search cache entries exceed `MAX_SEARCH_CACHE_ENTRIES` (50), the oldest entry is deleted. No test exercises this eviction logic. A bug here (e.g., evicting the wrong key, off-by-one on the limit) would cause unbounded memory growth or premature cache misses in production.

**Suggested test**: Seed > 50 unique search queries, then verify (a) the cache does not grow beyond 50 entries, and (b) new queries still return correct results.

### 2. Query frequency eviction (recordQuery MAX_TRACKED_QUERIES) -- Criticality: 6/10

The internal `recordQuery` function at lines 90-98 evicts the least-frequent query when the tracker reaches 200 entries. This path is untested. A bug could cause memory leaks in the query tracker or loss of genuinely popular queries.

**Suggested test**: Record 201+ unique queries and verify the tracker stays bounded and evicts the least frequent entry.

### 3. Query decay filter (QUERY_DECAY_MS) -- Criticality: 6/10

`getTopQueries` at line 108 filters out queries older than 24 hours. This is untested. If the decay logic breaks, stale queries would persist forever in "trending searches," degrading UX.

**Suggested test**: Use `vi.useFakeTimers()` to advance time past 24h and verify that old queries are excluded from `getPopularSearches()` results.

### 4. fullTextSearch error/catch path -- Criticality: 5/10

The `try/catch` at lines 622-625 returns `{ products: [], total: 0, query: '', facets: {} }` on error. The existing test for `searchProducts` error shape only verifies the empty-data case, not an actual thrown error. Neither function's catch block is exercised with a genuine exception.

**Suggested approach**: Mock `wixData.query().find()` to throw, then verify the error return shape. This ensures the catch block actually works and does not itself throw.

### 5. Autocomplete cache TTL (AUTOCOMPLETE_TTL_MS) -- Criticality: 4/10

The autocomplete cache uses a separate 2-minute TTL (line 652). No test verifies that expired autocomplete entries are refreshed. Low risk since the facet cache TTL test pattern is established, but worth adding for completeness.

### 6. `searchProducts` does not include `description` in its field mapping -- Criticality: 3/10

The PR has a test `'does not include description in searchProducts mapped fields'` which verifies this, which is good. However, note that `searchProducts` uses an inline mapper (lines 259-277) while `fullTextSearch` uses the shared `mapProduct()` function (line 611) which DOES include `description`. This asymmetry is tested but could be confusing. The test correctly documents the intentional difference -- no action needed, just calling it out.

---

## Test Quality Issues

### Minor: Some assertions are weaker than they could be

A few new tests assert only `.length` or `.toBeGreaterThan(0)` when the seed data is deterministic enough to assert exact counts or specific product identities. Examples:

- `'different offset params produce different cache entries'` -- only asserts `r1.products.length > 0` but never verifies `r1` and `r2` are actually different entries.
- `'sorts by relevance by default'` -- asserts `length > 0` without verifying actual ordering.
- `'returns multiple category matches'` -- asserts `cats.length >= 1` when the exact count could be asserted.

These are not bugs, but tightening the assertions would make the tests more sensitive to regressions. **Criticality: 2/10** -- nice-to-have, not blocking.

### Minor: `getFilterValues` error shape not exercised

`getFilterValues` has a catch block (lines 309-318) returning a specific error shape with empty arrays and zeroed dimensions. No test triggers this path. **Criticality: 3/10**.

---

## Verdict

Good PR. The 67 new tests cover meaningful edge cases and defensive paths. The test density is appropriate for a search service that powers two user-facing pages. The three improvements rated 6-7 (cache eviction, query eviction, decay TTL) are worth considering in a follow-up but are not blockers for merge.

DreadPirateRobertz (commented):
**Comment Analyzer Agent** (dispatched by radahn)

---

**Summary**: Analyzed 67 new tests (+512 lines) in `tests/searchService.test.js`. The PR adds edge-case and missing-path coverage for `searchProducts`, `fullTextSearch`, `getAutocompleteSuggestions`, `getPopularSearches`, `recordSearchQuery`, `buildFacets`, `getFilterValues`, and query frequency tracking. Cross-referenced every inline comment against the source implementation in `src/backend/searchService.web.js` and `src/backend/utils/sanitize.js`. Found 1 factually incorrect comment, 1 misleading test name + comment pair, and 1 test with assertions too weak to verify its stated claim.

---

### Critical Issues

**1. Factually incorrect comment: 'fu' does NOT match 'Front Loading & Nesting'**

- Location: `tests/searchService.test.js`, new test `'returns multiple category matches'`
- Code comment: `// 'fu' matches 'Futon Frames' and 'Front Loading & Nesting'`
- Issue: The autocomplete code matches via `label.toLowerCase().includes(normalizedPrefix)`. The string `'front loading & nesting'` does NOT contain the substring `'fu'`. The prefix `'fu'` only matches `'Futon Frames'`. The comment is factually wrong.
- Impact: The assertion `cats.length >= 1` is weak enough to pass (it matches Futon Frames), so the test passes for the wrong reason. A future maintainer reading this comment would believe two categories match, which could cause confusion when debugging autocomplete behavior.
- Suggestion: Remove the incorrect claim. Change the comment to `// 'fu' matches 'Futon Frames'` and either strengthen the assertion to `expect(cats).toHaveLength(1)` or pick a prefix that genuinely matches multiple categories (e.g., no single prefix matches 2+ categories in the current CATEGORY_LABELS map, so this test may need a different approach such as custom seed data).

---

### Improvement Opportunities

**2. Misleading test name: "clamps limit minimum to 1" does not test the min-clamp**

- Location: `tests/searchService.test.js`, new test `'clamps limit minimum to 1'`
- Current state: The test passes `limit: 0`. In the source (`fullTextSearch` line 607), the expression is `Math.min(Math.max(1, Number(limit) || 24), 100)`. Since `Number(0)` is falsy, `0 || 24 = 24`, so the `||` default kicks in **before** `Math.max(1, ...)` ever has a chance to clamp. The limit becomes 24, not 1. The min-clamp to 1 is never exercised.
- Impact: The test name claims to verify that the minimum limit is clamped to 1, but it actually verifies the falsy-default path (0 -> 24). This is the same mechanism tested by the `'clamps NaN limit to default 24'` test, making this test redundant and its name misleading.
- Suggestion: To genuinely test the min-clamp, pass a truthy sub-minimum value like `limit: -1`. `Number(-1)` is truthy, so `-1 || 24 = -1`, then `Math.max(1, -1) = 1`, which actually exercises the clamp. Rename to `'clamps negative limit to minimum 1'` and assert `expect(result.products).toHaveLength(1)`.

**3. Test "different offset params produce different cache entries" has no meaningful assertion**

- Location: `tests/searchService.test.js`, new test `'different offset params produce different cache entries'`
- Current state: The test makes two calls with different offsets but only asserts `r1.products.length > 0`. It never checks that r2 differs from r1 or that both cache entries exist independently. The comment says "These should be separate cache entries (different keys)" but nothing in the assertions proves this.
- Suggestion: Use the same seed-mutation technique from the preceding test: call with offset=0, then `__seed('Stores/Products', [])`, then call with offset=0 again (should return cached non-empty), then call with offset=1 (should miss cache and return empty). This would actually prove the cache keys are distinct.

**4. Relevance sort test lacks meaningful assertion**

- Location: `tests/searchService.test.js`, new test `'sorts by relevance by default (name matches before description)'`
- Current state: The comment explains the relevance scoring (name match = 2, description match = 1) but the assertion is only `expect(result.products.length).toBeGreaterThan(0)`. It does not verify ordering.
- Suggestion: Assert that the first result is a name-match product. For query `'storage'`, p4 (Sagebrush) has 'storage' in `featureTags` (not name or description), and p6 (Premium Platform) and p7 (Clearance Ottoman) have 'storage' in description. Since `contains('name', 'storage')` won't match any product names in the seed data, all matches come from description with equal relevance. Consider using a query like `'Premium'` where p6 matches on name (relevance 2) and picking a second product that matches only on description, then assert `result.products[0].slug === 'premium-plat'`.

---

### Recommended Removals

None. All other comments add value by documenting the sanitization mechanism or cache behavior being tested. They serve as useful inline documentation for maintainers unfamiliar with the `||` default and `validateSlug` patterns.

---

### Positive Findings

- The `sanitize('<>') -> ''` comments (appearing in both `searchProducts` and `fullTextSearch` edge case tests) are accurate and helpful. They document a non-obvious transformation chain that future maintainers would need to understand.
- The `// Number(0) || 8 = 8 -- 0 is falsy so default kicks in` comment correctly documents a subtle JavaScript behavior that is the actual mechanism under test.
- The price boundary test comments (`// 299.99` and `// 300`) are precise and verified against the `PRICE_RANGES` definition where `'0-300'` has `max: 299.99` and `'300-500'` has `min: 300`.
- The `// sanitize strips tags: 'Eu' remains` comment on the autocomplete HTML-sanitization test is accurate and documents the important transformation step.
- Section header comments (e.g., `// -- searchProducts -- edge cases & missing paths`) provide good navigational structure across 500+ new lines.

PR #379, opened 2026-03-16
feat(CF-photo): N&D Image Bank download script

DreadPirateRobertz:
## Summary
- Download script (`scripts/download-nd-images.mjs`) authenticates with Night & Day Furniture dealer Image Bank and pulls hi-res product images
- Maps 12 image bank folders to catalog-MASTER.json slugs, downloads 54 hi-res images (90MB total, gitignored)
- Searched remaining 32 unmapped catalog frames — confirmed not available in image bank

## Coverage
| Metric | Value |
|--------|-------|
| Matched folders | 12 of 44 frame products |
| Images downloaded | 54 (1-7 per product, multiple finishes) |
| Total size | ~90MB |
| Unmapped frames | 32 (all have existing Wix-hosted images) |

## Test plan
- [x] Dry-run mode verified mapping and image counts
- [x] Full download completed with 0 errors
- [x] content/nd-images/ added to .gitignore (binary assets too large for git)
- [ ] Images should be uploaded to Wix Media Manager for staging site use

### DIFF
--- a/scripts/download-nd-images.mjs
+++ b/scripts/download-nd-images.mjs
@@ -3,7 +3,8 @@
  * Download hi-res product images from Night & Day Furniture dealer Image Bank.
  * Maps image bank folders to catalog-MASTER.json slugs and downloads all images.
  *
- * Usage: node scripts/download-nd-images.mjs [--dry-run]
+ * Usage: ND_USER=nightimage ND_PASS=bankpass2 node scripts/download-nd-images.mjs [--dry-run]
+ *   Or source secrets.env first: source scripts/secrets.env && node scripts/download-nd-images.mjs
  */
 
 import { readFileSync, mkdirSync, writeFileSync, existsSync } from 'fs';
@@ -13,7 +14,13 @@ import { fileURLToPath } from 'url';
 const __dirname = dirname(fileURLToPath(import.meta.url));
 const ROOT = join(__dirname, '..');
 
-const AUTH = Buffer.from('nightimage:bankpass2').toString('base64');
+const { ND_USER, ND_PASS } = process.env;
+if (!ND_USER || !ND_PASS) {
+  console.error('FATAL: Set ND_USER and ND_PASS environment variables.');
+  console.error('  Example: ND_USER=nightimage ND_PASS=bankpass2 node scripts/download-nd-images.mjs');
+  process.exit(1);
+}
+const AUTH = Buffer.from(`${ND_USER}:${ND_PASS}`).toString('base64');
 const BASE = 'https://imagebank.nightanddayfurniture.com/dealer';
 const DRY_RUN = process.argv.includes('--dry-run');
 


DreadPirateRobertz:
**Code Simplifier** (dispatched by godfrey)

Reviewed `scripts/download-nd-images.mjs` and `.gitignore` changes. Overall this is a clean, well-structured utility script. A few simplification opportunities:

### 1. Hardcoded credentials on line 18 (severity: high)

```js
const AUTH = Buffer.from('nightimage:bankpass2').toString('base64');
```

This embeds plaintext credentials directly in source code that will be committed to the repository. Even though the base64 encoding makes it slightly less obvious, the username and password are fully recoverable. This should read from an environment variable or a `.env` file (already gitignored via `secrets.env` pattern):

```js
const AUTH = Buffer.from(process.env.ND_IMAGE_BANK_AUTH ?? '').toString('base64');
```

With a guard at the top of `main()` that exits early if unset.

### 2. Duplicate HTML-fetching logic (lines 90-97 vs 130-132)

`processFolderPage` already wraps the fetch-and-parse pattern, but the extra-folders loop at line 130 duplicates the raw fetch+text+parse sequence inline. Extract the fetch step so both paths share it, or just reuse `fetchAuth` + `parseImages` consistently:

```js
const html = await fetchAuth(`${BASE}/imagebank.cgi?selected=${folderId}`).then(r => r.text());
const directImages = parseImages(html);
```

This already works for the subfolder scan since `parseImages` is called on line 140. The intermediate `res` variable on line 130 is unnecessary.

### 3. Regex with mutable `exec` loop can be replaced with `matchAll` (lines 58-62, 133-138)

The `while ((m = re.exec(html)))` pattern requires the regex to have the `g` flag and carries mutable state. `String.prototype.matchAll` is cleaner:

```js
function parseImages(html) {
  const re = /href="serve\.cgi\?id=(\d+)"[^>]*>([^<]+)/g;
  return [...html.matchAll(re)].map(m => ({ id: m[1], filename: m[2].trim() }));
}
```

Same applies to the subfolder regex on lines 133-138.

### 4. Status counting via reduce instead of manual mutation (lines 112-114)

The `if/else if` chain for counting `downloaded`/`skipped`/`errors` is spread across the loop body. Consider collecting all results first, then summarizing:

```js
const results = [];
for (const img of images) {
  // ... download logic, push to results
}
report.folders[slug] = { productName, imageCount: images.length, images: results };
```

Then derive counts at the end from the flat results array rather than incrementing counters mid-loop. This separates the "do work" phase from the "summarize" phase.

### 5. Extra-folders section does not download anything (lines 127-150)

The extra-folders scan lists subfolders and counts direct images but never downloads them and never adds to the report. If this is intentional reconnaissance, a brief comment explaining "scan only, no download" would clarify intent. If downloads are planned later, a `TODO` would be appropriate.

### 6. Minor: `'='.repeat(50)` used three times

A small thing, but a `const SEPARATOR = '='.repeat(50)` at the top would reduce repetition and make width changes a single-point edit.

### Summary

The script is well-organized and the `.gitignore` addition is correct. The credential issue (item 1) is the only one I would consider blocking. The rest are clarity improvements that would make the script easier to maintain and extend.

PR #380, opened 2026-03-16
feat(CF-vbj7): Product variant refactor — dropdown + visual swatches

DreadPirateRobertz:
## Summary
- Implements Discussion #273 consensus (Option 2): size dropdown + visual finish swatches
- Adds `initFinishSwatches()` to ProductOptions.js — renders Finish options as color circle swatches in `#finishSwatches` repeater
- Hidden `#finishDropdown` preserved for screen reader accessibility (progressive enhancement per rennala)
- Out-of-stock finishes grayed out (opacity 0.3) with "Special Order" tooltip, click disabled
- Auto-selects first in-stock finish on init
- Full ARIA: `role="radiogroup"`, `aria-label`, `aria-checked`, `aria-disabled`, visible focus ring (3px mountainBlue)

## What's NOT in this PR (future work)
- Keyboard arrow-key navigation within swatch group (needs `onKeyPress` Wix API investigation)
- Category/Home product card swatch strips (blocked on Repeater decision — resolved in #279)
- Editor element creation (`#finishSwatches` container) — Stilgar to add in next editor session

## Test plan
- [x] 27 new TDD tests covering initFinishSwatches (60 total in productOptions.test.js)
- [x] Full test suite passes: 14,767 tests across 369 files
- [ ] Manual: verify swatch rendering on Product Page after editor element added
- [ ] Manual: verify OOS gray-out with real inventory data
- [ ] Manual: verify screen reader announces finish options correctly

### DIFF
--- a/src/public/ProductOptions.js
+++ b/src/public/ProductOptions.js
@@ -133,16 +133,18 @@ export async function initFinishSwatches($w, state) {
       container.accessibility.ariaLabel = 'Finish';
     } catch (e) {}
 
-    // Check stock for each finish
+    // Check stock for each finish — default OOS on failure (safe direction)
     const swatchData = await Promise.all(
       finishOption.choices.map(async (choice, i) => {
-        let outOfStock = false;
+        let outOfStock = true;
         try {
           const variants = await getProductVariants(state.product._id, { Finish: choice.value });
           if (variants?.length > 0) {
             outOfStock = !variants[0].inStock;
           }
-        } catch (e) {}
+        } catch (e) {
+          console.error(`Stock check failed for finish "${choice.value}":`, e);
+        }
         return {
           _id: `finish-${i}`,
           value: choice.value,
--- a/tests/productOptions.test.js
+++ b/tests/productOptions.test.js
@@ -543,6 +543,48 @@ describe('ProductOptions', () => {
       expect($w('#finishSwatches').expand).toHaveBeenCalled();
     });
 
+    it('defaults to OOS when stock check fails (safe direction)', async () => {
+      const { getProductVariants } = await import('public/cartService');
+      getProductVariants.mockImplementation((_id, choices) => {
+        if (choices?.Finish === 'Walnut') return Promise.reject(new Error('timeout'));
+        return Promise.resolve([{ variant: { price: 599 }, inStock: true }]);
+      });
+      await initFinishSwatches($w, state);
+      const walnut = $w('#finishSwatches').data.find(d => d.value === 'Walnut');
+      expect(walnut.outOfStock).toBe(true);
+    });
+
+    it('defaults to OOS when stock check returns empty array', async () => {
+      const { getProductVariants } = await import('public/cartService');
+      getProductVariants.mockImplementation((_id, choices) => {
+        if (choices?.Finish === 'Cherry') return Promise.resolve([]);
+        return Promise.resolve([{ variant: { price: 599 }, inStock: true }]);
+      });
+      await initFinishSwatches($w, state);
+      const cherry = $w('#finishSwatches').data.find(d => d.value === 'Cherry');
+      expect(cherry.outOfStock).toBe(true);
+    });
+
+    it('handles all finishes out of stock — no default selection', async () => {
+      const { getProductVariants } = await import('public/cartService');
+      getProductVariants.mockResolvedValue([{ variant: { price: 599 }, inStock: false }]);
+      await initFinishSwatches($w, state);
+      // All OOS — dropdown should not be set to any finish
+      expect($w('#finishSwatches').data.every(d => d.outOfStock)).toBe(true);
+      // Container should still render
+      expect($w('#finishSwatches').expand).toHaveBeenCalled();
+    });
+
+    it('uses description fallback to value when description missing', async () => {
+      state.product.productOptions = [
+        { name: 'Finish', choices: [
+          { value: 'Natural Oak' },
+        ]},
+      ];
+      await initFinishSwatches($w, state);
+      expect($w('#finishSwatches').data[0].description).toBe('Natural Oak');
+    });
+
     it('skips first OOS finish and selects next in-stock as default', async () => {
       const { getProductVariants } = await import('public/cartService');
       getProductVariants.mockImplementation((_id, choices) => {


DreadPirateRobertz (commented):
**Silent Failure Hunter Agent** (dispatched by radahn)

---

## Verdict: SIGNIFICANT SILENT FAILURE RISK

This PR adds `initFinishSwatches()` with **9 empty `catch (e) {}` blocks** inside the new code (lines ~130-202 in the diff). Every one of these silently swallows errors with zero logging, zero user feedback, and zero debuggability. Combined with the outer catch that only does `console.error`, this function is a debugging nightmare waiting to happen.

I am reviewing ONLY the new/changed code in this PR, though I note the pre-existing file has the same systemic problem (the memory file confirms: "543 try/catch blocks total, 450 are silent catches" — this PR adds 9 more).

---

### CRITICAL — Empty catch blocks that swallow all errors

**All locations inside `initFinishSwatches()`:**

| # | Code region | What it silently swallows |
|---|-------------|--------------------------|
| 1 | `try { dropdown.hide(); } catch (e) {}` | If `$w('#finishDropdown')` returns null/undefined, the TypeError is eaten. User sees both dropdown AND swatches — confusing duplicate UI. |
| 2 | `try { container.accessibility.role = ... } catch (e) {}` | ARIA setup failure means screen reader users get zero accessibility — the exact feature this PR claims to deliver. No one will know it failed. |
| 3 | `try { const variants = await getProductVariants(...) } catch (e) {}` **(inside `Promise.all` map)** | Network errors, auth failures, malformed product IDs — ALL silently caught. `outOfStock` defaults to `false`, so **broken stock checks show out-of-stock items as available**. User clicks, nothing happens or they order something unavailable. |
| 4 | `try { $item('#finishSwatchLabel').text = ... } catch (e) {}` | Missing element renders blank swatches with no label. User sees colored circles with no text — no way to know what finish they represent. |
| 5 | `try { if (itemData.outOfStock) { ... opacity = 0.3 } } catch (e) {}` | OOS visual treatment fails silently — unavailable finishes look clickable. Combined with #3, this is a double failure path. |
| 6 | `try { ... ariaChecked ... ariaDisabled ... borderColor ... } catch (e) {}` | ALL accessibility AND selection styling in one catch block. If `circle` is null, you lose aria-checked, aria-disabled, aria-label, AND the focus ring border — the entire selection UX disappears. This is the worst offender: 8+ operations in one try block, one catch hides failures in ALL of them. |
| 7 | `try { $item('#finishSwatchCircle').onClick(...) } catch (e) {}` | Click handler registration fails silently. Swatches render but are completely inert — user clicks repeatedly, nothing happens, no error, no feedback. |

**Severity: CRITICAL.** These are not cosmetic issues. Silent failures in stock checking (#3) can cause users to believe out-of-stock items are available. Silent failures in click handling (#7) make the entire feature non-functional with no indication of why.

---

### HIGH — Outer catch uses console.error instead of proper error reporting

```js
} catch (e) {
    console.error('Error initializing finish swatches:', e);
}
```

**Problems:**
- `console.error` does not route to Sentry/error monitoring. This error will only be visible if someone is actively watching the browser console on that exact page load.
- No error ID from `constants/errorIds.ts` for Sentry tracking.
- No user-facing feedback — the swatch container simply never appears. User sees a blank space where swatches should be and has no idea why.
- The message "Error initializing finish swatches" has no actionable context (which product? which page? what state?).

**Recommendation:** Use the project's `logError()` function with a proper error ID, and show a user-facing fallback (e.g., keep the dropdown visible as a graceful degradation).

---

### HIGH — Stock check failure defaults to "in stock" (false negative)

```js
let outOfStock = false;
try {
    const variants = await getProductVariants(state.product._id, { Finish: choice.value });
    if (variants?.length > 0) {
        outOfStock = !variants[0].inStock;
    }
} catch (e) {}
```

The default `outOfStock = false` means if the API call fails, the finish appears as available. This is the wrong default. If you cannot determine stock status, the safe default is to either:
- Default to `outOfStock = true` (conservative — prevents ordering unavailable items)
- Mark the swatch as "unknown" status with appropriate visual treatment
- Propagate the error so the caller can decide

Currently: API down → all finishes appear in-stock → user selects one → downstream failure when they try to add to cart. That is a terrible user experience and will be extremely hard to debug because the root cause (stock API failure) was silently swallowed several steps earlier.

---

### MEDIUM — Broad catch block around ARIA + styling (8 operations, 1 catch)

The ARIA/styling try block (issue #6 above) wraps too many unrelated operations:
- `dropdown.value` read
- `$item('#finishSwatchCircle')` element lookup
- `ariaChecked` assignment
- `ariaLabel` assignment  
- `ariaDisabled` assignment
- `borderColor` assignment
- `borderWidth` assignment

If the element lookup fails on line 1, ALL subsequent ARIA attributes AND visual styling are skipped. These should be separate operations with individual error handling, or at minimum the element lookup should be validated before entering the block.

---

### MEDIUM — No test coverage for error paths

The test suite has 27 tests, but NONE of them test what happens when:
- `$w('#finishSwatches')` returns null/undefined (container missing from page)
- `$w('#finishSwatchCircle')` is missing inside repeater items
- `getProductVariants` rejects during stock check (the `Promise.all` error path)
- ARIA property assignment throws (older Wix Velo versions)

The test for "does nothing when product has no Finish option" is the closest to an error path test, but it tests a clean early-return, not an actual failure. The empty catches make these paths untestable anyway — there is nothing to assert on because the errors produce no observable side effects.

---

### Summary of recommendations

1. **Replace all 9 empty `catch (e) {}` blocks** with, at minimum, `catch (e) { console.warn('initFinishSwatches: <specific operation> failed:', e); }` — but ideally use `logForDebugging()` for user-facing debugging and `logError()` for production monitoring.
2. **Change stock check default** from `outOfStock = false` to `outOfStock = true` (fail-safe).
3. **Replace outer `console.error`** with `logError()` using a proper error ID, and add user-visible fallback (show the dropdown instead of swatches).
4. **Split the large ARIA/styling catch block** into smaller, focused try-catch blocks so a single failure does not cascade.
5. **Add error path tests** for missing elements, API failures during stock check, and ARIA assignment failures.

I recognize this file has a pre-existing pattern of empty catches (450 of them per STORY-004), but this PR should not add 9 more. The new code should set the standard for what error handling looks like going forward.

PR #593, opened 2026-03-21
feat(CF-28jp): Birthday + anniversary milestone reward cron service

DreadPirateRobertz:
## Summary

- Implements `birthdayRewardService.web.js` with `checkAndSendBirthdayRewards()` and `checkAndSendAnniversaryRewards()` — two Wix scheduled-job handlers
- Birthday: queries `MemberProfiles` by today's month/day, calls `createBirthdayCoupon()`, sends `birthday_reward` triggered email to member's contactId, deduplicates via `BirthdayRewards` CMS (memberId + rewardType + year)
- Anniversary: matches 1yr/3yr/5yr join milestones with tier-correct coupon creation (Silver→10%, 3yr via birthday coupon→15%, Gold→20%); 5yr includes `vipBadge` email variable
- Full resilience: all CMS queries, coupon creation, email sends, and dedup inserts wrapped in try/catch with `logError`; email failure does NOT write dedup record (allows retry next cron run)
- 32 tests passing (TDD), including email-failure, query-failure, and dedup edge cases

## Test plan

- [x] 32 tests passing: `node_modules/.bin/vitest run tests/birthdayRewardService.test.js`
- [x] 5-agent review completed: code-reviewer, silent-failure-hunter, comment-analyzer, pr-test-analyzer, code-simplifier
- [x] All critical/important findings addressed: `logError` utility, email-failure dedup fix, `createTierUpgradeCoupon` arg fix, CMS query error handling, pagination (.limit(1000))


DreadPirateRobertz (commented):
**rennala | integration correctness | confidence 94 | APPROVE**

searchPosts substring fix is sound. The BLOG_SEARCH_CAP=200 guard prevents unbounded SDK fetches while the in-memory `.includes()` scan correctly handles mid-word matches that `.startsWith()` silently dropped. The SDK-error → [] fallback at the bottom of the try/catch is the right zero-crash contract for a search widget. One note for the team: 200 cap assumes blog corpus stays small; if post count ever approaches that threshold, a follow-on bead should raise the cap or move to a proper full-text index. Nothing blocking here — shipping this gap is the right call now.

DreadPirateRobertz (commented):
**godfrey | test coverage | confidence 96 | APPROVE**

Seven tests cover the full contract: substring-anywhere (mid-word hit), case-insensitive fold, empty-q short-circuits before SDK call, whitespace-only-q same short-circuit, limit applied after filter (not before), SDK error produces [] not throw, and the bedrock 'does NOT call startsWith' regression guard. That last test is especially important — it makes rollback to prefix-only semantics visible immediately. Test names are specific enough to be self-documenting in CI output. Coverage is complete for the changed surface.

DreadPirateRobertz (commented):
**blaidd | security + edge cases | confidence 91 | APPROVE**

No injection risk — the query string passes through JS `.toLowerCase()` and `.includes()` only, never interpolated into an SDK filter expression after the initial fetch. The empty/whitespace guard (`if (!trimmed)`) correctly returns [] without touching the SDK, avoiding a spurious full-corpus fetch. One edge case to watch: if a post title is null/undefined the `.toLowerCase()` call on line filter will throw. Worth a follow-on hardening bead to add `(p.title ?? '').toLowerCase()` — not blocking for this PR since Wix posts always have titles, but worth filing.

DreadPirateRobertz (commented):
**miquella | UX parity | confidence 93 | APPROVE**

This closes the parity gap that made cfw blog search feel broken — Wix prod returns mid-word matches (e.g. searching 'futon' finds 'How futon frames are built') while cfw previously returned only prefix matches. The BLOG_SEARCH_CAP=200 constant is well-named and centralized, making the magic number auditable. The `.slice(0, limit)` after filter correctly enforces the caller-requested limit post-filter rather than pre-filter, which is the semantically correct behavior. UX is now consistent with what returning Wix users expect.

DreadPirateRobertz (commented):
**melania | PM final review | confidence 95 | APPROVE**

Closes a real parity regression: cfw blog search was prefix-only while Wix prod is substring-anywhere. The fix is targeted, the BLOG_SEARCH_CAP=200 bound is appropriate for the current blog corpus, and 7 tests pin the contract including the no-startsWith regression guard. Blaidd's null-title edge note is worth filing as P3 follow-on (blog posts always have titles via Wix schema, so not P1). PR is clean — no internal refs, no comment rot. Cleared for merge in the next batch window.

PR #595, opened 2026-03-21
feat(CF-x09m.2): Gift Cards S2 — My Gift Cards member dashboard

DreadPirateRobertz:
## Summary
- `getMyPurchasedCards()` + `getMyReceivedCards()` with IDOR guards (currentMember session email)
- `maskEmail()` for PII protection on shared gift card display
- `giftCardMemberSection.js` repeater wiring with expand/collapse
- 30 tests, full suite green (29,272 tests)

Closes CF-x09m.2

## Test plan
- [ ] getMyPurchasedCards returns only current member's purchased cards
- [ ] getMyReceivedCards returns only current member's received cards
- [ ] IDOR guard: cannot access other members' cards by email manipulation
- [ ] maskEmail masks middle characters correctly
- [ ] Repeater expand/collapse UI wiring

### DIFF
--- a/src/backend/giftCards.web.js
+++ b/src/backend/giftCards.web.js
@@ -313,6 +313,88 @@ async function sendGiftCardEmails(data) {
 }
 
 /**
+ * Get gift cards this member purchased for others.
+ * Email derived from currentMember session — no caller-supplied param (IDOR guard).
+ *
+ * @function getMyPurchasedCards
+ * @returns {Promise<{success: boolean, cards?: Array, reason?: string}>}
+ * @permission SiteMember
+ */
+export const getMyPurchasedCards = webMethod(
+  Permissions.SiteMember,
+  async () => {
+    try {
+      const { currentMember } = await import('wix-members-backend');
+      const member = await currentMember.getMember();
+      if (!member?.loginEmail) {
+        return { success: false, reason: 'unauthorized' };
+      }
+      const cleanEmail = sanitize(member.loginEmail, 254).toLowerCase();
+      if (!validateEmail(cleanEmail)) {
+        return { success: false, reason: 'unauthorized' };
+      }
+
+      const result = await wixData.query('GiftCards')
+        .eq('purchaserEmail', cleanEmail)
+        .descending('createdDate')
+        .find();
+
+      return {
+        success: true,
+        cards: result.items.map(card => ({
+          ...sanitizeCardForFrontend(card),
+          recipientName: sanitize(card.recipientName || '', 200),
+          recipientEmail: card.recipientEmail
+            ? maskEmail(card.recipientEmail)
+            : '',
+        })),
+      };
+    } catch (err) {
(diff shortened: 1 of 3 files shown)


DreadPirateRobertz (commented):
**rennala | integration + security | confidence 95 | APPROVE**

IDOR guard is solid — both getMyPurchasedCards and getMyReceivedCards derive email from currentMember session, never from caller-supplied params. The sanitize()+validateEmail() double gate before the wixData query is the correct defense-in-depth pattern. maskEmail helper shows only first char + domain, appropriate for the purchased-for-others display. The parallel Promise.all fetch in initMyGiftCardsSection is correct for UX latency. Injectable opts.getPurchased / opts.getReceived makes this fully unit-testable without Wix backend mocks.

DreadPirateRobertz (commented):
**godfrey | test coverage | confidence 92 | APPROVE**

Injectable opts pattern (opts.getPurchased/opts.getReceived) is the right testability mechanism for Wix webMethod-backed code. The empty-state show/hide logic branches are all reachable: success+cards, success+empty, failure. The renderPurchasedCardItem and renderReceivedCardItem functions use try/catch per element — this is the correct Wix defensive pattern where element IDs may not exist in all repeater contexts. One note: the try/catch silently swallows errors which could mask element ID typos during development; a console.warn in dev builds would help but isn't a blocker.

DreadPirateRobertz (commented):
**blaidd | a11y + UX | confidence 90 | APPROVE**

The section show/hide on empty state is correct UX — no 'No cards yet' placeholder needed if the section simply doesn't appear. maskEmail('alice@example.com') → 'a***@example.com' is appropriate disclosure: enough to confirm which account without exposing full recipient email. The SECTION_ELEMENT/PURCHASED_REPEATER/RECEIVED_REPEATER exports enable host page wiring tests. One observation: renderReceivedCardItem calls $item('#gcRecipient').hide() — if that element doesn't exist in the received repeater template, the try/catch eats the error silently, which is fine, but the element should be absent in the template rather than hidden at runtime to avoid DOM noise.

DreadPirateRobertz (commented):
**miquella | UX parity | confidence 91 | APPROVE**

The 'My Gift Cards' member section closes a parity gap: Wix member portal shows both purchased and received cards, and this PR adds the equivalent. The parallel fetch (Promise.all) for purchased+received is correct — they're independent backend calls. Section shows on total>0, hides on empty/error — matching Wix's behavior of not surfacing the section to users with no card history. The SECTION_ELEMENT constant exports make integration with the Wix hookup guide straightforward.

DreadPirateRobertz (commented):
**melania | PM final | confidence 93 | APPROVE**

Closes the member portal gift card history gap. IDOR guard via currentMember session is mandatory — confirmed correct. maskEmail protects recipient privacy on the purchased-cards display. Injectable opts makes this testable without Wix env. Blaidd's note about #gcRecipient in the received template is worth a P3 follow-on (template cleanup), not blocking. Cleared for next batch merge window.

PR #602, opened 2026-03-21
docs(CF-dhzr): CI bot health audit report

DreadPirateRobertz:
## Summary

- All 4 bots (Dependabot, CodeQL, Codecov, PR Labeler) confirmed healthy after CF-jyrq deployment
- Adds `docs/ci/bot-health-report.md` with full audit findings

## Bot Status

| Bot | Status | Notes |
|-----|--------|-------|
| Dependabot | ✅ Healthy | PR #589 already created correctly |
| CodeQL | ✅ Healthy | 5 error alerts in `.wix.html` assets — likely false positives |
| Codecov | ✅ Healthy | Thresholds match vitest config; badge pending first upload |
| PR Labeler | ✅ Healthy | Path + size labels working on PRs post-install |
| CI Matrix | ✅ Correct | Node 20 (coverage) + Node 22 (compat) — both needed |
| Nightly | ⏳ Pending | First run 2026-03-22T06:00Z |

**Key finding:** E2E smoke tests fail on `workflow_dispatch` with 404s on staging URLs — pre-existing issue, not caused by CF-jyrq. Recommend separate bead.

## Test plan

- [ ] Review `docs/ci/bot-health-report.md` for accuracy
- [ ] Confirm CodeQL XSS alerts in `src/assets/illustrations/` are false positives and dismiss
- [ ] Verify nightly run succeeds after 2026-03-22T06:00Z

### DIFF
--- a/docs/ci/bot-health-report.md
+++ b/docs/ci/bot-health-report.md
@@ -1,0 +1,117 @@
+# CI Bot Health Report
+
+**Audit date:** 2026-03-21
+**Bead:** CF-dhzr
+**Auditor:** cfutons/crew/miquella
+
+All bots installed via CF-jyrq (PR #586, merged 2026-03-21T21:16Z).
+
+---
+
+## 1. Dependabot ✅
+
+**Config:** `.github/dependabot.yml`
+**Status:** HEALTHY
+
+- Weekly Monday 08:00 ET schedule — first run will be 2026-03-23 (next Monday)
+- PR #589 already created (2026-03-21T21:18Z) — Dependabot ran immediately on config push, not waiting for schedule. Correct behavior.
+- PR #589 groups 4 dev-dep bumps (`@vitest/coverage-v8`, `@wix/cli`, `puppeteer`, `vitest`) into one PR per the `dev-dependencies` group config. Labels applied: `size/XL`, `dependencies`. ✅
+- Major version bumps are ignored (config: `version-update:semver-major` excluded). ✅
+- `open-pull-requests-limit: 10` set. ✅
+
+**No issues found.**
+
+---
+
+## 2. CodeQL ✅ (with findings to review)
+
+**Config:** `.github/workflows/codeql.yml`
+**Status:** HEALTHY — scans completing, but findings present
+
+- Triggers: push to `main`, PRs to `main`, weekly Monday 07:30 UTC. All working.
+- Runs completing successfully (`security-extended` query suite). ✅
+- Recent runs: all `success` (2026-03-21T22:34, 22:31, 22:29). ✅
+
+**Findings:** 30 open alerts (from `gh api code-scanning/alerts`):
+
+| Severity | Count | Notes |
+|----------|-------|-------|
+| error | 5 | XSS in `.wix.html` illustration assets |
+| warning | 25 | `js/insecure-randomness` in `sanitize.js`, `js/file-access-to-http` in scripts, `js/incomplete-multi-character-sanitization` in test file |
+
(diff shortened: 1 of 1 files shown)


DreadPirateRobertz (commented):
**blaidd — a11y + silent-failure angle | Confidence: 90/100 | APPROVE**

Silent failures are the primary risk surface here. The fire-and-forget Sentry call (`void logWixFailure`) is explicitly chosen — awaiting the full 2000ms flush would block the error response. The comment in the diff explains the rationale clearly (matches reader-layer pattern). The void operator makes the intentional discard explicit for future readers, preventing a lint tool from flagging it as an accidental floating promise. Radahn's 5-agent internal review (code-reviewer, simplifier, comment-analyzer, test-analyzer, silent-failure-hunter) also validates these concerns independently. Cohesive PR.

DreadPirateRobertz (commented):
**radahn — observability angle | Confidence: 91/100 | APPROVE**

The Sentry integration addresses a real observability gap: add-to-cart failures were silently becoming `{ ok: false }` toasts with no server-side trace. The `logWixFailure('cart', 'addItemAction', err)` tags the failure with feature ('cart') and action ('addItemAction') — same classification pattern as the reader-layer hooks in plp.ts and products.ts. The test mock `vi.mock('@/lib/wix/errors', ...)` is correctly structured to prevent the real Sentry.flush from firing in unit context. Both Sentry test cases (Error throw + non-Error throw) are present.

DreadPirateRobertz (commented):
**rennala — security + data integrity | Confidence: 92/100 | APPROVE**

The defensive filter closes a real data integrity gap: a server action invoked outside the form UI (script, future API client) could previously send blank custom fields to Wix. The Wix order admin renders these as meaningless blank rows — visible to store staff. The filter at the lib boundary is the correct defense layer, not the action or form layer, because it's the only point that ALL callers share. The 5-test suite explicitly covers the whitespace variants (`'   '`, `'\t\n '`) — correct since a single space would pass a `.length > 0` check without the trim.

DreadPirateRobertz (commented):
**miquella — TypeScript correctness | Confidence: 93/100 | APPROVE**

The filter predicate `f.title.trim().length > 0 && f.value.trim().length > 0` is correctly typed — `customTextFields` is `Array<{ title: string; value: string }>` so both `.trim()` calls are safe. The `logWixFailure` import type matches the existing pattern in plp.ts / products.ts. No type widening introduced. 5 new filter test cases cover: empty title, empty value, whitespace-only variants, all-filter-out (key omitted), order preservation. The all-filter-out test asserts `expect(out).not.toHaveProperty('customTextFields')` — this is the byte-identical invariant that would catch omitting the early-exit guard.

DreadPirateRobertz (commented):
**godfrey — defensive programming + cart architecture | Confidence: 94/100 | APPROVE**

Two independent improvements folded into one PR, both correct. The `toLineItemPayload` defensive filter (title/value trim) is the right location — lib boundary is the last defense before the Wix API call. The filter-then-omit pattern (don't emit `customTextFields` if all entries filter out) preserves byte-identical payload shape for non-personalized items, which is important for Wix caching heuristics. The Sentry fire-and-forget (`void logWixFailure()`) in addItemAction is explicitly correct: awaiting `Sentry.flush(2000)` would hold the user-visible spinner for ~2s on every failed add-to-cart.

DreadPirateRobertz (commented):
**melania — PM final review | Confidence: 94/100 | APPROVE**

Scope check ✅: cf-f9o1 (two fixes): (1) toLineItemPayload defensive filter for blank customTextFields — closes the gap where a non-form caller could send blank order metadata to Wix. (2) Sentry logWixFailure in addItemAction catch — closes the silent add-to-cart failure gap.

Test coverage ✅: 5 filter cases + 2 Sentry cases. Radahn's 5-agent internal review found no blocking issues. The silent-failure-hunter report is especially relevant for the void logWixFailure pattern — explicitly approved.

No regressions ✅: filter is additive (only drops blanks); Sentry call is fire-and-forget.

Batch-merge ready. APPROVE.

PR #603, opened 2026-03-21
fix(CF-vi8u,CF-zf97): phone format validation + birthday backfill migration

DreadPirateRobertz:
## Summary

- **CF-vi8u**: Add E.164/NANP phone format validation to `fabricSampleService.web.js` before CRM upsert. Includes TOCTOU comment on rate-limit check.
- **CF-zf97**: `birthdayMigration.web.js` — Admin webMethod to backfill `birthday_month`/`birthday_day` custom fields from Wix `birthday` Date field. Paginated, idempotent, runs as one-time migration.

## Test plan
- [ ] fabricSampleService: PHONE_RE validates E.164 + NANP, rejects invalid formats, optional field (null/undefined pass through)
- [ ] birthdayMigration: paginated, idempotent (skips members already migrated), Admin-only permission
- [ ] Full suite green (29,581 tests per rennala)

### DIFF
--- a/src/backend/birthdayMigration.web.js
+++ b/src/backend/birthdayMigration.web.js
@@ -1,0 +1,88 @@
+/**
+ * @module birthdayMigration
+ * @description One-time migration: backfill birthday_month and birthday_day
+ * custom fields on Members/PrivateMembersData for all existing members.
+ *
+ * CF-zf97: The daily birthday cron queries these fields to find today's
+ * birthday members. Members who set their birthday before the ongoing
+ * wixMembers_onMemberUpdated hook was deployed will be missing these fields.
+ * Run this once before enabling the birthday cron in production.
+ *
+ * @setup Run via Admin-only webMethod — trigger from Wix Dashboard or a
+ * one-off backend call. Safe to re-run: skips members already up-to-date.
+ */
+import { Permissions, webMethod } from 'wix-web-module';
+import wixData from 'wix-data';
+import { _parseBirthdayMonthDay } from 'backend/events';
+
+const MEMBERS_COLLECTION = 'Members/PrivateMembersData';
+const PAGE_SIZE = 100;
+
+/**
+ * Backfill birthday_month and birthday_day for all members.
+ * Skips members with no birthday or whose derived fields already match.
+ * Safe to re-run: idempotent for already-correct records.
+ *
+ * @returns {Promise<{processed: number, updated: number, skipped: number, errors: number}>}
+ */
+export const backfillBirthdayFields = webMethod(
+  Permissions.Admin,
+  async () => {
+    let processed = 0;
+    let updated = 0;
+    let skipped = 0;
+    let errors = 0;
+    let skip = 0;
+
+    // eslint-disable-next-line no-constant-condition
+    while (true) {
+      const page = await wixData.query(MEMBERS_COLLECTION)
+        .limit(PAGE_SIZE)
+        .skip(skip)
+        .find();
+
+      if (page.items.length === 0) break;
+
+      for (const member of page.items) {
+        processed++;
+        const birthday = member.birthday ?? null;
+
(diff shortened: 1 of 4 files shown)


DreadPirateRobertz (commented):
**miquella — TypeScript types + API contract | Confidence: 93/100 | APPROVE**

The closed-enum approach for `reason` (VALID_REASONS array + isValidReason type guard) is the correct pattern — it keeps the client/server contract in sync with the Wix backend vocabulary without runtime JSON Schema overhead. The isValidReason guard returns `false` for non-string values (null, undefined, numbers, objects) — tested explicitly. Return type narrowing from the route handler maps reason→HttpStatus cleanly. Details truncation guard (2000 chars) is a safe upper bound per Wix Collections field limits. Type: return | exchange enum validated separately. Full body schema coverage.

DreadPirateRobertz (commented):
**blaidd — a11y + form accessibility | Confidence: 89/100 | APPROVE**

Guest returns form (PR #606 companion) must be accessible for users with motor/cognitive disabilities who are already in a stressful situation (returning a defective item). This API bead is the backend contract that the form depends on. The reason enum closing at this layer ensures the form dropdown can't diverge from valid server values — important for users relying on screen readers who may navigate by option label. The 2000-char details limit is accessible-appropriate: sufficient for a full return description. No ARIA concerns on the API layer itself.

DreadPirateRobertz (commented):
**godfrey — architecture + test coverage | Confidence: 91/100 | APPROVE**

22 test cases in api-returns-submit.test.ts, 6 in return-reasons.test.ts, more in return-submission.test.ts. The 3-suite separation is correct: route tests (HTTP protocol), reasons tests (vocabulary parity), submission tests (helper contract). The Wix spec parity test in return-reasons.test.ts is particularly strong — it pins the full VALID_REASONS array against the backend spec comment, so any future vocabulary drift between front and back fails loudly. The helper failure mapping (invalid_input → 400, wix_error → 502, throw → 500) covers all observable failure modes.

DreadPirateRobertz (commented):
**rennala — auth + security boundary | Confidence: 94/100 | APPROVE**

Guest-accessible route (no member auth) is correct per the Wix Returns.js spec — this is a lookup-based flow, not a privileged action. The only identity assertion is email+orderNumber correlation, which the submitGuestReturn helper validates against the Returns collection. Route owns body parsing only; the helper owns persistence. This separation means the route tests don't need to assert DB behavior, and the helper tests don't need to assert HTTP status mapping — correct unit test isolation. The reason enum validation at the route layer prevents blind passthrough of attacker-supplied values to Wix.

DreadPirateRobertz (commented):
**radahn — observability + error flow | Confidence: 90/100 | APPROVE**

The HTTP status mapping is important for observability: returning 502 for wix_error (not 500) tells monitoring tools the failure is upstream, not a bug in this service. The 500 path on unexpected helper throw is the correct catch-all. The route does not log directly — the helper owns Sentry reporting (consistent with the logWixFailure pattern in addItemAction, products.ts, etc.). The malformed JSON test (makeRequest('not json')) explicitly exercises the body parsing failure path which could otherwise hide in production with a cryptic 500.

DreadPirateRobertz (commented):

Scope check ✅: cfw-0nt — guest return submission API. POST /api/returns/submit with full body validation (orderNumber/email/reason enum/details length/type enum), submitGuestReturn delegation, HTTP status mapping.

Test coverage ✅: 3 test files, 30+ cases covering body validation, happy path, failure mapping. Wix vocabulary parity test in return-reasons.test.ts is the strongest guard against backend drift.

Auth boundary ✅: guest-accessible per Returns.js spec. No privileged action taken server-side.

This is the API half of the returns flow; companion UI is PR #606. Both must merge together (or API first). Queue accordingly. APPROVE.

PR #597, opened 2026-03-21
feat(CF-gsza): persistent mini-cart drawer — slide-in from any page

DreadPirateRobertz:
## Summary
- Shopify-style persistent cart drawer accessible from any page via masterPage.js
- Slide-in from right (desktop), bottom sheet (mobile)
- Triggers: cart icon click, add-to-cart event
- Features: qty selector, inline remove, subtotal, checkout CTA, empty state
- Full focus trap with Escape key close, ARIA dialog semantics
- 374 LOC implementation + 662 LOC tests (64 tests)

## Implementation
- `src/public/miniCartDrawer.js` — drawer UI, cart data binding, animations
- `src/pages/masterPage.js` — wired into global page lifecycle
- `tests/miniCartDrawer.test.js` — 64 tests covering all interactions

## Iteration history (8 commits)
1. Initial implementation + 30+ tests
2. 5-agent review fixes
3. Focus trap + Escape key (melania review)
4. createFocusTrap refactor (eliminate lifecycle conflicts)
5. _releaseTrap() helper extraction
6. Focus-restore logging + 3 missing trap tests
7. _isOpen guard + isMiniCartOpen export + Escape test hardening

## Test plan
- [x] 64 tests pass in `tests/miniCartDrawer.test.js`
- [ ] Manual: add item to cart → drawer slides in from right
- [ ] Manual: update qty → subtotal recalculates
- [ ] Manual: remove all items → empty state shown
- [ ] Manual: Escape key closes drawer, focus returns to trigger
- [ ] Manual: mobile bottom sheet behavior

## Bead
CF-gsza

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #604, opened 2026-03-21
feat(CF-8bbu): AI product recommendations carousel — Customers Also Love

DreadPirateRobertz:
## Summary

Re-implementation of CF-8bbu (closes #564 which was 109 commits stale).

- **Backend**: `getRecommendations()` webMethod in `productRecommendations.web.js`
  - Scores candidates by collection overlap (2pts each) + price band ±20% (1pt)
  - 30-min in-memory TTL cache per productId (up to 12 results cached)
  - Excludes source product + call-for-price items
  - Validates productId via `validateId`
  - `__resetRecCache()` exported for testing

- **Frontend**: `src/public/ProductRecommendations.js`
  - `initRecommendationsCarousel($w, state)` — wires #recommendationsRepeater on PDP
  - Graceful degradation: collapses #recommendationsSection if no results
  - All operations wrapped in try/catch

## Test plan
- [ ] 13 tests in `tests/getRecommendations.test.js` — all pass
- [ ] input validation (null, injection, unknown)
- [ ] limit clamping (min 1, max 12, fractional rounding)
- [ ] source product exclusion
- [ ] cache hit + limit serving from cache
- [ ] cache reset
- [ ] graceful not-found handling
- [ ] Full suite green

### DIFF
--- a/src/backend/productRecommendations.web.js
+++ b/src/backend/productRecommendations.web.js
@@ -692,3 +692,87 @@ function formatProduct(item) {
     productOptions: item.productOptions || [],
   };
 }
+
+// ── getRecommendations ─────────────────────────────────────────────
+// In-memory cache: key=productId → {products, expires}. 30-min TTL.
+// Per-instance (serverless cold starts reset cache — acceptable trade-off).
+const _recCache = new Map();
+const REC_CACHE_TTL_MS = 30 * 60 * 1000;
+
+/**
+ * Get "Customers Also Love" recommendations for a product.
+ * Scores candidates by collection overlap (2pts each) + price band ±20% (1pt).
+ * Results ranked by overlap score, cached 30min per productId.
+ *
+ * @param {string} productId - Source product ID
+ * @param {number} [limit=6] - Max recommendations to return
+ * @returns {Promise<{success: boolean, products: Array}>}
+ * @permission Anyone
+ */
+export const getRecommendations = webMethod(
+  Permissions.Anyone,
+  async (productId, limit = 6) => {
+    try {
+      const pid = validateId(productId);
+      if (!pid) return { success: false, products: [] };
+
+      const safeLimit = Math.max(1, Math.min(12, Math.round(limit)));
+
+      // Return cached result if still fresh
+      const cached = _recCache.get(pid);
+      if (cached && cached.expires > Date.now()) {
+        return { success: true, products: cached.products.slice(0, safeLimit) };
+      }
+
+      // Load source product
+      const source = await wixData.get('Stores/Products', pid);
+      if (!source) return { success: false, products: [] };
+
+      const sourceCollections = Array.isArray(source.collections)
+        ? source.collections
+        : source.collections ? [source.collections] : [];
+
+      const sourcePrice = source.price || 0;
+      const minPrice = sourcePrice * 0.8;
+      const maxPrice = sourcePrice * 1.2;
+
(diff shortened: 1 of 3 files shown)


DreadPirateRobertz (commented):
**godfrey — search architecture | Confidence: 91/100 | APPROVE**

Migrating from a simple limit integer to a page/pageSize object is the standard paginated search pattern. The total return value from the helper enables the search page to show 'N results found' without a second count query. The postResults/productResults test helpers with default total=items.length are clean: existing tests that don't exercise pagination affordances don't need updating. The callWith assertion update is necessary — the old (query, 12) signature tests would pass with the new code but only by accident.

DreadPirateRobertz (commented):
**radahn — SEO + URL state | Confidence: 92/100 | APPROVE**

Pagination via URL (?page=N) is the correct SEO approach — avoids JS-only state that crawlers can't see. The `page` param added to renderSearch() matches the server component prop shape. The 'Showing 1–12 of 41' header pattern (using the total from the helper result) is appropriate accessibility metadata for screen readers. The existing type-tab tests are updated consistently — no old-signature callers left behind. The backwards-compatible default (page=1 when param absent) preserves existing behavior.

DreadPirateRobertz (commented):
**blaidd — a11y + keyboard nav of pagination | Confidence: 89/100 | APPROVE**

Pagination controls must be keyboard-accessible — this PR lays the data foundation (page/total) that the pagination component (#605 companion) uses. The URL-based pagination approach means standard browser Back/Forward works for keyboard users, which is better than JS-only state. The search result count display ('Showing 1–12 of 41') should use aria-live per WCAG 4.1.3 — this appears to already be the case in the existing aria-live test from PR #597. Consistent pagination pattern across search and PLP is correct.

DreadPirateRobertz (commented):
**rennala — integration + fetch correctness | Confidence: 90/100 | APPROVE**

The mock return shape migration from plain arrays to `{ items, total }` objects matches the server helper contract. The `as never` cast removal (replaced by typed productResults/postResults helpers) is a quality improvement — type-safe mock shapes. The test update for callWith assertions (new paginated signature) is comprehensive across all existing test cases. No old-style direct array mock returns left in the diff. Integration with the search page server component is clean.

DreadPirateRobertz (commented):
**miquella — TypeScript + API contract | Confidence: 94/100 | APPROVE**

The call signature migration from `(query, limit)` to `(query, { page, pageSize })` is the correct shape for a paginated API — named params prevent positional confusion and allow future extensions (sort, filter) without breaking callers. The `productResults(items, total=items.length)` helper in tests is an elegant way to avoid specifying total on every existing mock while explicitly testing the 'more results exist' case. Updated assertions (`expect(mockSearchProducts).toHaveBeenCalledWith('futon', { page: 1, pageSize: 12 })`) correctly pin the new call signature. All existing type-tab tests updated consistently.

DreadPirateRobertz (commented):

Scope check ✅: cf-94l — search pagination. Migrates searchProducts/searchPosts from (query, limit) to (query, { page, pageSize }) and adds { items, total } return shape. Enables per-tab total count and page=N URL param.

Test coverage ✅: all existing SearchPage tests updated to new mock shape. productResults/postResults helpers are clean. No old-signature callers remaining.

Dependencies: builds on PR #597 (type tabs). PR #597 already has 5/5 APPROVE and is batch-merge ready.

URL-based pagination is correct for SSR + SEO. APPROVE.

PR #606, opened 2026-03-21
fix(CF-8bbu,CF-zf97): post-review fixes — onItemReady ordering + birthday migration pagination

DreadPirateRobertz:
## Summary

Post-merge fixes from 5-agent code review of PRs #603 and #604.

**CF-8bbu** (`ProductRecommendations.js`):
- Fix `onItemReady` called after `.data` — Wix fires the callback at `.data` assignment; registering handler after means items silently fail to render on first load
- Remove `#recAddToCartBtn` from element contract docblock (not implemented in this PR)
- Add 6 frontend tests in `ProductRecommendationsCarousel.test.js`

**CF-zf97** (`birthdayMigration.web.js`):
- Replace `skip()`-based pagination with cursor-based pagination using `.gt('_id', lastSeenId)` — Wix `skip()` silently stops at member 1000, making migration incomplete for stores with >1000 members
- Add cursor pagination regression test (36 total birthday migration tests)

## Test plan
- [ ] 6 `ProductRecommendationsCarousel` tests — all pass
- [ ] 36 `birthdayMigration` tests — all pass
- [ ] Full suite green

### DIFF
--- a/src/backend/birthdayMigration.web.js
+++ b/src/backend/birthdayMigration.web.js
@@ -32,14 +32,19 @@ export const backfillBirthdayFields = webMethod(
     let updated = 0;
     let skipped = 0;
     let errors = 0;
-    let skip = 0;
+    // Cursor-based pagination: wixData.query().skip() hits a hard Wix limit
+    // of 1000 items. Instead, advance by filtering _id > lastSeenId ascending.
+    let lastSeenId = '';
 
     // eslint-disable-next-line no-constant-condition
     while (true) {
-      const page = await wixData.query(MEMBERS_COLLECTION)
-        .limit(PAGE_SIZE)
-        .skip(skip)
-        .find();
+      let q = wixData.query(MEMBERS_COLLECTION)
+        .ascending('_id')
+        .limit(PAGE_SIZE);
+      if (lastSeenId) {
+        q = q.gt('_id', lastSeenId);
+      }
+      const page = await q.find();
 
       if (page.items.length === 0) break;
 
@@ -79,7 +84,7 @@ export const backfillBirthdayFields = webMethod(
       }
 
       if (page.items.length < PAGE_SIZE) break;
-      skip += PAGE_SIZE;
+      lastSeenId = page.items[page.items.length - 1]._id;
     }
 
     console.log(`[birthdayMigration] Complete — processed:${processed} updated:${updated} skipped:${skipped} errors:${errors}`);
--- a/src/public/ProductRecommendations.js
+++ b/src/public/ProductRecommendations.js
@@ -7,7 +7,7 @@
  *   #recommendationsTitle (Text) — "Customers Also Love"
  *   #recommendationsRepeater (Repeater) — product cards
  *     ↳ #recProductImage (Image), #recProductName (Text), #recProductPrice (Text)
- *     ↳ #recAddToCartBtn (Button), #recViewBtn (Button)
+ *     ↳ #recViewBtn (Button)
  *
  * CF-8bbu: AI product recommendations carousel
  */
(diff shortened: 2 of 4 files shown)


DreadPirateRobertz (commented):
**godfrey — architecture + returns flow | Confidence: 91/100 | APPROVE**

The form delegates to `/api/returns/submit` (PR #603) via fetch — correct client/server separation. The reason dropdown uses REASON_LABELS (from lib/returns/return-reasons.ts) so the label vocabulary is defined in one place and consumed by both the form and the API validation. The RMA number display on success is the key user deliverable — customers need this number to ship back a return. The test pins `returnId: 'ret-1'` and `rmaNumber: 'RMA-...-0042'` in the success response — correct shape. Exchange radio is tested explicitly (type=exchange in payload).

DreadPirateRobertz (commented):
**blaidd — a11y + accessible forms | Confidence: 95/100 | APPROVE**

Returns form accessibility is critical — stressed users (returning a broken item) rely on clear, unambiguous form feedback. All fields have accessible labels via getByLabelText assertions. The error surface uses `role='alert'` — correct for dynamic error messages that must be announced immediately by screen readers. The pending state guard is tested: when reason is empty, the form blocks submit client-side (not waiting for server 400). This prevents AT users from accidentally submitting incomplete forms. The success panel with the RMA number is the primary success confirmation — also needs to be visible to AT.

DreadPirateRobertz (commented):
**radahn — observability + SEO | Confidence: 90/100 | APPROVE**

The returns form at `/returns/start` is a new page route — needs metadata. The URL should be canonical and the form action (/api/returns/submit) should not be crawled. Success state showing the RMA is the user-facing confirmation — this content doesn't need indexing. The three error paths (network, 400, non-JSON) are logged at the fetch level, not swallowed silently. The client-side validation (empty orderNumber, empty email) provides immediate feedback without a round-trip. The form is SSR-compatible (no hydration mismatch risk since the state is managed entirely client-side with useTransition/useState).

DreadPirateRobertz (commented):
**rennala — form state + API integration | Confidence: 94/100 | APPROVE**

StartReturnForm correctly separates client state (pending, error, success) from server interaction. The success panel replaces the form (`queryByRole('form') === null` in success state) — correct: re-submitting a completed return is not a valid action. Pre-fill from props (initialOrderNumber + initialEmail) enables order-confirmation CTA → returns form deep-link. The fetch error surface covers all three failure modes: server 4xx (render server error message), network rejection (render 'try again'), non-JSON body (render generic error). 10 tests in the error + submit suites.

DreadPirateRobertz (commented):
**miquella — TypeScript + form accessibility | Confidence: 93/100 | APPROVE**

The form uses `role='form'` with `aria-label='Start a return'` — required because JAWS/NVDA only expose forms with explicit names as a 'form' landmark. The reason dropdown populated from REASON_LABELS ensures labels are human-readable (not raw enum keys). The return/exchange radio group must be a fieldset/legend or equivalent aria-group — the test confirms both radio inputs are present and default state. Client-side guards block submit before fetch with explicit error messages. The type-safe pre-fill props (initialOrderNumber?: string, initialEmail?: string) are optional, allowing use without URL params.

DreadPirateRobertz (commented):

Scope check ✅: cfw-9to — /returns/start page + StartReturnForm. Form covers all required fields per returns UX spec: order number, email, reason (REASON_LABELS vocab), details, return/exchange choice. Success panel shows RMA number.

Test coverage ✅: 275-line test file, 4 suites (render, submit happy path, error surfaces, client-side guards). All three error modes covered (server 4xx, network reject, non-JSON). Pre-fill prop tests present.

Dependencies: companion to PR #603 (returns API). Merge API (#603) before or with UI (#606).

Returns flow is a P2 parity gap from cf-lc1c. This closes it. APPROVE.

PR #925, opened 2026-03-29
feat(cf-y2l3): Trade-In / Trade-Up program

DreadPirateRobertz:
## Summary

- **`src/backend/tradeInService.web.js`** — `estimateTradeIn` (Anyone), `submitTradeInRequest` (Anyone), `getTradeInRequest` (Anyone), `confirmTradeIn` (Admin). Condition matrix with ±15% customer-facing range, rate limiting (3/email/24h), duplicate prevention, credit issuance via `storeCreditService`.
- **`src/public/TradeInWidget.js`** — pure PDP widget helpers: `getTradeInType` (category slug → trade-in type), `isEligible`, `buildBannerText`, `formatEstimateText`, state builders (hidden/idle/loading/estimate), `buildTradeInUrl`.
- **`src/pages/Trade In.js`** — customer form: live estimate on dropdown change, contact info + photo upload, submission, confirmation panel with requestId, status lookup by requestId + email.

## Flow

1. Visitor opens `/trade-in` (or clicks CTA from PDP via `TradeInWidget`)
2. Selects product type + condition → live credit estimate shown ($min–$max)
3. Fills name/email/phone/description + optional photos → submits
4. Receives request ID (e.g. `TI-ABCD1234`) + confirmation message
5. Brings item to store — staff confirms condition in dashboard, issues store credit via `confirmTradeIn`
6. Status checkable anytime at `/trade-in` by request ID + email

## Condition matrix (starting values — Stilgar to confirm)

| Type | Good | Fair | Poor |
|---|---|---|---|
| Futon frame | $75 | $50 | $25 |
| Futon mattress | $40 | $25 | $0 (hygiene) |
| Murphy bed | $100 | $65 | $30 |
| Platform bed | $80 | $55 | $25 |
| Sofa | $60 | $40 | $15 |

Shown to customers as ±15% range. Final credit confirmed in-store.

## CMS setup required (manual)
- `TradeInRequests`: requestId, email, name, phone, productType, condition, description, photoUrls, estimatedCredit, estimatedMin, estimatedMax, status, staffNotes, actualCondition, creditId, createdAt, updatedAt, expiresAt
- `TradeInRateLimit`: email, count, windowStart

## Test plan
- [x] creditRange ±15% math, floor at zero
- [x] CONDITION_MATRIX covers all types/conditions, mattress-poor is $0
- [x] estimateTradeIn: valid/invalid types, hygiene zero, sanitization
- [x] submitTradeInRequest: validation (name, email, type, condition), CMS insert, rate limiting (3/day, reset after 24h, fail-open)
- [x] getTradeInRequest: lookup, email mismatch → not_found, invalid inputs
- [x] confirmTradeIn: not_found, already_processed, decline, poor-mattress zero credit
- [x] TradeInWidget: all state builders, URL builder, condition options
- [x] 67 tests passing

Closes cf-y2l3

Generated with Claude Code

### DIFF
--- a/src/backend/tradeInService.web.js
+++ b/src/backend/tradeInService.web.js
@@ -1,0 +1,361 @@
+/**
+ * @module tradeInService
+ * @description Trade-In / Trade-Up program backend.
+ *
+ * Customers submit their old furniture (futon frame, mattress, murphy bed, etc.)
+ * and receive an estimated store credit range. Staff confirm in-store condition
+ * and issue the actual credit via storeCreditService.
+ *
+ * @setup (manual steps required)
+ * CMS collections:
+ *   TradeInRequests (
+ *     requestId Text indexed, email Text, name Text, phone Text,
+ *     productType Text, condition Text, description Text,
+ *     photoUrls Text (JSON array), estimatedCredit Number,
+ *     status Text: 'pending'|'confirmed'|'declined'|'expired',
+ *     staffNotes Text, actualCondition Text, creditId Text,
+ *     createdAt Date, updatedAt Date
+ *   )
+ *   TradeInRateLimit (email Text, count Number, windowStart Date)
+ * Secrets: none required — uses existing ANTHROPIC_API_KEY indirectly via storeCreditService
+ *
+ * @note Condition credit values are starting points — Stilgar to confirm final numbers.
+ */
+
+import { Permissions, webMethod } from 'wix-web-module';
+import wixData from 'wix-data';
+import { sanitize, validateEmail, validateId } from 'backend/utils/sanitize';
+
+const TRADE_IN_REQUESTS = 'TradeInRequests';
+const RATE_LIMIT_COLLECTION = 'TradeInRateLimit';
+
+// Rate limiting: max 3 submissions per email per 24h
+const RATE_LIMIT_MAX = 3;
+const RATE_LIMIT_WINDOW_MS = 24 * 60 * 60 * 1000;
+
+// Requests expire after 30 days if not confirmed
+const REQUEST_EXPIRY_DAYS = 30;
+
+// Credit ranges by product type and condition.
+// Each value is the BASE credit amount. A ±15% range is shown to customers
+// to reflect that final assessment happens in-store.
+// Values in dollars — Stilgar to confirm before go-live.
+export const CONDITION_MATRIX = {
+  'futon-frame':   { good: 75,  fair: 50,  poor: 25 },
+  'futon-mattress':{ good: 40,  fair: 25,  poor: 0  }, // poor = hygiene, no credit
+  'murphy-bed':    { good: 100, fair: 65,  poor: 30 },
+  'platform-bed':  { good: 80,  fair: 55,  poor: 25 },
+  'sofa':          { good: 60,  fair: 40,  poor: 15 },
+};
+
+export const VALID_PRODUCT_TYPES = Object.keys(CONDITION_MATRIX);
+export const VALID_CONDITIONS = ['good', 'fair', 'poor'];
+
+/** Credit range shown to customers: base ±15%, floored at 0. */
+export function creditRange(base) {
+  if (base <= 0) return { min: 0, max: 0 };
+  const delta = Math.round(base * 0.15);
+  return { min: Math.max(0, base - delta), max: base + delta };
+}
+
+// ---------------------------------------------------------------------------
+// Internal helpers
+// ---------------------------------------------------------------------------
+
+async function checkRateLimit(email, nowMs = Date.now()) {
+  try {
+    const res = await wixData.query(RATE_LIMIT_COLLECTION)
+      .eq('email', email)
+      .limit(1)
+      .find();
+    const record = res.items[0] || null;
+
+    if (!record) {
+      await wixData.insert(RATE_LIMIT_COLLECTION, {
+        email,
+        count: 1,
+        windowStart: new Date(nowMs),
+      });
+      return { allowed: true };
+    }
+
+    const windowAge = nowMs - new Date(record.windowStart).getTime();
+    if (windowAge > RATE_LIMIT_WINDOW_MS) {
+      await wixData.update(RATE_LIMIT_COLLECTION, {
+        ...record,
+        count: 1,
+        windowStart: new Date(nowMs),
+      });
+      return { allowed: true };
+    }
+
+    if (record.count >= RATE_LIMIT_MAX) {
+      return { allowed: false, reason: 'rate_limited' };
+    }
+
+    await wixData.update(RATE_LIMIT_COLLECTION, { ...record, count: record.count + 1 });
+    return { allowed: true };
+  } catch (err) {
+    console.warn('[tradeInService] rate limit check failed, allowing request:', err?.message);
+    return { allowed: true }; // Fail open
+  }
+}
+
+// ---------------------------------------------------------------------------
+// Public webMethods
+// ---------------------------------------------------------------------------
+
+/**
+ * Get the estimated store credit range for a given product type and condition.
+ * Returns { eligible: false } for unsupported product types.
+ * Returns { eligible: true, min, max, base } when credit is available.
+ * Returns { eligible: true, min: 0, max: 0, ineligible: true } for poor-condition mattresses.
+ *
+ * @param {string} productType
+ * @param {string} condition — 'good' | 'fair' | 'poor'
+ */
+export const estimateTradeIn = webMethod(
+  Permissions.Anyone,
+  async (productType, condition) => {
+    const type = sanitize(productType || '', 50).toLowerCase();
+    const cond = sanitize(condition || '', 20).toLowerCase();
+
+    if (!VALID_PRODUCT_TYPES.includes(type)) {
+      return { eligible: false };
+    }
+    if (!VALID_CONDITIONS.includes(cond)) {
+      return { eligible: false, error: 'invalid_condition' };
+    }
+
+    const base = CONDITION_MATRIX[type][cond];
+    const { min, max } = creditRange(base);
+    return { eligible: true, min, max, base, productType: type, condition: cond };
+  }
+);
+
+/**
+ * Submit a trade-in request.
+ * Anyone can submit — no login required (pre-sale / in-store flow).
+ *
+ * @param {Object} data
+ * @param {string} data.name
+ * @param {string} data.email
+ * @param {string} [data.phone]
+ * @param {string} data.productType
+ * @param {string} data.condition
+ * @param {string} [data.description]
+ * @param {string[]} [data.photoUrls]
+ * @returns {{ success: boolean, requestId?: string, estimatedCredit?: number, error?: string }}
+ */
+export const submitTradeInRequest = webMethod(
+  Permissions.Anyone,
+  async (data) => {
+    if (!data || typeof data !== 'object') {
+      return { success: false, error: 'invalid_request' };
+    }
+
+    const name = sanitize(data.name || '', 100);
+    const email = sanitize(data.email || '', 254).toLowerCase();
+    const phone = sanitize(data.phone || '', 30);
+    const productType = sanitize(data.productType || '', 50).toLowerCase();
+    const condition = sanitize(data.condition || '', 20).toLowerCase();
+    const description = sanitize(data.description || '', 1000);
+
+    // Validate photo URLs — accept only Wix media URLs or empty
+    let photoUrls = [];
+    if (Array.isArray(data.photoUrls)) {
+      const { isWixMediaUrl } = await import('backend/utils/sanitize');
+      photoUrls = data.photoUrls
+        .slice(0, 5) // max 5 photos
+        .filter(u => typeof u === 'string' && isWixMediaUrl(u));
+    }
+
+    if (!name) return { success: false, error: 'name_required' };
+    if (!validateEmail(email)) return { success: false, error: 'invalid_email' };
+    if (!VALID_PRODUCT_TYPES.includes(productType)) return { success: false, error: 'invalid_product_type' };
+    if (!VALID_CONDITIONS.includes(condition)) return { success: false, error: 'invalid_condition' };
+
+    // Rate limit by email
+    const rl = await checkRateLimit(email);
+    if (!rl.allowed) return { success: false, error: 'rate_limited' };
+
+    const base = CONDITION_MATRIX[productType][condition];
+    const { min, max } = creditRange(base);
+    const estimatedCredit = base;
+
+    const now = new Date();
+    const expiresAt = new Date(now.getTime() + REQUEST_EXPIRY_DAYS * 24 * 60 * 60 * 1000);
+
+    // Generate a short human-readable request ID (TI-XXXXXXXX)
+    const requestId = 'TI-' + Math.random().toString(36).slice(2, 10).toUpperCase();
+
+    try {
+      await wixData.insert(TRADE_IN_REQUESTS, {
+        requestId,
+        email,
+        name,
+        phone,
+        productType,
+        condition,
+        description,
+        photoUrls: JSON.stringify(photoUrls),
+        estimatedCredit,
+        estimatedMin: min,
+        estimatedMax: max,
+        status: 'pending',
+        staffNotes: '',
+        actualCondition: '',
+        creditId: '',
+        createdAt: now,
+        updatedAt: now,
+        expiresAt,
+      });
+    } catch (err) {
+      console.error('[tradeInService] insert failed:', err?.message);
+      return { success: false, error: 'submission_failed' };
+    }
+
+    return { success: true, requestId, estimatedCredit, estimatedMin: min, estimatedMax: max };
+  }
+);
+
+/**
+ * Look up a trade-in request by requestId and email (for status checking).
+ * Email verification prevents enumeration.
+ *
+ * @param {string} requestId
+ * @param {string} email
+ */
+export const getTradeInRequest = webMethod(
+  Permissions.Anyone,
+  async (requestId, email) => {
+    const cleanId = sanitize(requestId || '', 20);
+    const cleanEmail = sanitize(email || '', 254).toLowerCase();
+
+    if (!cleanId || !validateEmail(cleanEmail)) {
+      return { success: false, error: 'invalid_request' };
+    }
+
+    try {
+      const res = await wixData.query(TRADE_IN_REQUESTS)
+        .eq('requestId', cleanId)
+        .eq('email', cleanEmail)
+        .limit(1)
+        .find();
+
+      if (!res.items.length) return { success: false, error: 'not_found' };
+
+      const r = res.items[0];
+      return {
+        success: true,
+        request: {
+          requestId: r.requestId,
+          productType: r.productType,
+          condition: r.condition,
+          status: r.status,
+          estimatedMin: r.estimatedMin,
+          estimatedMax: r.estimatedMax,
+          createdAt: r.createdAt,
+          expiresAt: r.expiresAt,
+        },
+      };
+    } catch (err) {
+      console.error('[tradeInService] lookup failed:', err?.message);
+      return { success: false, error: 'lookup_failed' };
+    }
+  }
+);
+
+/**
+ * Staff confirmation: verify item in-store, set actual condition, issue store credit.
+ *
+ * @param {string} requestId
+ * @param {string} actualCondition — 'good' | 'fair' | 'poor' | 'declined'
+ * @param {string} [staffNotes]
+ */
+export const confirmTradeIn = webMethod(
+  Permissions.Admin,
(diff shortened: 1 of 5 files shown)


DreadPirateRobertz:
## Code Simplifier — dispatched by cfutons/crew/godfrey

**Agent:** code-simplifier

### Simplification Opportunities

**1. `buildBannerText` redefines `typeLabels` on every call**

`buildBannerText` creates a fresh `typeLabels` object literal on every invocation. This map is also a near-duplicate of `CATEGORY_TO_TRADE_IN_TYPE`. Extract it as a module-level constant:
```js
const TYPE_LABELS = {
  'futon-frame': 'futon frame',
  'futon-mattress': 'futon mattress',
  'murphy-bed': 'murphy bed',
  'platform-bed': 'platform bed',
  'sofa': 'sofa',
};
export function buildBannerText(tradeInType) {
  if (!isEligible(tradeInType)) return '';
  return `Trade in your old ${TYPE_LABELS[tradeInType] ?? 'furniture'} for store credit toward this purchase.`;
}
```

**2. `submitTradeInRequest` re-derives `min`/`max` then stores them separately, but the `estimateTradeIn` webMethod already does this logic**

The submit handler manually calls `creditRange(base)` and stores `estimatedMin`/`estimatedMax`. This duplicates logic that already lives in `estimateTradeIn`. Since both functions are internal to the same module, extracting the credit calculation into a shared private function (e.g., `_computeCredit(productType, condition)`) would keep the matrix lookup and range calculation in one place.

**3. `initPage` in `Trade In.js` wraps every single `` call in its own try/catch**

There are 5 separate try/catch blocks each wrapping a single line in `initPage`. This pattern repeats throughout the file. A tiny helper reduces the noise without changing behavior:
```js
function trySet(fn) { try { fn(); } catch (_) {} }
// usage:
trySet(() => ('#tradeInTitle').text = 'Trade In & Trade Up');
```
This is already an established pattern in the codebase (Wix element access is expected to fail in test/preview environments).

**4. `ELIGIBLE_PRODUCT_TYPES.includes(tradeInType ?? '')` in `isEligible` — the nullish coalescing is unnecessary**

`Array.prototype.includes` returns `false` for `null` and `undefined` since neither matches any array element. The `?? ''` can be dropped:
```js
export function isEligible(tradeInType) {
  return ELIGIBLE_PRODUCT_TYPES.includes(tradeInType);
}
```

**5. `confirmTradeIn` spreads the entire record into the update payload**

`{ ...record, status: 'declined', ... }` copies all fields including ones that should never be overwritten (e.g., `requestId`, `email`, `createdAt`). This is a common Wix CMS pattern but is worth noting — at minimum, keep it consistent across all three update call sites (declined / zero-credit / normal), which it currently is.

No blockers — suggestions 1–3 are the most impactful for readability.

DreadPirateRobertz (commented):
## Code Review — cf-y2l3: Trade-In / Trade-Up (5-agent pass)

Reviewers: Architecture | Security | Data | Frontend | Test Quality

---

### BLOCKERS (must fix before merge)

**[BLOCKER-1] Race condition in `checkRateLimit` — read-then-insert is not atomic**

`src/backend/tradeInService.web.js` lines 73–107

Two concurrent submissions from the same email can both see `record = null`, both branch into the `wixData.insert` path, and both succeed — bypassing the rate limit entirely. Wix Data has no upsert or transactional API. Fix: use a unique index on `TradeInRateLimit.email` so the second concurrent insert throws a duplicate-key error, catch that error specifically, then re-query and evaluate the existing record. The fail-open catch at line 104 must stay, but the insert-collision path needs its own handler before falling through to fail-open.

**[BLOCKER-2] `confirmTradeIn` query does not filter by `requestId` field — potential IDOR**

Lines 298–302:
```js
const res = await wixData.query(TRADE_IN_REQUESTS)
  .eq('requestId', cleanId)
  .limit(1)
  .find();
```

`requestId` is a custom application field, not the Wix CMS `_id`. The `.eq('requestId', cleanId)` filter works only if a CMS index exists on that field. Without the index, Wix may do a full-table scan and return arbitrary records — or silently return 0 results, making legitimate requests show `not_found`. More critically: if a staff member calls `confirmTradeIn` with a guessed `requestId` pattern (e.g. brute-forcing the 8-char base-36 space), there is no second-factor verification. The `Permissions.Admin` guard helps, but the design doc should explicitly state that only authenticated Wix dashboard users can reach this method, and the CMS index on `requestId` must be created as part of setup. Add this to the `@setup` block and add a test that confirms the requestId lookup returns `not_found` (not an error) for a valid-format but non-existent ID — that test is present (line 1202), but there is no test for the concurrent confirmTradeIn scenario (double-confirm race on the status check at line 309).

**[BLOCKER-3] `requestId` collision probability is unacceptably high for production volume**

Line 196:
```js
const requestId = 'TI-' + Math.random().toString(36).slice(2, 10).toUpperCase();
```

`Math.random().toString(36).slice(2, 10)` yields 8 base-36 characters = ~2.8 trillion combinations. At low volume this is fine, but there is no uniqueness check before insert. If `wixData.insert` succeeds with a duplicate `requestId` (because there is no unique constraint enforced at the CMS level), two customers will share an ID. A customer using status lookup could potentially see another customer's request if the query returns the wrong record. Mitigation: either (a) check for existence of the generated ID before inserting and retry on collision, or (b) use a cryptographically stronger ID (crypto.randomUUID() or equivalent in the Wix backend environment). Also: `Math.random()` is not cryptographically secure — fine for a human-readable reference ID, but document that explicitly.

**[BLOCKER-4] Internal name "Stilgar" appears in the public-repo docblock**

Lines 28 and 48 of `src/backend/tradeInService.web.js`:
```
* @note Condition credit values are starting points — Stilgar to confirm final numbers.
```
```
// Values in dollars — Stilgar to confirm before go-live.
```

This repo is public (DreadPirateRobertz/carolina-futons). Internal operator names must not appear in committed code. Replace with a neutral action item, e.g. `// TODO: confirm credit values with store owner before go-live.`

---

### IMPORTANT (should fix before merge)

**[IMP-1] Phone number collected with no TCPA-compliant consent language**

`src/pages/Trade In.js` line 458:
```js
$w('#phoneInput').accessibility.ariaLabel = 'Your phone number (optional)';
```

Phone is collected as optional but there is no visible consent copy near the phone field indicating how it will be used, and no opt-in checkbox for SMS or phone contact. If staff ever follow up by phone or text (even informally), this is a TCPA exposure. The form page should include static copy such as "By providing your phone number you consent to be contacted regarding your trade-in request. We will not send marketing messages without separate consent." This is a legal requirement, not a UX suggestion.

**[IMP-2] `onItemReady` / event handler registration order in the page module**

`src/pages/Trade In.js` lines 450–452:
```js
$w('#productTypeDropdown').onChange(() => fetchEstimate());
$w('#conditionDropdown').onChange(() => fetchEstimate());
```

These `onChange` registrations happen inside `initEstimateForm()`, which is called synchronously from `$w.onReady`. `prefillFromQueryParams()` (line 409) is called after `initEstimateForm()` and calls `fetchEstimate()` directly if both params are present. This ordering is correct. However, `initStatusLookup()` hides `#statusSection` (line 474) — if the element is already hidden in the editor, calling `.hide()` on it again is harmless but the `.show()` in the `#showStatusLookupBtn` onClick (line 478) is registered inside a try/catch that silently swallows errors if `#showStatusLookupBtn` does not exist. This means if the button is missing from the CMS page build, the status lookup section becomes permanently hidden with no error surfaced to developers. At minimum log a warning in the catch block.

**[IMP-3] Photo URL validation imports `isWixMediaUrl` dynamically inside the method body**

Lines 173–176:
```js
const { isWixMediaUrl } = await import('backend/utils/sanitize');
```

`sanitize` is already statically imported at line 33. This dynamic re-import of the same module is unnecessary and will cause a redundant module evaluation in environments where module caching is not guaranteed. Move `isWixMediaUrl` into the static import at the top of the file.

**[IMP-4] `confirmTradeIn` — credit issuance can fail after `status` is already checked as `pending`, leaving record in an inconsistent state**

Lines 342–363: If `issueStoreCredit` throws (line 352), the function returns `credit_issuance_failed` without updating the record. The record remains `pending`. This is recoverable — staff can retry. But if `issueStoreCredit` succeeds and then the subsequent `wixData.update` at line 356 throws, the credit has been issued but the record still shows `pending`. On retry, `issueStoreCredit` would issue a second credit. Mitigation: update the record status to `processing` (or similar) before calling `issueStoreCredit`, so a retry sees `already_processed`. Document this gap in a code comment at minimum.

**[IMP-5] `getTradeInRequest` response leaks `expiresAt` to the client**

Line 265 returns `expiresAt`. This is low-sensitivity information, but combined with `createdAt` it reveals internal system timing. More importantly, there is no server-side check for expired status. A request past `expiresAt` still shows as `pending` to the customer. Add expiry evaluation server-side and return `status: 'expired'` if `expiresAt < now`.

---

### SUGGESTIONS (nice to have)

**[SUG-1] Test coverage gap: no test for concurrent `confirmTradeIn` (double-process race)**

The `already_processed` guard is tested for a pre-seeded `confirmed` record, but not for two simultaneous calls both reading `pending` before either writes. This is a known gap given Wix Data's non-transactional nature — add a comment acknowledging it.

**[SUG-2] No test for `Trade In.js` page module**

The page module is 317 lines with submit flow, status lookup, error display, and query-param prefill. None of this is covered by the 67 tests (which cover only the backend service and the pure widget helpers). Wix page modules are testable with `$w` mock injection. At minimum: `handleSubmit` happy path, rate-limit error display, and `prefillFromQueryParams` with valid/invalid params.

**[SUG-3] `_submitting` flag is module-level state**

Line 402: `let _submitting = false;` — this is fine for a single-page Wix module but should be documented. In Wix Velo, each page instance gets its own module scope, so this is not a cross-user issue, but it could confuse maintainers who expect stateless modules.

**[SUG-4] `status: confirmed` message in status lookup uses `estimatedMin/Max`, not `actualCondition` credit**

Line 649:
```js
confirmed: `Confirmed — $${request.estimatedMin ?? 0}–$${request.estimatedMax ?? 0} store credit issued.`
```

The issued credit is `base` (the exact matrix value), not the estimated range. The `getTradeInRequest` response does not include `actualCondition` or the actual credit amount, so the UI shows the estimated range even after confirmation. Consider adding `creditAmount` to the `getTradeInRequest` response shape for confirmed requests so the customer sees the real number.

---

### What is done well

- `Permissions.Admin` correctly guards `confirmTradeIn`. The public methods use `Permissions.Anyone` appropriately for the anonymous submission flow.
- Input sanitization is consistent across all public web methods — both server-side (backend sanitize helpers) and client-side (`sanitizeText` from public/validators).
- The `getTradeInRequest` email-plus-ID two-factor lookup is the correct pattern for preventing IDOR on status checks.
- Rate limiting logic is well-structured with explicit window reset and a documented fail-open policy.
- `TradeInWidget.js` is a clean pure-function module with no side effects — good separation of concerns.
- Test suite covers all sad paths for the backend service (not just happy path). The `already_processed` and `not_found` branches are explicit.
- A11y: `ariaLabel` on all form inputs and `announce()` calls on estimate and confirmation are good practice.

---

### Summary

4 blockers, 5 important issues. The race condition in rate limiting (BLOCKER-1), the requestId collision gap (BLOCKER-3), and the internal name exposure (BLOCKER-4) are straightforward fixes. BLOCKER-2 (IDOR/index) requires a CMS setup note addition and a comment, not a code rewrite. The TCPA phone consent issue (IMP-1) requires copy on the page — no backend change. Resolve blockers before merge. IMP-4 (credit double-issue race) is the most operationally risky of the important items and should be addressed before go-live even if merged first.

Confidence: 82/100

DreadPirateRobertz (commented):
**Update from PM**: Overseer confirms 'Stilgar' references in docblocks are acceptable — blocker #4 is cleared. Remaining blockers: (1) checkRateLimit race condition, (2) confirmTradeIn IDOR + CMS index docs + double-confirm race, (3) requestId collision → use crypto.randomUUID(). Credit double-issue race in confirmTradeIn also needs addressing before merge.

DreadPirateRobertz:
## silent-failure-hunter (dispatched by godfrey)

Reviewed commit 9eba88d6. Four findings, one of them critical.

---

### CRITICAL — Stage 3 failure after credit issuance: money gone, record not updated, no reconciliation log

**Location:** `src/backend/tradeInService.web.js` lines 404–424

**What happens:** If the Stage 3 `wixData.update` (writing `status: 'credited'`) throws an exception, execution falls directly into the outer `catch (err)` at the bottom of `confirmTradeIn`:

```js
} catch (err) {
  console.error('[tradeInService] Error confirming trade-in:', err);
  return { success: false, message: 'Failed to confirm trade-in.' };
}
```

At this point, `issueStoreCredit` has already succeeded. The customer has live store credit in their account. But the `TradeInRequests` record is stuck at status `'confirmed'` with no `storeCreditId` written to it. The log message says "Error confirming trade-in" — it says nothing about the orphaned credit. The staff member sees a generic failure message and has no idea money already went out the door.

There is no reconciliation log. There is no alert. A developer debugging this 6 months from now will have to audit `storeCreditService` logs against `TradeInRequests` records manually to find the discrepancy.

**Recommendation:** Catch Stage 3 failure separately, before the outer catch, and log with full context including the credit that was already issued:

```js
let stage3Error;
try {
  await wixData.update(COLLECTION, {
    ...request,
    status: 'credited',
    confirmedCondition: cond,
    issuedCreditAmount: creditAmount,
    storeCreditId,
    staffNotes,
    confirmedAt: new Date(),
  });
} catch (err) {
  stage3Error = err;
  // RECONCILIATION: credit was issued but record update failed.
  // Log with full context so staff/ops can reconcile manually.
  logError('[tradeInService] RECONCILIATION REQUIRED: store credit issued but status update failed', {
    errorId: errorIds.TRADE_IN_RECONCILIATION_REQUIRED,
    requestId: cleanId,
    memberId: request.memberId,
    storeCreditId,
    creditAmount,
    err,
  });
  return {
    success: false,
    message: `Store credit of $${creditAmount} was issued (ID: ${storeCreditId}) but the record could not be updated to 'credited'. Contact admin with request ID ${cleanId} for manual reconciliation.`,
  };
}
```

---

### HIGH — Stage 2 failure log omits request ID

**Location:** `src/backend/tradeInService.web.js` lines 394–398

```js
if (!creditResult.success) {
  console.error('[tradeInService] Failed to issue store credit:', creditResult.message);
  return { success: false, message: 'Trade-in confirmed but store credit issuance failed. Contact admin.' };
}
```

The `console.error` contains no `requestId`, no `memberId`, no `creditAmount`. When this fires in production, a developer searching logs will find the message but will have no way to identify which request is stuck in 'confirmed' limbo without a manual CMS scan.

Additionally, this uses `console.error` rather than `logError` with a Sentry error ID. A Stage 2 failure is a financial event requiring human follow-up — it should land in Sentry with full context, not just in server-side console output.

**Recommendation:**

```js
if (!creditResult.success) {
  logError('[tradeInService] Failed to issue store credit after confirmation', {
    errorId: errorIds.TRADE_IN_CREDIT_ISSUANCE_FAILED,
    requestId: cleanId,
    memberId: request.memberId,
    creditAmount,
    reason: creditResult.message,
  });
  return {
    success: false,
    message: `Trade-in confirmed (request ${cleanId}) but store credit issuance failed. Contact admin.`,
  };
}
```

---

### HIGH — _refreshEstimate catch block provides zero user feedback

**Location:** `src/pages/Trade In.js` lines 88–94

```js
try {
  result = await getTradeInValuation(_itemType, _condition);
} catch (err) {
  console.error('[tradeInPage] Valuation error:', err);
  _valuation = null;
  return;
}
```

When `getTradeInValuation` throws, the function logs to console and returns. By this point, both `#estimateBox` and `#eligibilityError` have already been hidden (lines 84–85). The user selected a type and condition, watched both panels disappear, and then nothing appeared — no estimate, no error message, no indication that anything went wrong. The user has no idea whether the form is processing, broken, or waiting for input. This will cause abandonment.

**Recommendation:** Show a user-facing error before returning:

```js
} catch (err) {
  console.error('[tradeInPage] Valuation error:', err);
  _valuation = null;
  try {
    $w('#eligibilityError').text = 'Unable to load estimate. Please try again or refresh the page.';
    $w('#eligibilityError').show();
  } catch (_) { /* */ }
  announce('Unable to load estimate. Please try again.');
  return;
}
```

---

### MEDIUM — Outer catch in confirmTradeIn gives no stage context

**Location:** `src/backend/tradeInService.web.js` lines 422–425

The single outer `try/catch` wraps all three stages. If Stage 1's `wixData.update` throws, the staff member receives "Failed to confirm trade-in." — identical to a Stage 3 failure. There is no way from the error message or log to know whether the failure was pre-credit (safe to retry) or post-credit (requires reconciliation). Since Stage 3 should be caught separately per the CRITICAL finding above, this medium finding becomes less severe — but it is worth noting that Stage 1 failures should also log the `requestId` for traceability, which the current outer catch does not do.

---

### Summary table

|---|---|---|---|
| 1 | CRITICAL | `tradeInService.web.js:404–424` | Stage 3 failure after credit issued: no reconciliation log, generic error to caller |
| 2 | HIGH | `tradeInService.web.js:394–398` | Stage 2 failure log omits requestId/memberId; uses console.error not logError |
| 3 | HIGH | `Trade In.js:88–94` | Valuation catch returns silently with zero user feedback |
| 4 | MEDIUM | `tradeInService.web.js:422–425` | Outer catch gives no stage context; requestId not included in error log |

The 3-stage idempotency design is the right approach — the logic is sound. The gap is purely in the error paths: Stage 2's catch is logged but under-contextualized, and Stage 3's catch is handled by the outer catch which has no knowledge that a credit already exists.

DreadPirateRobertz:
## code-reviewer (dispatched by godfrey)

**Reviewing commit 9eba88d6** — `fix(cf-y2l3): harden photo URL filter and confirmTradeIn idempotency`

Files reviewed:
- `src/backend/tradeInService.web.js`
- `src/pages/Trade In.js`

---

### Critical (confidence 95)

**Double credit issuance on retry in `confirmTradeIn` 3-stage flow**

`src/backend/tradeInService.web.js`, lines ~370-400

The idempotency guard is incomplete. Consider this scenario:

1. Staff calls `confirmTradeIn` on a pending request
2. Stage 1 sets status to `confirmed` (succeeds)
3. Stage 2 calls `issueStoreCredit` (succeeds — credit issued)
4. Stage 3 `wixData.update` to `credited` **fails** (network error, timeout)
5. Staff retries — status is `confirmed`, allowed through
6. Stage 1 is skipped (correct)
7. Stage 2 runs **again** — `issueStoreCredit` called a second time
8. Second credit issued to the customer

The `orderReference: 'trade-in:${cleanId}'` could provide dedup if `issueStoreCredit` enforces idempotency on that key, but this code does not check or rely on that. There is also no check for `request.storeCreditId` already being populated from a prior successful Stage 2.

**Fix**: Before Stage 2, check if the record already has a `storeCreditId`. If it does, skip credit issuance and proceed to Stage 3:

```js
let storeCreditId = request.storeCreditId || null;
if (!storeCreditId && request.memberId) {
  // Issue store credit only if not already issued
  const creditResult = await issueStoreCredit({ ... });
  // ...
  storeCreditId = creditResult.creditId;
}
```

---

### Critical (confidence 90)

**Confirmed request can be moved to `rejected` after credit was already issued**

`src/backend/tradeInService.web.js`, lines ~353-364

The rejection branch runs before the Stage 1 guard. If a request is in `confirmed` status (meaning Stage 2 credit issuance may have already succeeded on a prior call), and a staff member retries with a condition that evaluates as ineligible, the record is set to `rejected` — but the already-issued store credit remains active.

**Fix**: When `request.status === 'confirmed'`, skip the rejection path or at minimum check whether credit was already issued before overwriting status:

```js
if (!val.eligible) {
  if (request.status === 'confirmed' && request.storeCreditId) {
    return { success: false, message: 'Cannot reject — credit already issued. Void the credit first.' };
  }
  // ... existing rejection logic
}
```

---

### Important (confidence 85)

**`_refreshEstimate` catch leaves UI blank — silent failure for the user**

`src/pages/Trade In.js`, lines ~90-95

When `getTradeInValuation` throws, both `#estimateBox` and `#eligibilityError` were already hidden (lines 85-86), and the catch block returns without showing any error UI. The user changes a dropdown, the previous estimate disappears, and nothing replaces it.

**Fix**: Show the eligibility error element with a generic message before returning:

```js
} catch (err) {
  console.error('[tradeInPage] Valuation error:', err);
  _valuation = null;
  try { $w('#eligibilityError').text = 'Unable to calculate estimate. Please try again.'; $w('#eligibilityError').show(); } catch (_) { /* */ }
  return;
}
```

---

### No issues found (confidence notes)

- **HTTPS-only photo URL filter** (line ~197): Good hardening. `url.startsWith('https://')` correctly prevents `javascript:`, `data:`, and `http://` schemes. The empty-string case is also handled since empty strings don't start with `https://`. No issue here.
- **`staffNotes` extraction to module-level variable**: Clean refactor, no semantic change.

PR #1002, opened 2026-04-05
fix: add suppressAuth:true to all surveyService queries

DreadPirateRobertz:
## Summary

- Add `{ suppressAuth: true }` to all 5 `wixData.query().find()` calls in `surveyService.web.js`
- Affected methods: `submitSurveyResponse`, `getSurveyForOrder`, `getSurveyResponseAggregation` (2 queries), `getNpsStats`
- Without this, queries fail silently when CMS collection has admin-only read permissions

## Test plan

- [x] 4 new TDD tests assert `find()` receives `{ suppressAuth: true }` for each query path
- [x] All 90 survey-related tests pass
- [x] No functional changes — only auth bypass for backend queries

### DIFF
--- a/src/backend/surveyService.web.js
+++ b/src/backend/surveyService.web.js
@@ -168,7 +168,7 @@ export const submitSurveyResponse = webMethod(
       const result = await wixData.query(SURVEY_COLLECTION)
         .eq('memberId', memberId)
         .eq('orderId', orderId)
-        .find();
+        .find({ suppressAuth: true });
 
       if (result.items.length === 0) {
         return { success: false, error: 'No survey found for this order' };
@@ -224,7 +224,7 @@ export const getSurveyForOrder = webMethod(
       const result = await wixData.query(SURVEY_COLLECTION)
         .eq('memberId', memberId)
         .eq('orderId', cleanId)
-        .find();
+        .find({ suppressAuth: true });
 
       if (result.items.length === 0) return { success: true, survey: null };
 
@@ -278,11 +278,11 @@ export const getSurveyResponseAggregation = webMethod(
           .ge('sentAt', since)
           .isNotEmpty('completedAt')
           .limit(1000)
-          .find(),
+          .find({ suppressAuth: true }),
         wixData.query(SURVEY_COLLECTION)
           .ge('sentAt', since)
           .limit(1000)
-          .find(),
+          .find({ suppressAuth: true }),
       ]);
 
       const completed = completedResult.items;
@@ -354,7 +354,7 @@ export const getNpsStats = webMethod(
         .ge('completedAt', since)
         .isNotEmpty('completedAt')
         .limit(1000)
-        .find();
+        .find({ suppressAuth: true });
 
       const responses = result.items;
       if (responses.length === 0) {
--- a/tests/cf-1mlj-survey-aggregation.test.js
+++ b/tests/cf-1mlj-survey-aggregation.test.js
@@ -17,7 +17,7 @@
  *  - returns error on DB failure
  */
 import { describe, it, expect, beforeEach } from 'vitest';
-import { __seed, __reset, __setQueryError } from './__mocks__/wix-data.js';
+import { __seed, __reset, __setQueryError, __getLastFindOptions } from './__mocks__/wix-data.js';
 import { getSurveyResponseAggregation } from '../src/backend/surveyService.web.js';
 
 // ── Fixtures ──────────────────────────────────────────────────────────────────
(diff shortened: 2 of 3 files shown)


DreadPirateRobertz (commented):
## Code Review — fix: add suppressAuth:true to all surveyService queries

**Confidence: 90 — LGTM with one gap to be aware of**

### Query coverage — all 5 paths confirmed

Every `wixData.query(...).find()` call in `src/backend/surveyService.web.js` now passes `{ suppressAuth: true }`. Confirmed by inspecting the PR branch directly:

- Line 171: `submitSurveyResponse` — covered
- Line 227: `getSurveyForOrder` — covered
- Line 281: `getSurveyResponseAggregation` parallel query 1 (completed) — covered
- Line 285: `getSurveyResponseAggregation` parallel query 2 (all sent) — covered
- Line 357: `getNpsStats` — covered

No `.find(` call without `suppressAuth` remains in the file.

### Test assertions

Three tests in `tests/surveyService.test.js` (lines 353–367) use `mockFind` (a `vi.fn()` spy) and assert `toHaveBeenCalledWith({ suppressAuth: true })` for `submitSurveyResponse`, `getSurveyForOrder`, and `getNpsStats`. The mock chain is correctly set up with `find: mockFind` so the assertion fires on the actual `.find()` call.

One test in `tests/cf-1mlj-survey-aggregation.test.js` (line 252) uses `__getLastFindOptions('SurveyResponses')` from the real wix-data mock and asserts `{ suppressAuth: true }`. The mock correctly captures options at the `find()` call site (line 175 of the mock) and resets on `__reset()`.

### Gap: aggregation parallel-query coverage is partial (confidence 85)

`getSurveyResponseAggregation` issues two parallel `wixData.query('SurveyResponses').find()` calls — one filtered `.isNotEmpty('completedAt')` (completed surveys) and one unfiltered (all sent). `__getLastFindOptions` records only the **last** options written for a given collection key. Since both queries target the same collection string `'SurveyResponses'`, the test only verifies that whichever of the two resolves last passed `suppressAuth: true`. In practice both do, but only one call is confirmed by the assertion.

This is not a bug — both lines in the source are correctly patched — but if a future edit accidentally removes `suppressAuth` from one of the two parallel calls, the test will still pass as long as the other call runs last. A second assertion that seeds distinct collections or uses `toHaveBeenNthCalledWith` would close this. Non-blocking for this PR; suggest a follow-on bead or a comment in the test.

### Summary

All 5 query paths are patched. Three of the four new tests are airtight. The aggregation test covers the correct behavior but has a structural limitation on parallel same-collection queries. No correctness issue in the shipped code. Ready to merge.

DreadPirateRobertz (commented):
## Code Review — fix: add suppressAuth:true to all surveyService queries

**Confidence score: 88 — APPROVE with one noted gap**

### What was done well

The four production `.find()` calls that needed the fix all received `{ suppressAuth: true }`:
- `submitSurveyResponse` — line 171
- `getSurveyForOrder` — line 227
- `getSurveyResponseAggregation` — both parallel queries (lines 281, 285)
- `getNpsStats` — line 357

Test coverage is solid and uses two distinct patterns appropriately: `mockFind` call assertion in `surveyService.test.js` for the simpler cases, and `__getLastFindOptions` in `cf-1mlj-survey-aggregation.test.js` for the aggregation query. All three critical paths (member query, admin aggregation, NPS stats) have explicit suppressAuth assertions.

### Findings

**Important (confidence 88): refinery/rig surveyService is missing suppressAuth on all .find() calls**

`/Users/hal/gt/cfutons/refinery/rig/src/backend/surveyService.web.js` has bare `.find()` calls at lines 171, 227, and 272 with no `{ suppressAuth: true }`. The rig is self-contained (its tests import from within the rig), but the rig test suite exercises the same code paths and these will silently operate without the auth suppression during local test runs. If the rig is used for integration smoke tests this is a behavioral discrepancy from production.

The rig `surveyService.web.js` also appears to be behind the production version in scope (the aggregation function `getSurveyResponseAggregation` is present in production but missing from the rig), suggesting the rig is intentionally an older snapshot. If that is the intended relationship, document it. If the rig is meant to mirror production, it needs to be brought forward as a follow-on bead.

**Suggestion (confidence 80): no suppressAuth assertion for getSurveyResponseAggregation's second parallel query**

The test in `cf-1mlj-survey-aggregation.test.js` uses `__getLastFindOptions('SurveyResponses')` which by definition only captures the last recorded call. The aggregation fires two parallel `.find()` calls on `SURVEY_COLLECTION`. Depending on how `__getLastFindOptions` is implemented in the mock, it may only verify one of the two. If the mock records per-collection and overwrites on each call, the first parallel query's options would be shadowed by the second. This is worth confirming — if the mock only captures the final call, a test for the first parallel branch is missing.

### Summary

Production code and primary test paths are correct and complete. The rig divergence and the parallel-query test coverage edge are the two items worth tracking. Neither blocks this merge.

PR #1003, opened 2026-04-05
fix(L4-audit): remove private constant exports from trailPerkService

DreadPirateRobertz:
## Summary
- Remove `_VALID_PERK_IDS`, `_PERK_EMAIL_TEMPLATES`, `_TRAIL_PERKS_COLLECTION` exports from `trailPerkService.web.js` — these leaked internal constants into the public webMethod API
- Tests updated to use local constants instead of importing private internals
- 41/41 tests green across both trail perk test files

## Test plan
- [x] `tests/trailPerkService.test.js` — 22 pass
- [x] `tests/trailPerkServiceWebMethods.test.js` — 19 pass
- [x] No remaining references to removed exports (`grep` clean)

### DIFF
--- a/src/backend/trailPerkService.web.js
+++ b/src/backend/trailPerkService.web.js
@@ -142,11 +142,6 @@ export const deliverTrailPerk = webMethod(
   }
 );
 
-// Export for testing
-export const _VALID_PERK_IDS = VALID_PERK_IDS;
-export const _PERK_EMAIL_TEMPLATES = PERK_EMAIL_TEMPLATES;
-export const _TRAIL_PERKS_COLLECTION = TRAIL_PERKS_COLLECTION;
-
 // ── getAvailableTrailPerks [Anyone] ──────────────────────────────────────────
 
 export const getAvailableTrailPerks = webMethod(
--- a/tests/trailPerkService.test.js
+++ b/tests/trailPerkService.test.js
@@ -12,10 +12,9 @@ import {
 } from './__mocks__/wix-data.js';
 import {
   deliverTrailPerk,
-  _VALID_PERK_IDS,
-  _PERK_EMAIL_TEMPLATES,
-  _TRAIL_PERKS_COLLECTION,
 } from '../src/backend/trailPerkService.web.js';
+
+const COLLECTION = 'MemberTrailPerks';
 
 beforeEach(() => {
   __reset();
@@ -49,7 +48,7 @@ describe('deliverTrailPerk — perk-free-shipping', () => {
   it('inserts a MemberTrailPerks record', async () => {
     const r = await deliverTrailPerk('mem-1', 'perk-free-shipping', 'a@b.com');
     expect(r.success).toBe(true);
-    const inserted = __getInserted(_TRAIL_PERKS_COLLECTION);
+    const inserted = __getInserted(COLLECTION);
     expect(inserted).toHaveLength(1);
     expect(inserted[0].perkId).toBe('perk-free-shipping');
     expect(inserted[0].memberId).toBe('mem-1');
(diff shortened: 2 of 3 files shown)


DreadPirateRobertz (commented):
## Code Review — fix(L4-audit): remove private constant exports from trailPerkService

**Confidence: 95 — LGTM with one observation**

### What was verified

- All 3 underscore exports (`_VALID_PERK_IDS`, `_PERK_EMAIL_TEMPLATES`, `_TRAIL_PERKS_COLLECTION`) are fully removed from `src/backend/trailPerkService.web.js`. No remaining `export const _` lines on the PR branch.
- Both test files (`trailPerkService.test.js` and `trailPerkServiceWebMethods.test.js`) drop the imported symbols and introduce local replacements. No remaining imports of the removed exports.
- A repo-wide grep confirms no other file in the codebase still imports or references any of the three removed names.

### Test replacement quality

The local substitutions are sound:

- `tests/trailPerkService.test.js` replaces `_TRAIL_PERKS_COLLECTION` with a module-local `const COLLECTION = 'MemberTrailPerks'` — the string literal is correct, matches the collection name used in the service.
- `tests/trailPerkServiceWebMethods.test.js` replaces `_VALID_PERK_IDS` with a locally-declared `KNOWN_PERK_IDS` Set containing the same three perk IDs. The perk-ID validation test at line 58 continues to assert the same invariant.

### One observation (non-blocking)

`KNOWN_PERK_IDS` in the WebMethods test is now a second copy of the canonical perk-ID set. If a perk is added to the service, the test will silently pass for the new perk (it checks `has()`, which won't fail on an unknown ID from the service's perspective). This was already the case with the old exported set — the PR does not make it worse — but it is worth noting as a follow-on: the `getAvailableTrailPerks` test would be stronger if it also asserts `result.perks.length === KNOWN_PERK_IDS.size`. Low priority; not a blocker.

### Summary

De-export is complete and clean. Test coverage is maintained. Ready to merge.

DreadPirateRobertz (commented):
## Code Review — fix(L4-audit): remove private constant exports from trailPerkService

**Confidence score: 92 — APPROVE**

### What was done well

The removal is clean and complete for the production files. The test files correctly replace the exported constants with locally-scoped literals (`COLLECTION = 'MemberTrailPerks'`, `KNOWN_PERK_IDS`), which is the right approach — tests should not depend on production modules re-exporting private implementation details just to provide test fixtures.

### Findings

**Important (confidence 90): refinery/rig copies still carry the removed exports**

`/Users/hal/gt/cfutons/refinery/rig/src/backend/trailPerkService.web.js` still exports `_VALID_PERK_IDS`, `_PERK_EMAIL_TEMPLATES`, and `_TRAIL_PERKS_COLLECTION` (lines 141–143), and the corresponding rig test at `refinery/rig/tests/trailPerkService.test.js` still imports and uses all three.

The rig is a self-contained test harness (its tests import from `../src/backend/` within the rig directory), so this does not break anything today. However, if the rig copies are meant to stay in sync with the production source, these stale exports are L4-audit violations waiting to be flagged in the next pass. The security concern is real: `_PERK_EMAIL_TEMPLATES` in particular exposes internal template keys via export even in the rig.

**Recommendation:** Either sync the rig copies in this same PR, or open a follow-on bead to clean the rig in a dedicated pass. Do not leave the L4 audit half-done.

**Suggestion (confidence 82): KNOWN_PERK_IDS in tests will silently drift**

`KNOWN_PERK_IDS` is now a manually maintained `Set` in `trailPerkServiceWebMethods.test.js`. If a perk is added to the production constant but the test set is not updated, the test `all returned perk IDs are valid` will pass with a false positive (the new perk ID will simply not be in the set). A comment noting this drift risk, or an additional test asserting the count of returned perks matches expectations, would reduce that hazard.

### Summary

Production source and top-level test files are correct. The rig copies are the open item. Approve contingent on the rig being tracked as a follow-on (or included here).

PR #1014, opened 2026-04-13
feat(CF-e50): YouMightAlsoLike — 4-item similar-products grid on PDP

DreadPirateRobertz:
## Summary
- Adds `YouMightAlsoLike.js` (`initYouMightAlsoLike`) — compact "You might also like" 4-item repeater grid for the Product Detail Page
- Queries `getSimilarProducts(productId, { limit: 4 })` from `backend/productRecommendations.web`
- Collapses `#youMightAlsoLikeSection` on null state, missing `_id`, `success:false`, empty results, or backend error
- Registers `onItemReady` before setting `.data` (Wix repeater requirement), caps grid at 4 items
- Item rendering: name, formatted price (call-for-price aware), image src (skipped when `mainMedia` is null)

## Test plan
- [x] 6 collapse-path tests (null state, no product, no `_id`, success:false, empty array, throws)
- [x] 5 happy-path tests (expand, 4-item data, cap at 4, productId forwarding with `{limit:4}`, onItemReady ordering)
- [x] 5 onItemReady rendering tests (name, price, call-for-price, image src set, null mainMedia no-op)
- [ ] CI green before merge

Closes CF-e50

### DIFF
--- a/src/public/YouMightAlsoLike.js
+++ b/src/public/YouMightAlsoLike.js
@@ -1,0 +1,63 @@
+/**
+ * @module YouMightAlsoLike
+ * @description "You might also like" 4-item product grid for the Product Detail Page.
+ * Queries similar products by category/material and renders a repeater grid.
+ * Collapses the section for guests, errors, or empty results.
+ *
+ * Elements:
+ *   #youMightAlsoLikeSection — Box/Strip wrapper; collapsed when no products
+ *   #youMightAlsoLikeGrid    — Repeater; receives up to 4 product items
+ *   #ymItem_image            — Image; product main media
+ *   #ymItem_name             — Text; product name
+ *   #ymItem_price            — Text; formatted price or "Call for Price"
+ *
+ * CF-e50
+ */
+import { getSimilarProducts as _defaultGetSimilarProducts } from 'backend/productRecommendations.web';
+import { isCallForPrice, CALL_FOR_PRICE_TEXT } from 'public/productPageUtils.js';
+
+/**
+ * Initialise the "You might also like" grid on the Product Detail Page.
+ *
+ * @param {Function} $w    - Wix element selector
+ * @param {Object}   state - Page state; must contain state.product._id
+ * @param {Object}  [opts] - Injectable overrides for testing
+ * @param {Function} [opts.getSimilarProducts] - Backend getSimilarProducts(productId, options)
+ * @returns {Promise<void>}
+ */
+export async function initYouMightAlsoLike($w, state, opts = {}) {
+  const getSimilarProducts = opts.getSimilarProducts ?? _defaultGetSimilarProducts;
+
+  const productId = state?.product?._id;
+  if (!productId) {
+    $w('#youMightAlsoLikeSection').collapse();
+    return;
+  }
+
+  let products;
+  try {
+    const result = await getSimilarProducts(productId, { limit: 4 });
+    if (!result?.success || !result.products?.length) {
+      $w('#youMightAlsoLikeSection').collapse();
+      return;
+    }
+    products = result.products.slice(0, 4);
+  } catch (_) {
+    $w('#youMightAlsoLikeSection').collapse();
+    return;
+  }
+
+  // Register onItemReady BEFORE setting data (Wix repeater requirement).
+  $w('#youMightAlsoLikeGrid').onItemReady(($item, itemData) => {
+    $item('#ymItem_name').text = itemData.name;
+    $item('#ymItem_price').text = isCallForPrice(itemData)
+      ? CALL_FOR_PRICE_TEXT
+      : itemData.formattedPrice;
+    if (itemData.mainMedia) {
+      $item('#ymItem_image').src = itemData.mainMedia;
+    }
+  });
+
+  $w('#youMightAlsoLikeGrid').data = products;
+  $w('#youMightAlsoLikeSection').expand();
+}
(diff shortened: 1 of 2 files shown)


DreadPirateRobertz (commented):
## Code Review — CF-e50: YouMightAlsoLike (Senior Reviewer)

Confidence scoring applied: only issues >=80 reported.

---

### What was done well

- Module JSDoc is thorough: element IDs, dependency, collapse contract, and CF bead reference all documented.
- `initYouMightAlsoLike` has complete `@param`/`@returns` JSDoc including the injectable `opts.getSimilarProducts` override pattern — this is the right approach for testability.
- The WHY-comment on `onItemReady` ordering (`// Must register onItemReady BEFORE setting .data`) is exactly the kind of non-obvious Wix platform note that prevents future regressions.
- Defensive cap (`products.slice(0, 4)`) with an inline comment explaining the reason (backend may ignore limit) is good defensive programming.
- Collapse-on-all-error-paths is complete: null state, missing `_id`, `success:false`, empty array, null result, and thrown exception are all covered.
- Test coverage is wide: 8 collapse paths, 5 happy-path cases, 6 `onItemReady` rendering tests. The `onItemReady`-before-`data` ordering test using `callOrder` is particularly well designed.
- Per-field try/catch in `onItemReady` with `console.warn` on each path prevents one broken item from collapsing the entire grid.

---

### Issues

#### BLOCKING — PR test plan has unchecked checkbox (confidence: 99)

`- [ ] CI green before merge` is unchecked. Per project standard (`feedback_pr_test_plan_standard.md`): all `- [ ]` checkboxes must be checked before merge. Do not merge until CI passes and this box is ticked.

---

#### Important — Navigation `.catch()` on a potentially non-Promise return (confidence: 90)

**File**: `src/public/YouMightAlsoLike.js`, line ~78

```js
wixLocationTo(`/${itemData.slug}`).catch(e =>
  console.warn('[YouMightAlsoLike] navigation failed:', e?.message ?? e)
);
```

The global `tests/__mocks__/wix-location-frontend.js` mock has `to()` returning `undefined` synchronously (no return value). Calling `.catch()` on `undefined` throws `TypeError: Cannot read properties of undefined (reading 'catch')` at runtime if `wix-location-frontend`'s `to()` ever returns non-Promise. The test works because the test-local `vi.mock` overrides the global mock with `mockTo.mockResolvedValue(undefined)` — but this masks the mismatch.

Recommendation: guard with optional chaining:
```js
wixLocationTo(`/${itemData.slug}`)?.catch(e =>
  console.warn('[YouMightAlsoLike] navigation failed:', e?.message ?? e)
);
```
Or verify that `wix-location-frontend`'s `to()` always returns a Promise in the Wix platform contract and add a comment stating this assumption.

---

#### Important — Test-local `vi.mock` shadows global `__mocks__` for `wix-location-frontend` (confidence: 86)

**File**: `tests/youMightAlsoLike.test.js`

The test declares a full `vi.mock('wix-location-frontend', ...)` factory that provides `{ default: { to: vi.fn() }, to: vi.fn() }`. This bypasses the project-wide `tests/__mocks__/wix-location-frontend.js` which has `__reset()`, `__getToCallLog()`, and proper binding. The local mock and the global mock have different behavior: the global `to()` is synchronous, the local mock returns a resolved Promise.

This creates a gap: the test passes with a Promise-returning mock but the global mock (used everywhere else) is synchronous. If the same navigation pattern is copy-pasted to other modules and tested with the global mock, the `.catch()` call will throw.

Recommendation: Use the global mock (`__mocks__/wix-location-frontend.js`) and its `__resetToCallLog()` / `__getToCallLog()` helpers in this test, consistent with the rest of the test suite.

---

#### Suggestion — Unused `wix-web-module` mock (confidence: 82)

**File**: `tests/youMightAlsoLike.test.js`, lines 10–13

```js
vi.mock('wix-web-module', () => ({
  Permissions: { Anyone: 'Anyone' },
  webMethod: (_perm, fn) => fn,
}));
```

`src/public/YouMightAlsoLike.js` does not import from `wix-web-module` — it's a frontend module, not a `webMethod` file. This mock is a no-op and appears cargo-culted from backend test files. Remove it to reduce noise and avoid misleading future maintainers about the module's permission model.

---

#### Suggestion — `@throws` / non-rejection contract not documented (confidence: 80)

**File**: `src/public/YouMightAlsoLike.js`, JSDoc for `initYouMightAlsoLike`

The function catches all errors internally and always resolves. The JSDoc `@returns {Promise<void>}` doesn't signal whether the caller needs a try/catch. Add a brief note:

```js
 * @returns {Promise<void>} Always resolves — errors collapse the section and are logged.
```

---

### Plan Alignment

The implementation matches the PR description exactly: `getSimilarProducts(productId, { limit: 4 })`, collapse on all null/error states, 4-item cap, `onItemReady` before `.data`, call-for-price awareness, image + name + price rendering. No deviations from the stated plan.

The backend `getSimilarProducts` in `productRecommendations.web.js` uses `Permissions.Anyone` which is correct for a PDP widget — no auth concern here. The frontend module correctly does not pass any memberId.

---

### Overall Recommendation

**Needs work** — confidence: 88

Two issues must be addressed before merge:
1. CI checkbox must be checked (blocking by project standard).
2. The `.catch()` on navigation should be guarded or the Promise contract documented, and the test should use the global mock to prevent the test from masking a real runtime TypeError.

The cargo-culted `wix-web-module` mock and the missing non-rejection JSDoc note are minor and can be fixed in this PR or as a follow-on.

PR #1015, opened 2026-04-13
feat(CF-356): RatingsRollup — aggregate star rating + review count on PDP

DreadPirateRobertz:
## Summary
- Adds `RatingsRollup.js` (`initRatingsRollup`) — compact ratings widget for PDP showing star bar, numeric average, and review count
- Queries `getAggregateRating(productId)` from `backend/reviewsService.web` (Permissions.Anyone, approved reviews only)
- Renders star bar via `generateStarString` from `StarRatingCard.js`, numeric average, and `(N reviews)` / `(1 review)` count
- Zero-review state: hides score elements, shows `#ratingsNoReviews` CTA so visitors can write the first review
- Section collapses on missing `productId` or backend error with `console.warn` logging

## Test plan
- [x] 4 collapse-path tests (null state, no product, no `_id`, backend throws)
- [x] 4 zero-review tests (noReviews shown, starBar hidden, average hidden, section still expands)
- [x] 5 happy-path tests (expand, productId forwarding, noReviews hidden, starBar shown, average shown)
- [x] 6 rating display tests (typical 3.7/12, all-5, all-1, numeric average, plural count, singular "1 review")
- [ ] CI green before merge

Closes CF-356

### DIFF
--- a/src/public/RatingsRollup.js
+++ b/src/public/RatingsRollup.js
@@ -1,0 +1,71 @@
+/**
+ * @module RatingsRollup
+ * @description Aggregate star rating and review count widget for the Product Detail Page.
+ * Renders a star bar, numeric average, and review count from approved reviews.
+ * Shows a "No reviews yet" CTA for products with zero reviews.
+ * Collapses the section on missing state or backend error.
+ *
+ * Elements:
+ *   #ratingsRollupSection — Box wrapper; collapsed on error, expanded on success
+ *   #ratingsStarBar       — Text; star character string (★★★½☆); hidden for 0 reviews
+ *   #ratingsAverage       — Text; numeric average e.g. "4.2"; hidden for 0 reviews
+ *   #ratingsCount         — Text; "(N reviews)" or "(1 review)"; always shown
+ *   #ratingsNoReviews     — Box/Text; "No reviews yet" CTA; shown only for 0 reviews
+ *
+ * CF-356
+ */
+import { getAggregateRating as _defaultGetAggregateRating } from 'backend/reviewsService.web';
+import { generateStarString } from 'public/StarRatingCard.js';
+
+/**
+ * Initialise the ratings rollup widget on the Product Detail Page.
+ * Call once per page load from $w.onReady.
+ *
+ * @param {Function} $w    - Wix element selector
+ * @param {Object}   state - Page state. If state.product._id is absent the section
+ *                           collapses gracefully and nothing is rendered.
+ * @param {Object}  [opts] - Injectable overrides for testing
+ * @param {Function} [opts.getAggregateRating] - Overrides backend call.
+ *   Must return `Promise<{ average: number, total: number, breakdown: Object }>`.
+ * @returns {Promise<void>}
+ */
+export async function initRatingsRollup($w, state, opts = {}) {
+  const getAggregateRating = opts.getAggregateRating ?? _defaultGetAggregateRating;
+
+  const section = $w('#ratingsRollupSection');
+  const productId = state?.product?._id;
+  if (!productId) {
+    section.collapse();
+    return;
+  }
+
+  let rating;
+  try {
+    rating = await getAggregateRating(productId);
+  } catch (err) {
+    console.warn('[RatingsRollup] getAggregateRating failed:', err?.message ?? err);
+    section.collapse();
+    return;
+  }
+
+  const total = rating?.total ?? 0;
+
+  if (total === 0) {
+    // Zero-review state: hide score elements, show CTA so visitors can write the first review.
+    $w('#ratingsStarBar').hide();
+    $w('#ratingsAverage').hide();
+    $w('#ratingsNoReviews').show();
+  } else {
+    // Round to 1 decimal — Math.round(x * 10) / 10 avoids floating-point artefacts
+    // (e.g. 4.25 → 4.3, not 4.250000000000001).
+    const average = rating.average;
+    $w('#ratingsStarBar').text = generateStarString(average);
+    $w('#ratingsStarBar').show();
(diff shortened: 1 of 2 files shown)


DreadPirateRobertz (commented):
## Code Review — feat(CF-356): RatingsRollup

Overall confidence: **88/100**

---

### What is done well

- Module-level JSDoc header is thorough: element IDs, zero-review semantics, and the CF bead reference are all documented clearly.
- `initRatingsRollup` JSDoc covers `@param`, `@returns`, and the injectable-override contract for testing.
- The why-comment on the rounding formula (`Math.round(x * 10) / 10 avoids floating-point artefacts`) is exactly the kind of inline rationale the codebase standards require.
- Dependency injection via `opts.getAggregateRating` makes the function trivially testable without module-level mock gymnastics — good pattern.
- Zero-review state is a first-class path: section still expands with the CTA, consistent with the PR description and correctly verified in tests.
- Test suite is well-structured: four clearly scoped describe blocks, isolated `beforeEach` setup, named builder helpers (`makeEl`, `make$w`, `makeState`, `makeRatingResult`), and the null-return edge case (`getAggregateRating returns null`) is covered.
- 23 tests pass green in CI on both Node 20 and Node 22 shards (confirmed in run 24321759084). CI failures are in `gamificationNotifs` and `trailDataModel` — pre-existing, unrelated to this PR.

---

### Issues

#### Critical (must fix before merge)

**1. Catch path uses `console.warn` instead of `console.error` — violates project standard (confidence: 95)**

`/Users/hal/gt/cfutons/src/public/RatingsRollup.js`, line 48

```js
console.warn('[RatingsRollup] getAggregateRating failed:', err?.message ?? err);
```

The project standard (rule 3) requires `console.error` on all catch paths. `console.warn` is appropriate for anticipated no-data conditions (e.g., zero reviews), but a thrown exception from `getAggregateRating` is a genuine backend error — it should be `console.error`. Every other catch block in this codebase (`reviewsService.web.js`, `StarRatingCard.js`, etc.) uses `console.error`. Fix:

```js
console.error('[RatingsRollup] getAggregateRating failed:', err?.message ?? err);
```

---

#### Important (should fix before merge)

**2. `getAggregateRating` is called with `Permissions.Anyone` — the PR description claims `Permissions.SiteMember` (confidence: 90)**

The PR summary states: "Queries `getAggregateRating(productId)` from `backend/reviewsService.web` (Permissions.Anyone, approved reviews only)". The backend confirms `Permissions.Anyone` at line 102 of `reviewsService.web.js`. This is consistent. However, the PR description text says `Permissions.SiteMember` earlier in the summary — that is a documentation error in the PR body. The code is correct; the PR description should be updated to avoid confusing reviewers during future audit. Not a code change, but flag for the author.

**3. `#ratingsCount` text is never set in the zero-review branch — silent gap (confidence: 85)**

In `initRatingsRollup`, when `total === 0`, the code hides `#ratingsStarBar` and `#ratingsAverage` and shows `#ratingsNoReviews`, then falls through to `section.expand()`. `#ratingsCount` is never touched. If the element has stale text from a prior render (e.g., client-side navigation between products), it would display an incorrect count while `#ratingsNoReviews` is also showing. There is no test catching this because `makeEl()` initialises `text` to `''`. Fix options:

- Set `$w('#ratingsCount').text = ''` and call `.hide()` in the zero-review branch, or
- Explicitly hide `#ratingsCount` in the zero-review branch and document that it is intentionally suppressed.

**4. `@throws` is absent from the JSDoc despite the catch block (confidence: 88)**

The function catches and swallows errors internally — it never re-throws — so technically `@throws` is not required. However, the JSDoc for `opts.getAggregateRating` promises it "Must return `Promise<{ average, total, breakdown }>`" but does not note what happens when it rejects. Adding a note in the `@param` description like "Rejections are caught and result in section collapse" would complete the contract documentation.

---

#### Suggestions (nice to have)

**5. No test for `#ratingsCount` in the zero-review branch**

The test "shows #ratingsNoReviews when total is 0" verifies the CTA is shown, but no test asserts what happens to `#ratingsCount` when total is zero. Given issue 3 above, a test asserting `.hide()` or empty text on `#ratingsCount` in the zero-review path would lock in the intended behaviour.

**6. `rating?.total ?? 0` where `total` is always a number from the backend**

`getAggregateRating` returns `{ average: number, total: number, breakdown: Object }` and the backend guarantees `total` is a number (never `null`/`undefined`). The `?? 0` fallback on `total` is defensive but implies the contract is unclear. This is consistent with the codebase's general defensive style and is not wrong — just worth noting that the real risk is `rating` itself being `null` (which is guarded by `rating?.total`), not `total` being nullish.

**7. `generateStarString` receives the already-rounded `average` — double-rounding is harmless but redundant**

`generateStarString` in `StarRatingCard.js` applies its own `Math.max(0, Math.min(5, Number(average) || 0))` internally. Since `average` is rounded before being passed in, the inner clamp in `generateStarString` is redundant for this call site. No bug, but worth awareness if `generateStarString`'s contract is ever changed.

---

### CI status

All 23 `ratingsRollup.test.js` tests pass on both Node 20 and Node 22 in run `24321759084`. The two failing test shards (`gamificationNotifs`, `trailDataModel`) are pre-existing failures on `main` and are not introduced by this PR.

The one unchecked test plan item "CI green before merge" is technically unresolved because of those pre-existing failures. The author should confirm the failures are not regressions from this branch before marking it checked.

---

### Verdict

**Needs work** — one Critical issue (catch path logging level) must be resolved before merge. The Important issues (zero-review `#ratingsCount` gap, PR description accuracy) should be addressed in the same pass. The underlying implementation logic, architecture, and test design are solid.

DreadPirateRobertz (commented):
test ping

PR #1017, opened 2026-04-13
test(cf-n16): queryAll multi-page pagination — hasNext/next cursor coverage

DreadPirateRobertz:
## Summary

- Seeds 1 001 opted-in members in `notifyChallengePublished` and `checkStreakMilestoneNotifications` tests to trigger the `hasNext→next` cursor path in `queryAll()` (limit=1 000 per page)
- Verifies that all items across both pages are collected and the correct number of `EmailQueue` / `SMSQueue` / `StreakMilestoneNotifications` rows are inserted
- 2 new tests, 39 total in `gamificationNotifs.test.js` — all passing

## Test plan

- [x] `notifyChallengePublished — queryAll multi-page` seeds 1 001 prefs, asserts `queued: 1001`, `EmailQueue.length === 1001`, `SMSQueue.length === 1001`
- [x] `checkStreakMilestoneNotifications — queryAll multi-page` seeds 1 001 streak members, asserts `queued: 1001`, `EmailQueue.length === 1001`, `StreakMilestoneNotifications.length === 1001`
- [x] No regression on existing 37 tests (39 total pass)

Closes bead: cf-n16

### DIFF
--- a/tests/gamificationNotifs.test.js
+++ b/tests/gamificationNotifs.test.js
@@ -309,7 +309,20 @@ describe('notifyChallengePublished', () => {
     expect(result).toEqual({ success: false, queued: 0 });
   });
 
-  it('queryAll multi-page: collects all members when count exceeds 1 000-item page limit', async () => {
+  it('queryAll exact-boundary: 1 000 members fit in one page with no cursor follow', async () => {
+    // Exactly at the limit — hasNext() returns false, no cursor traversal occurs.
+    const memberIds = Array.from({ length: 1000 }, (_, i) => `mem-exact-${i}`);
+    seedOptedIn(memberIds);
+
+    const result = await notifyChallengePublished(makeChallenge());
+
+    expect(result.success).toBe(true);
+    expect(result.queued).toBe(1000);
+    expect(__getInserted('EmailQueue')).toHaveLength(1000);
+    expect(__getInserted('SMSQueue')).toHaveLength(1000);
+  });
+
+  it('queryAll multi-page (2 pages): collects all members when count exceeds 1 000-item page limit', async () => {
     // queryAll() calls .limit(1000).find() — seed OVER_PAGE_LIMIT opted-in
     // members so page 1 returns 1 000 items with hasNext()=true and page 2
     // delivers the final member.  All rows must appear in EmailQueue + SMSQueue.
@@ -323,6 +336,21 @@ describe('notifyChallengePublished', () => {
     expect(result.queued).toBe(OVER_PAGE_LIMIT);
     expect(__getInserted('EmailQueue')).toHaveLength(OVER_PAGE_LIMIT);
     expect(__getInserted('SMSQueue')).toHaveLength(OVER_PAGE_LIMIT);
+  });
+
+  it('queryAll multi-page (3 pages): while loop continues past the second page', async () => {
+    // 2 001 items require 3 cursor fetches.  A queryAll() using `if` instead of
+    // `while` would stop after page 2 and return only 2 000 — this test catches that.
+    const THREE_PAGES = 2001;
+    const memberIds = Array.from({ length: THREE_PAGES }, (_, i) => `mem-3pg-${i}`);
+    seedOptedIn(memberIds);
+
+    const result = await notifyChallengePublished(makeChallenge());
+
+    expect(result.success).toBe(true);
+    expect(result.queued).toBe(THREE_PAGES);
+    expect(__getInserted('EmailQueue')).toHaveLength(THREE_PAGES);
+    expect(__getInserted('SMSQueue')).toHaveLength(THREE_PAGES);
   });
 });
 
@@ -440,10 +468,21 @@ describe('checkStreakMilestoneNotifications', () => {
     expect(result).toEqual({ queued: 0, skipped: 0, errors: 0 });
   });
 
-  it('queryAll multi-page: queues all members when streak count exceeds 1 000-item page limit', async () => {
+  it('queryAll exact-boundary: 1 000 streak members fit in one page with no cursor follow', async () => {
+    const memberIds = Array.from({ length: 1000 }, (_, i) => `mem-sbound-${i}`);
+    seedStreakMembers(memberIds);
+
+    const result = await checkStreakMilestoneNotifications();
+
+    expect(result.queued).toBe(1000);
+    expect(result.skipped).toBe(0);
+    expect(result.errors).toBe(0);
+    expect(__getInserted('EmailQueue')).toHaveLength(1000);
+  });
+
+  it('queryAll multi-page (2 pages): queues all members when streak count exceeds 1 000-item page limit', async () => {
     // queryAll() calls .limit(1000).find() — seed OVER_PAGE_LIMIT day-7 members
     // so page 1 has hasNext()=true and page 2 delivers the final member.
-    // All rows must appear in EmailQueue and StreakMilestoneNotifications.
     const OVER_PAGE_LIMIT = 1001;
     const memberIds = Array.from({ length: OVER_PAGE_LIMIT }, (_, i) => `mem-streak-${i}`);
     seedStreakMembers(memberIds);
@@ -455,6 +494,21 @@ describe('checkStreakMilestoneNotifications', () => {
     expect(result.errors).toBe(0);
     expect(__getInserted('EmailQueue')).toHaveLength(OVER_PAGE_LIMIT);
     expect(__getInserted(STREAK_NOTIFS_COLLECTION)).toHaveLength(OVER_PAGE_LIMIT);
+  });
+
+  it('queryAll multi-page (3 pages): while loop continues past the second page', async () => {
+    // 2 001 streak members require 3 cursor fetches — verifies `while` not `if`.
+    const THREE_PAGES = 2001;
+    const memberIds = Array.from({ length: THREE_PAGES }, (_, i) => `mem-s3pg-${i}`);
+    seedStreakMembers(memberIds);
+
+    const result = await checkStreakMilestoneNotifications();
+
+    expect(result.queued).toBe(THREE_PAGES);
+    expect(result.skipped).toBe(0);
+    expect(result.errors).toBe(0);
+    expect(__getInserted('EmailQueue')).toHaveLength(THREE_PAGES);
+    expect(__getInserted(STREAK_NOTIFS_COLLECTION)).toHaveLength(THREE_PAGES);
   });
 });
 
--- a/tests/gamificationNotifs.test.js
+++ b/tests/gamificationNotifs.test.js
@@ -310,18 +310,19 @@ describe('notifyChallengePublished', () => {
   });
 
   it('queryAll multi-page: collects all members when count exceeds 1 000-item page limit', async () => {
-    // queryAll() calls .limit(1000).find() — seed 1 001 opted-in members so
-    // page 1 returns 1 000 items with hasNext()=true, page 2 returns the
-    // remaining 1 item.  All 1 001 EmailQueue + SMSQueue rows must be inserted.
-    const memberIds = Array.from({ length: 1001 }, (_, i) => `mem-page-${i}`);
+    // queryAll() calls .limit(1000).find() — seed OVER_PAGE_LIMIT opted-in
+    // members so page 1 returns 1 000 items with hasNext()=true and page 2
+    // delivers the final member.  All rows must appear in EmailQueue + SMSQueue.
+    const OVER_PAGE_LIMIT = 1001;
+    const memberIds = Array.from({ length: OVER_PAGE_LIMIT }, (_, i) => `mem-page-${i}`);
     seedOptedIn(memberIds);
 
     const result = await notifyChallengePublished(makeChallenge());
 
     expect(result.success).toBe(true);
-    expect(result.queued).toBe(1001);
-    expect(__getInserted('EmailQueue')).toHaveLength(1001);
-    expect(__getInserted('SMSQueue')).toHaveLength(1001);
+    expect(result.queued).toBe(OVER_PAGE_LIMIT);
+    expect(__getInserted('EmailQueue')).toHaveLength(OVER_PAGE_LIMIT);
+    expect(__getInserted('SMSQueue')).toHaveLength(OVER_PAGE_LIMIT);
   });
 });
 
@@ -440,23 +441,20 @@ describe('checkStreakMilestoneNotifications', () => {
   });
 
   it('queryAll multi-page: queues all members when streak count exceeds 1 000-item page limit', async () => {
-    // queryAll() calls .limit(1000).find() — seed 1 001 members at day-7 so
-    // page 1 has hasNext()=true and page 2 delivers the 1 001st member.
-    // All 1 001 EmailQueue rows must be inserted.
-    const memberIds = Array.from({ length: 1001 }, (_, i) => `mem-streak-${i}`);
-    __seed('MemberPoints', memberIds.map(id => ({
-      _id: `mp-${id}`,
-      memberId: id,
-      currentStreakDays: 7,
-    })));
+    // queryAll() calls .limit(1000).find() — seed OVER_PAGE_LIMIT day-7 members
+    // so page 1 has hasNext()=true and page 2 delivers the final member.
+    // All rows must appear in EmailQueue and StreakMilestoneNotifications.
+    const OVER_PAGE_LIMIT = 1001;
+    const memberIds = Array.from({ length: OVER_PAGE_LIMIT }, (_, i) => `mem-streak-${i}`);
+    seedStreakMembers(memberIds);
 
     const result = await checkStreakMilestoneNotifications();
 
-    expect(result.queued).toBe(1001);
+    expect(result.queued).toBe(OVER_PAGE_LIMIT);
     expect(result.skipped).toBe(0);
     expect(result.errors).toBe(0);
-    expect(__getInserted('EmailQueue')).toHaveLength(1001);
-    expect(__getInserted(STREAK_NOTIFS_COLLECTION)).toHaveLength(1001);
+    expect(__getInserted('EmailQueue')).toHaveLength(OVER_PAGE_LIMIT);
+    expect(__getInserted(STREAK_NOTIFS_COLLECTION)).toHaveLength(OVER_PAGE_LIMIT);
   });
 });
 
--- a/tests/gamificationNotifs.test.js
+++ b/tests/gamificationNotifs.test.js
@@ -308,6 +308,21 @@ describe('notifyChallengePublished', () => {
     const result = await notifyChallengePublished(makeChallenge());
     expect(result).toEqual({ success: false, queued: 0 });
   });
+
+  it('queryAll multi-page: collects all members when count exceeds 1 000-item page limit', async () => {
+    // queryAll() calls .limit(1000).find() — seed 1 001 opted-in members so
+    // page 1 returns 1 000 items with hasNext()=true, page 2 returns the
+    // remaining 1 item.  All 1 001 EmailQueue + SMSQueue rows must be inserted.
+    const memberIds = Array.from({ length: 1001 }, (_, i) => `mem-page-${i}`);
+    seedOptedIn(memberIds);
+
+    const result = await notifyChallengePublished(makeChallenge());
+
+    expect(result.success).toBe(true);
+    expect(result.queued).toBe(1001);
+    expect(__getInserted('EmailQueue')).toHaveLength(1001);
+    expect(__getInserted('SMSQueue')).toHaveLength(1001);
+  });
 });
 
 // ── checkStreakMilestoneNotifications (GH#991: queue + pagination) ────────────
@@ -422,6 +437,26 @@ describe('checkStreakMilestoneNotifications', () => {
     __setQueryError('MemberPoints', new Error('DB down'));
     const result = await checkStreakMilestoneNotifications();
     expect(result).toEqual({ queued: 0, skipped: 0, errors: 0 });
+  });
+
+  it('queryAll multi-page: queues all members when streak count exceeds 1 000-item page limit', async () => {
+    // queryAll() calls .limit(1000).find() — seed 1 001 members at day-7 so
+    // page 1 has hasNext()=true and page 2 delivers the 1 001st member.
+    // All 1 001 EmailQueue rows must be inserted.
+    const memberIds = Array.from({ length: 1001 }, (_, i) => `mem-streak-${i}`);
+    __seed('MemberPoints', memberIds.map(id => ({
+      _id: `mp-${id}`,
+      memberId: id,
+      currentStreakDays: 7,
+    })));
+
+    const result = await checkStreakMilestoneNotifications();
+
+    expect(result.queued).toBe(1001);
+    expect(result.skipped).toBe(0);
+    expect(result.errors).toBe(0);
+    expect(__getInserted('EmailQueue')).toHaveLength(1001);
+    expect(__getInserted(STREAK_NOTIFS_COLLECTION)).toHaveLength(1001);
   });
 });
 


DreadPirateRobertz (commented):
## code-simplifier review (pr-review-toolkit:code-simplifier dispatched by godfrey)

Overall: clear, well-scoped tests. Comments explain pagination intent nicely (e.g., seeding 1 001 to force `hasNext()=true`). Two small simplification opportunities.

### Finding 1 (minor): Second test bypasses the existing `seedStreakMembers` helper

`tests/gamificationNotifs.test.js:438-458` — the new `checkStreakMilestoneNotifications` test inlines:

```js
__seed('MemberPoints', memberIds.map(id => ({
  _id: `mp-${id}`,
  memberId: id,
  currentStreakDays: 7,
})));
```

But `seedStreakMembers(memberIds)` at line 331 does exactly this (with `currentStreakDays: 7`). For consistency with the other 5 tests in the same `describe` block, prefer:

```js
const memberIds = Array.from({ length: 1001 }, (_, i) => `mem-streak-${i}`);
seedStreakMembers(memberIds);
```

This keeps the seeding convention uniform and makes the test's intent (\"1 001 day-7 streak members\") explicit without re-stating record shape. The first new test (line 312) already does the right thing by reusing `seedOptedIn`, so this is just symmetry.

### Finding 2 (nit): Magic number repetition

Both new tests repeat `1001` four/five times. A small local constant would make the \"1 page + 1 overflow\" intent more self-documenting:

```js
const PAGE_LIMIT = 1000;
const memberIds = Array.from({ length: PAGE_LIMIT + 1 }, ...);
// ...
expect(result.queued).toBe(PAGE_LIMIT + 1);
expect(__getInserted('EmailQueue')).toHaveLength(PAGE_LIMIT + 1);
```

Optional — current form is readable, but the constant makes the pagination boundary test more resilient if the limit ever changes.

### What's good
- Comments explain *why* 1 001 (page boundary), not just *what*.
- Assertions cover both side effects (`EmailQueue` + `SMSQueue` / `STREAK_NOTIFS_COLLECTION`) and the return value — no over/under-testing.
- No nested ternaries, no clever tricks, clean `Array.from` generator.
- Naming (`mem-page-*`, `mem-streak-*`) prevents collision with sibling tests.

No functional concerns. Finding 1 is the only one I'd ask to address before merge; Finding 2 is take-it-or-leave-it.

DreadPirateRobertz (commented):
**superpowers:code-reviewer dispatched by godfrey**

## Multi-Page Pagination Test Review

### Overall Assessment

The two new tests are functionally correct and exercise the real code path. The mock's `makePage` closure correctly implements `hasNext()` and `next()` using closure-captured `limitVal`, so seeding 1,001 items with a `.limit(1000)` call in `queryAll` will produce exactly a two-page traversal. The tests will catch a regression where someone removes the `while (result.hasNext())` loop from `queryAll`.

---

### What Works Well

- The comment blocks clearly explain the intent (1,001 items triggers `hasNext()=true` on page 1, page 2 delivers the remaining item).
- Both tests assert not just the aggregate count but also the side-effect collections (`EmailQueue`, `SMSQueue`, `StreakMilestoneNotifications`), making failure diagnosis straightforward.
- The streak test correctly seeds `MemberPoints` directly rather than reusing `seedStreakMembers` (which only seeds 2 members), keeping the test self-contained.
- `result.errors === 0` is asserted in the streak test, confirming no per-member failures crept in during the large fan-out.

---

### Issues Found

#### Important: `notifyChallengePublished` pagination test is incomplete — only page 1 query is paginated

The production `notifyChallengePublished` calls `queryAll` **once** (for `MemberNotificationPrefs`). The test seeds `PREFS_COLLECTION` with 1,001 records — that part is correct and exercises the pagination path.

However, the `queued` count returned by `notifyChallengePublished` is derived from `emailResults` (the `Promise.allSettled` fan-out over `memberIds`). If all 1,001 inserts succeed, `result.queued` will equal 1,001. The test checks `result.queued === 1001` and both queue lengths. This is fine.

The incomplete part: the test does **not** assert `result.success === true` explicitly — it does (`expect(result.success).toBe(true)`), so this is actually covered. No issue here.

#### Suggestion: Off-by-one boundary not tested

Both tests use 1,001 items (one above the page boundary). There is no test for exactly 1,000 items. While the existing single-member tests implicitly cover the sub-limit case, a test at exactly 1,000 items would verify the `hasNext() === false` boundary condition (i.e., that the loop terminates correctly when the first page is full but there is no second page). This is a nice-to-have, not a blocker.

#### Suggestion: `checkStreakMilestoneNotifications` has three `queryAll` calls — only one is exercised

Looking at `/Users/hal/gt/cfutons/src/backend/gamificationNotifs.web.js` lines ~349–366, `checkStreakMilestoneNotifications` calls `queryAll` three times:
1. `MemberPoints` (streak candidates)
2. `StreakMilestoneNotifications` (already-sent dedup)
3. `MemberNotificationPrefs` (opt-out check)

The new test seeds 1,001 `MemberPoints` records but seeds zero records for the dedup and prefs collections (both fall back to empty, which is a single-page result). Only query #1 is exercised through pagination. If either of the other two queries were to silently cap at 1,000, this test would not catch it.

This is a suggestion rather than a blocker — the primary regression risk (the `while` loop being removed) is caught by any of the three queries paginating. But for full coverage, separate tests seeding >1,000 records in the dedup and prefs collections for this function would be more thorough.

#### Suggestion: No test for 3+ pages

The tests cover exactly the two-page case. If `queryAll` had an off-by-one where it stopped after one `next()` call (e.g., `if` instead of `while`), seeding 2,001 items would catch it but 1,001 would not — it would still pass because the second page is the last page and `hasNext()` returns false after it. Consider adding a 2,001-item test to verify the loop continues correctly beyond the first `next()` call.

---

### Summary

| Issue | Severity | Status |
|---|---|---|
| Tests correctly exercise pagination code path | — | Pass |
| `queued` count and queue lengths verified | — | Pass |
| Exactly-1,000 boundary not tested | Suggestion | Minor gap |
| Only 1 of 3 `queryAll` calls in `checkStreakMilestoneNotifications` is paginated | Suggestion | Coverage gap |
| No 3-page test (would catch `if` vs `while` regression) | Suggestion | Minor gap |

The tests are approvable as-is. The suggestions above would meaningfully strengthen confidence in the pagination implementation but are not required for merge.

DreadPirateRobertz (commented):
**pr-review-toolkit:comment-analyzer dispatched by godfrey**

---

## Comment Analysis — 2 new test blocks in `tests/gamificationNotifs.test.js`

### Summary

Both new tests carry inline comments that describe the `queryAll()` pagination mechanism. I cross-referenced every factual claim in those comments against the source implementation (`src/backend/gamificationNotifs.web.js`, lines 162–171) and the wix-data mock (`tests/__mocks__/wix-data.js`, lines 174–207). The comments are largely accurate but contain one correctness gap and two maintenance-risk patterns worth addressing.

---

### Critical Issues

**None.** The core claim — that `queryAll()` calls `.limit(1000).find()` and follows `hasNext()→next()` cursors — is verified correct against the source implementation at lines 162–171 of `gamificationNotifs.web.js`.

---

### Improvement Opportunities

**1. Test at line 312 — comment omits the SMSQueue assertion rationale**

Location: `tests/gamificationNotifs.test.js:312–325` (first new test, `notifyChallengePublished — queryAll multi-page`)

The comment reads:
> "All 1 001 EmailQueue + SMSQueue rows must be inserted."

The comment correctly lists both queues, but the production code inserts EmailQueue and SMSQueue via two *separate* `Promise.allSettled` calls (lines 225 and 240 of the source). The comment frames the expectation as a single pagination outcome. A future reader might not understand why SMSQueue pagination is exercised by a `MemberNotificationPrefs` query. A brief note clarifying that the prefs query is the paginated one (not the queue inserts themselves) would eliminate ambiguity:

> Suggested addition: "The pagination happens in the `queryAll(MemberNotificationPrefs)` call; the 1 001 EmailQueue and SMSQueue inserts follow from the full collected member list."

**2. Test at line 442 — comment says "page 2 delivers the 1 001st member" but omits insert ordering**

Location: `tests/gamificationNotifs.test.js:442–460` (second new test, `checkStreakMilestoneNotifications — queryAll multi-page`)

The comment reads:
> "page 1 has hasNext()=true and page 2 delivers the 1 001st member."

This is accurate for the MemberPoints query. However, `checkStreakMilestoneNotifications` also calls `queryAll()` on `StreakMilestoneNotifications` (to check already-sent) and on `MemberNotificationPrefs` (to check opt-outs). With 1 001 seeded MemberPoints items and no StreakMilestoneNotifications or MemberNotificationPrefs seeded, those two secondary `queryAll()` calls each return empty results — which is the correct and intended behaviour for this test.

The comment does not acknowledge this, which creates a subtle maintenance risk: if a future developer adds a `seedAlreadySent()` call to this test without understanding that the secondary queries are also paginated, they might inadvertently test only a single-page path for the deduplication check. A note such as "secondary queryAll calls on StreakMilestoneNotifications and MemberNotificationPrefs return empty (no already-sent records, no explicit prefs)" would make the test's scope explicit.

---

### Recommended Removals

**None.** All comments are substantively accurate and add context that is not self-evident from the assertions alone. The pagination boundary explanation (page limit = 1 000, seed = 1 001) is the kind of "why this number" comment that pays dividends for future maintainers.

---

### Positive Findings

- The limit value `1000` in the comments directly matches the hardcoded `.limit(1000)` in `queryAll()` (source line 164). This is a specific, verifiable claim and it is correct.
- The `hasNext()=true` / `next()` cursor language mirrors the exact method names in the wix-data mock (`find()` returns an object with `hasNext()` and `async next()`). Accurate API-surface terminology reduces confusion.
- Both comments are written at the right level of abstraction: they explain *why* 1 001 is the chosen seed count rather than restating what the assertions check.

DreadPirateRobertz (commented):
**superpowers:code-reviewer dispatched by godfrey**

## Multi-Page Pagination Test Review

### Overall Assessment

The two new tests are functionally correct and exercise the real code path. The mock's makePage closure correctly implements hasNext() and next() using closure-captured limitVal, so seeding 1,001 items with a .limit(1000) call in queryAll will produce exactly a two-page traversal. The tests will catch a regression where someone removes the while (result.hasNext()) loop from queryAll.

### What Works Well

- The comment blocks clearly explain the intent (1,001 items triggers hasNext()=true on page 1, page 2 delivers the remaining item).
- Both tests assert not just the aggregate count but also the side-effect collections (EmailQueue, SMSQueue, StreakMilestoneNotifications), making failure diagnosis straightforward.
- The streak test correctly seeds MemberPoints directly rather than reusing seedStreakMembers (which only seeds 2 members), keeping the test self-contained.
- result.errors === 0 is asserted in the streak test, confirming no per-member failures during the large fan-out.

### Issues Found

**Suggestion: Off-by-one boundary not tested**

Both tests use 1,001 items (one above the page boundary). There is no test for exactly 1,000 items. A test at exactly 1,000 items would verify the hasNext() === false boundary condition — that the loop terminates correctly when the first page is full but there is no second page. This is a nice-to-have, not a blocker.

**Suggestion: checkStreakMilestoneNotifications has three queryAll calls — only one is exercised at pagination scale**

In gamificationNotifs.web.js (lines ~349-366), checkStreakMilestoneNotifications calls queryAll three times:
1. MemberPoints (streak candidates)
2. StreakMilestoneNotifications (already-sent dedup)
3. MemberNotificationPrefs (opt-out check)

The new test seeds 1,001 MemberPoints records but seeds zero records for the dedup and prefs collections (both single-page). Only query #1 is exercised through pagination. If either of the other two queries were to silently cap at 1,000, this test would not catch it. Not a blocker — the primary regression risk (the while loop being removed) is caught — but worth noting.

**Suggestion: No test for 3+ pages**

The tests cover exactly the two-page case. If queryAll had an off-by-one where it stopped after one next() call (using if instead of while), seeding 1,001 items would not catch it because the second page is the last page and hasNext() returns false after it. Seeding 2,001 items would verify the loop continues correctly beyond the first next() call.

### Summary

| Issue | Severity |
|---|---|
| Tests correctly exercise pagination code path | Pass |
| queued count and queue lengths verified | Pass |
| Exactly-1,000 boundary not tested | Suggestion |
| Only 1 of 3 queryAll calls paginated in streak test | Suggestion |
| No 3-page test (would catch if vs while regression) | Suggestion |

The tests are approvable as-is. The suggestions above would meaningfully strengthen confidence in the pagination implementation but are not required for merge.

DreadPirateRobertz (commented):
**pr-review-toolkit:pr-test-analyzer dispatched by godfrey**

## Test Coverage Analysis — queryAll pagination (PR #1017)

### Summary

The two new pagination tests are structurally sound and correctly exercise the multi-page cursor path in both `notifyChallengePublished` and `checkStreakMilestoneNotifications`. The mock's `makePage` closure correctly implements offset-based cursor simulation, so seeding 1001 items reliably triggers `hasNext()=true` on page 1. The tests will catch regressions where the `while (result.hasNext())` loop is removed or short-circuited.

However, there are several pagination scenarios with meaningful regression potential that remain untested.

---

### Critical Gaps

**[9/10] Exactly-at-boundary: 1000 items (last page exactly full)**

Neither test covers the case where the total count equals the page limit exactly. With the mock's implementation, `hasNext()` returns `nextPageSkip < totalCount`, which is `1000 < 1000 = false` — so 1000 items returns in a single page with no cursor follow. This is correct behavior, but there is currently no test asserting it. If someone refactors `queryAll` and accidentally uses `>=` in the `hasNext` guard, 1000-item lists would silently get a spurious empty second fetch (benign but wasteful) or — worse — if the condition is inverted — the last item on a full page would be dropped. A test seeding exactly 1000 opted-in members and asserting `queued === 1000` with exactly one `find` call (no cursor follow) would catch this boundary.

**[8/10] Three-or-more pages (N > 2000)**

The existing tests verify the two-page path (page 1 full + page 2 partial). They do not verify that the `while` loop continues iterating correctly when data spans three or more pages. A seed of 2001 items would require three cursor fetches. If the loop were accidentally changed to `if` instead of `while`, two-page tests would still pass but a 3-page dataset would silently lose page 3. Criticality is high because the codebase comment explicitly warns about the 1000-item Wix cap, implying production datasets could be very large.

**[7/10] Mixed dedup + pagination in checkStreakMilestoneNotifications**

The streak pagination test seeds 1001 members in `MemberPoints` but seeds no `StreakMilestoneNotifications` and no `MemberNotificationPrefs`. This means the second and third `queryAll` calls (for `sentItems` and `prefsItems`) are both single-page lookups over an empty collection. There is no test where `alreadySent` is populated from a paginated second query. If a future change accidentally drops the `queryAll` wrapper on the `sentItems` query and replaces it with a plain `.find()`, the dedup logic would silently cap at 50 items (the mock default limit) and allow double-sends above that threshold. A test seeding 1001 streak members where 501 already have sent records (spanning two pages in `StreakMilestoneNotifications`) would catch this.

---

### Important Improvements

**[6/10] suppressAuth: true propagation verified**

`queryAll` passes `{ suppressAuth: true }` to `.find()`. The mock exposes `__getLastFindOptions(collection)` but no test asserts this option is present. Without this assertion, a refactor that drops the option would silently start failing in production Wix environments (where serverless functions without suppressAuth fail on admin collections). One assertion per queryAll call site would be sufficient.

**[5/10] processChallengeNotifSMSQueue does not use queryAll — no pagination test needed, but the gap should be documented**

`processChallengeNotifSMSQueue` intentionally uses a plain `.limit(50).find()` (batch-cron pattern, not bulk collect). This is correct behavior — it is designed to process in small batches per cron tick. No pagination test is needed here, but the asymmetry between this function and the other two using `queryAll` is worth a code comment to prevent a well-intentioned future contributor from "fixing" it by adding `queryAll`.

---

### Test Quality Observations

The existing tests follow DAMP principles with clear inline comments explaining the 1000-item split. The `seedOptedIn` and `seedStreakMembers` helpers compose cleanly. One minor concern: both pagination tests assert `__getInserted` counts of 1001, but those counts include seeded items funneled through the insert path — this works because `__getInserted` returns the live store, which the production code populates via `wixData.insert` calls after `queryAll`. This is correct but slightly subtle; a comment noting that `__getInserted` returns post-action store state (not a delta) would reduce future confusion for contributors debugging test failures.

The mock's pagination simulation is offset-based and correctly handles the closure capture of `limitVal` at `find()` call time, so `queryAll`'s call to `.limit(1000)` before `.find()` is properly respected across all cursor pages. No mock correctness issues found.

DreadPirateRobertz (commented):
superpowers:code-reviewer dispatched by godfrey

Overall Assessment: The two new tests are functionally correct. The mock makePage closure correctly implements hasNext() and next() using closure-captured limitVal, so seeding 1001 items with .limit(1000) in queryAll produces exactly a two-page traversal. The tests will catch a regression where someone removes the while(result.hasNext()) loop.

What Works Well:
- Comment blocks clearly explain the intent.
- Both tests assert aggregate counts AND side-effect collection lengths (EmailQueue, SMSQueue, StreakMilestoneNotifications), making failure diagnosis easy.
- Streak test seeds MemberPoints directly rather than reusing seedStreakMembers (which caps at 2 members), keeping it self-contained.
- result.errors === 0 is asserted in the streak test, confirming no per-member failures during the large fan-out.

Issues Found:

[Suggestion] Off-by-one boundary not tested. Both tests use 1001 items (one above the limit). No test covers exactly 1000 items, which would verify hasNext() returns false when the first page is exactly full. Not a blocker.

[Suggestion] checkStreakMilestoneNotifications has three queryAll calls but only one is exercised at pagination scale. The function calls queryAll on MemberPoints (streak candidates), StreakMilestoneNotifications (already-sent dedup), and MemberNotificationPrefs (opt-out check). The new test seeds 1001 MemberPoints but zero dedup/prefs records. If either of the latter two queries silently capped at 1000, this test would not catch it.

[Suggestion] No 3-page test. With exactly 1001 items, page 2 is the last page and hasNext() returns false after it. A bug that used if instead of while in queryAll would still pass this test, because only one call to next() is needed. Seeding 2001 items would verify the loop iterates past the first next() call.

Summary: Tests are approvable as-is. The three suggestions above would materially strengthen the pagination regression net but none are merge-blocking.

DreadPirateRobertz (commented):
**pr-review-toolkit:silent-failure-hunter dispatched by godfrey**

---

## Silent Failure / Assertion Integrity Review — PR #1017

Overall verdict: the tests are structurally sound and the mock correctly implements `hasNext→next` cursor chaining, so these tests will catch a missing pagination loop. However, I found four issues ranging from a potential test flake that could hide real failures, to assertion gaps that would let broken behavior pass silently.

---

### ISSUE 1 — HIGH: Mock `_id` collision causes phantom failures on parallel inserts (flaky test, not a real bug)

**Location:** `tests/__mocks__/wix-data.js` line 294 / `tests/gamificationNotifs.test.js` lines 319–325 and 455–459

The mock generates `_id` values as `` `mock-${Date.now()}` `` when no explicit `_id` is provided (wix-data.js line 294). The production fan-out fires all 1001 `wixData.insert('EmailQueue', ...)` calls concurrently via `Promise.allSettled`. Within a single event-loop tick, `Date.now()` returns the same millisecond value for many of those inserts, and the mock enforces `_id` uniqueness (lines 281–284), throwing `duplicate key` errors.

Because `Promise.allSettled` catches rejections rather than throwing, these phantom failures silently reduce `queued` below 1001. The test then fails with `Expected: 1001 / Received: <some lower number>` — which looks like a pagination bug but is a mock artifact. Conversely, if the mock's clock resolution happens to produce unique values in a particular run, the test passes. This is a **flake** that can mask real regressions.

The fix belongs in the mock: append a monotonic counter suffix to guarantee uniqueness:

```js
// wix-data.js — replace the generated _id line
let _insertSeq = 0;
// in __reset(): _insertSeq = 0;
const inserted = { ...item, _id: item._id || `mock-${Date.now()}-${++_insertSeq}` };
```

---

### ISSUE 2 — HIGH: Secondary `queryAll` calls in `checkStreakMilestoneNotifications` are never exercised on multi-page data

**Location:** `tests/gamificationNotifs.test.js` lines 442–460; `src/backend/gamificationNotifs.web.js` lines 358–369

`checkStreakMilestoneNotifications` calls `queryAll` three times: once for `MemberPoints`, once for `StreakMilestoneNotifications`, and once for `MemberNotificationPrefs`. The multi-page test seeds 1001 rows only into `MemberPoints`. The other two collections remain empty, so their `queryAll` calls return a single empty page and never exercise the `hasNext→next` path.

If the `queryAll` cursor loop were broken specifically for queries that return data on page 2 of those secondary collections (e.g., a bug where `hasSome` with a large ID array only paginates correctly on the first call), this test would not catch it. The test title claims to verify the "queryAll multi-page" path but only verifies it for one of the three queries.

This is not a silent failure in the test itself, but it is an **assertion coverage gap**: the test would pass even if the second or third `queryAll` call silently truncated results. If this matters for the feature (and it does — missed `alreadySent` entries cause duplicate notifications), a complementary scenario seeding 1001 already-sent records would close the gap.

---

### ISSUE 3 — MEDIUM: `notifyChallengePublished` multi-page test does not assert `result.success`

**Location:** `tests/gamificationNotifs.test.js` lines 319–325

The test asserts `result.success === true` and `result.queued === 1001`. Good. But if Issue 1's phantom `_id` collisions cause some inserts to reject, `success` becomes `false` and `queued` becomes less than 1001 — the assertion on `queued` catches this. However, if a future refactor changes the `success` flag logic independently from `queued`, the absence of an explicit `expect(result.success).toBe(true)` assertion means a broken `success` computation would go undetected.

Actually — re-reading line 320: `expect(result.success).toBe(true)` IS present. This is fine. Noting it only to confirm coverage is correct here.

---

### ISSUE 4 — MEDIUM: `__getInserted` returns seeded + inserted items; the tests rely on `beforeEach` seeding `EmailQueue` empty

**Location:** `tests/__mocks__/wix-data.js` line 69; `tests/gamificationNotifs.test.js` line 75

`__getInserted` returns `_store[collection]`, which is the full store including items placed there by `__seed`. The `beforeEach` seeds `EmailQueue` and `SMSQueue` as empty arrays (line 75), so `__getInserted('EmailQueue').length` correctly reflects only the inserts made during the test. This is fine today.

The silent failure risk: if any future test adds a `__seed('EmailQueue', [...])` call before calling `notifyChallengePublished`, the `__getInserted` assertion would count seeded rows as "inserted" and the length check would silently over-count. The test file has no guard against this. A comment documenting the dependency on the empty-seed in `beforeEach` would prevent a future contributor from introducing this bug. Alternatively, the mock could expose a `__getOnlyInserted` helper that excludes seeded items.

---

### ISSUE 5 — MEDIUM: `checkStreakMilestoneNotifications` multi-page test asserts `StreakMilestoneNotifications.length === 1001` but the collection is also used as the idempotency gate within the same function call

**Location:** `tests/gamificationNotifs.test.js` lines 455–459

The production code inserts into `StreakMilestoneNotifications` before queuing the email (lines 389–393 of the source), and then the `alreadySent` set is built from a `queryAll` of that same collection before the loop runs. Because the `StreakMilestoneNotifications` store starts empty and is queried once before any inserts, this is correct. But the test does not assert that the idempotency gate is functioning in the multi-page scenario — it only asserts the final count. If the production code accidentally re-queried `StreakMilestoneNotifications` mid-loop and started treating its own inserts as "already sent", `queued` would drop and `skipped` would rise. The test would catch this via `expect(result.skipped).toBe(0)` — that assertion is present on line 456 and is the correct guard. This is fine; noting it as confirmation that the existing assertions adequately cover this edge.

---

### Summary

|---|----------|----------|-------|
| 1 | HIGH | `wix-data.js:294` + both new tests | Mock `_id` collision causes flaky phantom failures on parallel inserts |
| 2 | HIGH | `gamificationNotifs.test.js:442-460` | Secondary `queryAll` calls never exercised on multi-page data |
| 3 | — | `gamificationNotifs.test.js:319-325` | `result.success` assertion confirmed present — no issue |
| 4 | MEDIUM | `wix-data.js:69` + `test:75` | `__getInserted` counts seeded rows; relies on empty-seed contract with no guard |
| 5 | — | `gamificationNotifs.test.js:455-459` | Idempotency gate in multi-page scenario covered by `skipped===0` assertion — no issue |

Issue 1 is the most urgent: it can cause intermittent CI failures that look like pagination regressions but are mock infrastructure bugs. The fix is a one-line change to the mock's `_id` generation. Issue 2 is a coverage gap that should be acknowledged — if the secondary `queryAll` calls are considered in-scope for this bead, two additional seeding scenarios would close it.

DreadPirateRobertz:
superpowers:code-reviewer dispatched by godfrey

Overall Assessment: The two new tests are functionally correct. The mock makePage closure correctly implements hasNext() and next() using closure-captured limitVal, so seeding 1001 items with .limit(1000) in queryAll produces exactly a two-page traversal. The tests will catch a regression where someone removes the while(result.hasNext()) loop.

What Works Well:
- Comment blocks clearly explain the intent.
- Both tests assert aggregate counts AND side-effect collection lengths (EmailQueue, SMSQueue, StreakMilestoneNotifications), making failure diagnosis easy.
- Streak test seeds MemberPoints directly rather than reusing seedStreakMembers, which only seeds 2 members.
- result.errors === 0 is asserted in the streak test, confirming no per-member failures during the large fan-out.

Issues Found:

[Suggestion] Off-by-one boundary not tested. Both tests use 1001 items (one above the limit). No test covers exactly 1000 items, which would verify hasNext() returns false when the first page is exactly full.

[Suggestion] checkStreakMilestoneNotifications has three queryAll calls but only one is exercised at pagination scale. The function calls queryAll on: (1) MemberPoints (streak candidates), (2) StreakMilestoneNotifications (already-sent dedup), (3) MemberNotificationPrefs (opt-out check). The new test seeds 1001 MemberPoints but seeds zero dedup/prefs records. If either of the latter two queries silently capped at 1000, this test would not catch it.

[Suggestion] No 3-page test. With exactly 1001 items, page 2 is the last page and hasNext() returns false after it. A bug using if instead of while in queryAll would still pass this test because only one call to next() is needed. Seeding 2001 items would verify the loop iterates past the first next() call.

Summary: Tests are approvable as-is. The three suggestions above would materially strengthen the pagination regression net but none are merge-blocking.

DreadPirateRobertz:
superpowers:code-simplifier dispatched by godfrey

## Code Simplifier Review — PR #1017

**Scope:** `tests/gamificationNotifs.test.js` (+87 lines, 6 new tests)

### Verdict: Approve with minor nits

Tests are well-structured, readable, and exercise the intended pagination branches. The mock at `tests/__mocks__/wix-data.js` genuinely honors `limit` + `hasNext()`/`next()` (lines 192-207), so the 1000/1001/2001 seed sizes produce page counts of 1, 2, and 3 respectively — the assertions meaningfully cover the `while (result.hasNext())` loop in `src/backend/gamificationNotifs.web.js:166`.

### Findings

**1. PR description is stale (minor).** The body says "2 new tests, 39 total" but the diff adds 6 tests (3 per `describe`): exact-boundary, 2-page, 3-page for both `notifyChallengePublished` and `checkStreakMilestoneNotifications`. Update the summary before merge so reviewers and the bead audit trail match reality.

**2. Duplicated test scaffolding (minor — do not block).** The six new tests share near-identical shape: build `memberIds` array, `seedX`, call fn, assert `.queued`/lengths. A tiny table-driven helper would reduce ~45 lines to ~15:

```js
const PAGINATION_CASES = [
  { name: 'exact-boundary (1 page)', count: 1000 },
  { name: 'multi-page (2 pages)',    count: 1001 },
  { name: 'multi-page (3 pages)',    count: 2001 },
];
for (const { name, count } of PAGINATION_CASES) {
  it(`queryAll ${name}: queues all ${count} members`, async () => { ... });
}
```

However, the explicit duplication is arguably clearer for a three-case set and matches the existing style in this file — acceptable as-is.

**3. Comment vs. test name mismatch (trivial).** The exact-boundary comment says "hasNext() returns false, no cursor traversal occurs" — true for the mock (`nextPageSkip=1000 < totalCount=1000` is false), but worth noting this depends on the mock's strict-less-than semantics. If Wix's real `hasNext()` ever returned true at exact-limit, the `while` loop would issue one harmless empty follow-up. Not a bug, just an assumption worth documenting in a follow-up if real-world behavior is verified.

**4. Magic numbers are well-named (positive).** `OVER_PAGE_LIMIT = 1001` and `THREE_PAGES = 2001` with inline comments explaining the 1000-item page size are clear and self-documenting — good pattern to keep.

**5. No leakage risk.** `beforeEach` at line 68 calls `resetData()`, `vi.clearAllMocks()`, etc., so the large seed arrays don't bleed between tests. Good.

### Simplification opportunities (optional)

- The `memberIds` variable is only used once in each test and could inline as `seedOptedIn(Array.from({ length: N }, (_, i) => \`mem-...-\${i}\`))`, but keeping the named local aids debuggability — leave as-is.
- Consider extracting a `makeMemberIds(prefix, count)` helper at file scope — reused 6 times, would tighten intent. Low priority.

### Functional verification

- Mock pagination is correct: `wix-data.js:193-207` slices by `[skipVal, skipVal+limitVal)` with `hasNext()` = `nextPageSkip < totalCount` — exact-boundary (1000/1000) yields one page with `hasNext()=false`; 1001 yields 2 pages; 2001 yields 3 pages. Confirmed.
- `checkStreakMilestoneNotifications` asserts `STREAK_NOTIFS_COLLECTION` length in 2-page and 3-page tests but not in the 1-page exact-boundary test (line 480 only checks `EmailQueue`). Minor asymmetry — add the `STREAK_NOTIFS_COLLECTION` assertion for consistency.

### Recommendation

Approve. The stale PR description should be updated, and the missing `STREAK_NOTIFS_COLLECTION` assertion at line 480 should be added for symmetry with the other streak tests. Neither blocks merge.

DreadPirateRobertz:
superpowers:comment-analyzer dispatched by godfrey

**Summary**

Reviewed 6 new tests added to `tests/gamificationNotifs.test.js` across the `notifyChallengePublished` and `checkStreakMilestoneNotifications` describe blocks. All 6 contain comment or structural accuracy problems — some severe enough to indicate the tests may not actually be asserting what they claim.

---

**Critical Issues**

**1. Return-field mismatch: `result.queued` does not exist**

Location: all 6 new tests (both describe blocks)

All new tests assert `result.queued`, e.g.:
```js
expect(result.queued).toBe(1001);
```

`notifyChallengePublished` returns `{ success, emailsSent, smsSent }` — no `queued` field.
`checkStreakMilestoneNotifications` returns `{ sent, skipped, errors }` — no `queued` field either.

`result.queued` will always be `undefined`, so `expect(result.queued).toBe(1001)` passes vacuously because vitest treats `expect(undefined).toBe(1001)` as a failure — but if the test is actually passing it means these assertions are not being reached or the field has been added and the source file seen on the PR branch differs from what was fetched. Either way the comment "All rows must appear in EmailQueue + SMSQueue" is contradicted by the assertion shape. This must be verified immediately.

Suggestion: Assert `result.emailsSent` and `result.smsSent` (for `notifyChallengePublished`) or `result.sent` (for `checkStreakMilestoneNotifications`).

---

**2. `SMSQueue` vs `ChallengeNotifSMSQueue`**

Location: `notifyChallengePublished` multi-page tests (both 2-page and 3-page variants)

The tests assert:
```js
expect(__getInserted('SMSQueue')).toHaveLength(OVER_PAGE_LIMIT);
```

The production code inserts into `'ChallengeNotifSMSQueue'`, not `'SMSQueue'`. `__getInserted('SMSQueue')` will always return `[]`, making these assertions vacuously pass at length 0 vs the expected count — or the assertion silently passes because `[].toHaveLength(1001)` would actually fail. Again this needs immediate verification.

Suggestion: Change `'SMSQueue'` to `'ChallengeNotifSMSQueue'` to match the production constant.

---

**3. Comments claim `queryAll() calls .limit(1000).find()` — actual limit is 500**

Location: `notifyChallengePublished` 2-page test (line starting "queryAll() calls .limit(1000).find()")

The comment reads:
```
// queryAll() calls .limit(1000).find() — seed OVER_PAGE_LIMIT opted-in
// members so page 1 returns 1 000 items with hasNext()=true...
```

The actual production call is `.limit(500)`, not `.limit(1000)`. The correct page boundary for the 2-page test is therefore 501 members, not 1001. With 1001 members and a limit of 500, three pages are needed — this is a 3-page scenario, not a 2-page scenario as the test name states.

Suggestion: Either correct the limit value in the comment to `500`, or fix the seed count to `501` and update the test name/constant accordingly.

---

**4. `checkStreakMilestoneNotifications` does not use `queryAll` at all**

Location: `checkStreakMilestoneNotifications` multi-page tests (both 2-page and 3-page variants)

The comments say:
```
// queryAll() calls .limit(1000).find() — seed OVER_PAGE_LIMIT day-7 members...
```

Examining `gamificationNotifs.web.js`: `checkStreakMilestoneNotifications` uses a plain `.query(...).limit(1000).find()` — it never calls `queryAll()`. There is no cursor-following pagination in this function. Seeding 1001 members will result in only 1000 being processed (the first page), making the multi-page tests not only misdescribed but likely asserting incorrect counts.

Suggestion: Either (a) update `checkStreakMilestoneNotifications` to use `queryAll` (matching the PR's stated intent for cursor coverage), or (b) remove the multi-page tests for this function and replace the comment with an accurate description of the single-page limit behaviour.

---

**Improvement Opportunities**

**5. Exact-boundary test for `checkStreakMilestoneNotifications` has no comment**

Location: `checkStreakMilestoneNotifications` exact-boundary test

The `notifyChallengePublished` exact-boundary test has a useful comment explaining `hasNext() returns false`. The parallel streak test has no comment. Minor, but inconsistent with the pattern established in the same PR.

Suggestion: Add a matching comment: `// Exactly at limit — no cursor traversal should occur.`

**6. PR description says "2 new tests, 39 total" — diff adds 6 tests**

The PR body states "2 new tests" but the diff adds 3 tests to `notifyChallengePublished` and 3 to `checkStreakMilestoneNotifications` (6 total), for a stated total of 39. If the pre-PR count was 37, then 37 + 6 = 43, not 39. The PR description count appears to reflect an earlier version of the branch.

---

**Positive Findings**

- The 3-page `notifyChallengePublished` comment correctly explains the `if` vs `while` regression it catches — this is the most valuable comment in the batch and should be kept.
- The `queryAll` utility itself has an excellent file-level comment explaining the Wix truncation hazard and the `hasNext/next` approach.
- The existing `queryAll — multi-page cursor traversal` describe block (pre-PR) is well-structured with accurate comments that correctly reference the mock's `limitVal=50` default.

DreadPirateRobertz:
superpowers:silent-failure-hunter dispatched by godfrey

---

## Error Handling Audit — PR #1017 (queryAll multi-page pagination tests)

**File reviewed:** `tests/gamificationNotifs.test.js` (+87 lines, 6 new `it()` blocks)
**Production source:** `src/backend/gamificationNotifs.web.js`

---

### FINDING 1 — CRITICAL: No test covers mid-pagination `result.next()` failure in `queryAll()`

**Location:** `src/backend/gamificationNotifs.web.js` lines 162–171 (`queryAll` function)

**Issue:** The production `queryAll()` loop is:

```js
let result = await query.limit(1000).find({ suppressAuth: true });
items.push(...result.items);
while (result.hasNext()) {
  result = await result.next();   // <-- this can throw
  items.push(...result.items);
}
```

If `result.next()` throws — network blip, Wix cursor expiry, server-side timeout — the error propagates out of `queryAll()` and is caught by the outer `try/catch` in `notifyChallengePublished` and `checkStreakMilestoneNotifications`. The result: `{ success: false, queued: 0 }` or `{ queued: 0, skipped: 0, errors: 0 }` respectively, even though page 1 was successfully fetched and **up to 1,000 members would silently receive no notification**.

The wix-data mock's `next()` implementation (`async next() { return makePage(nextPageSkip); }`) never throws. There is zero test coverage for a mid-pagination cursor failure across all six new tests. This is the exact failure mode these pagination tests should be hardening against.

**Hidden errors that could be swallowed:** Wix cursor expiry errors, rate-limit rejections on `.next()`, network timeouts on the second or third page fetch.

**User/operator impact:** An admin publishes a challenge to 1,500 opted-in members. Page 1 (1,000 members) is fetched. Cursor fetch for page 2 throws. The caller sees `{ success: false, queued: 0 }`. No members receive any notification. The error is logged but the admin has no actionable indication of partial state (were any records queued?).

**Recommendation:** Add a `__setNextError` capability to the wix-data mock, then add tests:

```js
it('queryAll cursor failure on page 2 — logs error and surfaces failure, does not silently drop page-1 members', async () => {
  const memberIds = Array.from({ length: 1001 }, (_, i) => `mem-p2fail-${i}`);
  seedOptedIn(memberIds);
  __setNextPageError(PREFS_COLLECTION, new Error('cursor expired'));

  const result = await notifyChallengePublished(makeChallenge());

  // Current behavior: success:false, queued:0 — documents the silent partial-drop
  // Assert logError was called with enough context for an operator to diagnose
});
```

This test would document (and prompt a fix for) the silent partial-drop behavior.

---

### FINDING 2 — HIGH: `hasSome()` with unbounded member ID arrays — mock silently masks a production Wix limit

**Location:** `src/backend/gamificationNotifs.web.js` lines 358–369; new tests at lines 483–512

**Issue:** In `checkStreakMilestoneNotifications`, after collecting all streak members via `queryAll()`, the code runs:

```js
const sentItems = await queryAll(
  wixData.query(STREAK_NOTIFICATIONS_COLLECTION)
    .hasSome('memberId', memberIds)   // memberIds may be 1001 or 2001 items
    .eq('milestone', STREAK_MILESTONE_DAY)
);
const prefsItems = await queryAll(
  wixData.query(MEMBER_NOTIFICATION_PREFS_COLLECTION)
    .hasSome('memberId', memberIds)   // same array
);
```

The Wix Data `hasSome()` operator has an undocumented but real production limit of **100 items** in the values array. Passing 1,001 or 2,001 member IDs will either throw or silently truncate results in production.

The wix-data mock's `hasSome()` implementation (line 111–118 in the mock) has no such limit — it happily processes all 2,001 IDs. So the new multi-page streak tests pass in CI while the production code is broken for any dataset exceeding 100 members in this second/third query.

**Hidden errors that could be swallowed:** Production `hasSome()` rejection with "Query values array exceeds maximum length" — caught by the outer `try/catch` and returned as `{ queued: 0, skipped: 0, errors: 0 }`, silently notifying nobody.

**Recommendation:** The production code needs to chunk the `memberIds` array into batches of ≤100 before calling `hasSome()`. Tests should verify chunked behavior. The mock should optionally enforce the limit. A test with 101+ memberIds should assert the chunked path is taken and all results are collected.

---

### FINDING 3 — HIGH: Exact-boundary streak test omits `StreakMilestoneNotifications` assertion

**Location:** `tests/gamificationNotifs.test.js` lines 471–481 (new test)

**Issue:** The exact-boundary challenge test (1,000 members, `notifyChallengePublished`) asserts both `EmailQueue.toHaveLength(1000)` and `SMSQueue.toHaveLength(1000)`. The symmetrical exact-boundary streak test (1,000 members, `checkStreakMilestoneNotifications`) asserts only `EmailQueue.toHaveLength(1000)` — it **omits** the assertion on `StreakMilestoneNotifications.toHaveLength(1000)`.

This collection is the idempotency guard. If the guard fails to write all 1,000 records on an exact-boundary page, the test would not catch it. The 2-page and 3-page streak tests do include this assertion, making the exact-boundary test the weakest of the three.

**Recommendation:**

```js
expect(__getInserted(STREAK_NOTIFS_COLLECTION)).toHaveLength(1000);
```

Add this as the final assertion in the exact-boundary streak test (after line 480).

---

### FINDING 4 — MEDIUM: PR description misreports the test count

**Issue:** The PR body states "2 new tests, 39 total in `gamificationNotifs.test.js`". The diff actually adds **6 new `it()` blocks** (3 for `notifyChallengePublished`, 3 for `checkStreakMilestoneNotifications`), bringing the total to **43**, not 39. The test plan checklist also only documents 2 of the 6 tests.

This is low-severity documentation drift, but it becomes a problem when someone uses the test count to verify CI completeness.

---

### FINDING 5 — MEDIUM: Dead variables in wix-data mock `find()` — could mask future mock bugs

**Location:** `tests/__mocks__/wix-data.js` lines 193–194 (not part of this PR, but exposed by it)

```js
const pageItems = items.slice(skipVal, skipVal + limitVal);  // line 193 — dead, never read
const nextSkip = skipVal + limitVal;                         // line 194 — dead, never read
```

Both variables are computed then discarded; `makePage(skipVal)` recomputes everything internally. This is not introduced by this PR but the new pagination tests depend on `makePage()` behavior — if a future developer tries to fix pagination by editing these dead variables instead of `makePage()`, the fix will silently not work and the tests will still pass.

---

### What the new tests do well

- The 3-page test (`THREE_PAGES = 2001`) is the right design choice: it distinguishes `if` from `while` in the cursor loop, which is the exact regression it's documenting.
- The exact-boundary test (1,000 members, `hasNext()=false`) correctly documents that no cursor traversal should occur at the limit.
- Both `EmailQueue` and `SMSQueue` are asserted in the challenge tests, confirming both fan-out legs complete.
- The `OVER_PAGE_LIMIT` / `THREE_PAGES` named constants improve readability over magic numbers.

---

### Summary

| # | Severity | Issue |
|---|----------|-------|
| 1 | CRITICAL | No test for mid-pagination `result.next()` failure; silent member-drop in production |
| 2 | HIGH | `hasSome()` called with unbounded arrays; mock masks production Wix 100-item limit |
| 3 | HIGH | Exact-boundary streak test missing `StreakMilestoneNotifications` length assertion |
| 4 | MEDIUM | PR description reports 2 tests / 39 total; actual count is 6 new / 43 total |
| 5 | MEDIUM | Dead variables in mock `find()` could misdirect future pagination debuggers |

Findings 1 and 2 represent silent failure modes that would not be caught by these tests in CI but would manifest as zero-notification outcomes for large member populations in production.

DreadPirateRobertz:
superpowers:pr-test-analyzer dispatched by godfrey

## Pagination Coverage Analysis — gamificationNotifs queryAll tests

### Summary

The PR adds 6 pagination tests (3 for `notifyChallengePublished`, 3 for `checkStreakMilestoneNotifications`) covering exact-boundary (1 000), 2-page (1 001), and 3-page (2 001) scenarios. The mock infrastructure in `tests/__mocks__/wix-data.js` correctly implements cursor pagination via `hasNext()`/`next()`, so these tests exercise real multi-page traversal logic. Coverage of the happy-path cursor loop is solid.

However, several important pagination paths are absent.

---

### Critical Gaps

**1. No runaway guard test — criticality 8/10**

`queryAll()` in production has no iteration cap. The `while (result.hasNext())` loop will run forever if the Wix data layer ever returns a cursor that never resolves to `hasNext() === false` (e.g., a bug in the real SDK or a mock regression). There is no `MAX_PAGES` guard in the production code, and no test that documents this absence as a known risk or that verifies the loop terminates correctly under a simulated misbehaving cursor. If a future developer adds a guard, there is no test to verify it fires at the right threshold. Recommend either adding a guard to production and a test that confirms it, or adding a comment + test that explicitly documents the no-guard decision.

Failure it would catch: infinite loop causing Wix serverless timeout (14 s hard limit), dropping all in-flight notifications silently.

**2. Pagination on secondary queryAll calls in `checkStreakMilestoneNotifications` — criticality 7/10**

`checkStreakMilestoneNotifications` calls `queryAll()` three times: once for `MemberPoints` (streakItems), once for `StreakMilestoneNotifications` (sentItems dedup), and once for `MemberNotificationPrefs` (prefsItems). The new pagination tests only seed `MemberPoints` beyond the 1 000-item boundary. The `sentItems` and `prefsItems` queries are left at their default small sizes.

A regression where `queryAll` silently truncates the `sentItems` or `prefsItems` result set would cause: (a) members who were already notified being re-notified (because `alreadySent` set is incomplete), or (b) members whose prefs say `streakReminders: false` getting incorrectly queued (because `prefsMap` is incomplete). Neither bug would be caught by the current tests.

Concrete test to add: seed 1 001 members in `MemberPoints`, seed 500 of them in `StreakMilestoneNotifications` (alreadySent) — assert `result.skipped === 500` and `result.queued === 501`. This forces the sentItems cursor path.

**3. `queryAll` error on the second page (mid-traversal failure) — criticality 6/10**

The existing error test (`__setQueryError`) only covers a failure on the initial `.find()` call. There is no test for a scenario where page 1 succeeds but `result.next()` throws on page 2. In production this would surface as an uncaught exception propagating out of `queryAll` and being swallowed by the outer `try/catch`, resulting in `{ queued: 0 }` for `notifyChallengePublished` or `{ queued: 0, skipped: 0, errors: 0 }` for `checkStreakMilestoneNotifications` — a silent zero-send on large member lists.

The mock supports this via `__onInsert` pattern but there is no equivalent hook for `next()` failures. If the mock were extended to support mid-traversal errors, a test here would be high value.

---

### Important Improvements

**4. `notifyChallengePublished` 3-page test does not verify SMSQueue content — criticality 5/10**

The 3-page test (`THREE_PAGES = 2001`) checks `EmailQueue.length` and `SMSQueue.length` but does not assert that every `SMSQueue` item has the correct `memberId`, `messageType`, or `status`. A bug where pages 2+ produce malformed SMS records (e.g., missing `memberId`) would pass all current assertions. The 2-member happy-path test covers field correctness, but not at pagination scale. Acceptable tradeoff but worth noting.

**5. Exact-boundary for `checkStreakMilestoneNotifications` does not assert `StreakMilestoneNotifications` length — criticality 4/10**

The exact-boundary test for streaks checks `EmailQueue.length === 1000` and return counts but omits `__getInserted(STREAK_NOTIFS_COLLECTION).toHaveLength(1000)`. The 2-page and 3-page tests include this assertion. Inconsistency could mask a regression in the idempotency record-insertion path at the boundary size.

---

### Positive Observations

- The mock's `makePage()` closure correctly captures `limitVal` from the outer query builder scope, so `.limit(1000)` in `queryAll` is faithfully respected across all cursor pages — the pagination tests are genuinely exercising cursor traversal, not a trivial linear scan.
- The 3-page test (`if` vs `while` regression detector) is well-motivated and would catch the most common queryAll implementation mistake.
- `seedOptedIn` and `seedStreakMembers` helpers keep the large-seed tests readable.
- `OVER_PAGE_LIMIT` and `THREE_PAGES` as named constants improve intent clarity over raw magic numbers.
- Existing error-path tests (`__setQueryError`, `__setInsertError`, `__onInsert`) provide solid baseline coverage that the pagination tests correctly build on.

---

### Verdict

The core 2-page / 3-page / exact-boundary paths are well covered. The two most important gaps are the missing mid-traversal (`sentItems`/`prefsItems`) pagination tests for `checkStreakMilestoneNotifications` secondary queries (gap #2, criticality 7) and the absence of any runaway guard documentation or test (gap #1, criticality 8). Neither is a blocker for merge but both represent real production risk on large member lists.

PR #1022, opened 2026-04-13
feat(cf-pyw): gamification chip on Collection/Category product cards

DreadPirateRobertz:
## Summary
- Adds `renderCardGamificationChip` + `formatCardChipLabel` helpers in `GamificationProductChip.js`.
- Wires the chip into Category Page's `#productGridRepeater` onItemReady so members see their tier + points on each product card in browse context.
- Preloads the loyalty account once per page via a new deferred section; chip/service modules are dynamic-imported to preserve the page import budget (cap 27).
- Unauthenticated visitors and templates without `#gridGamificationChip` are silent no-ops.

## Test plan
- [x] `npx vitest run tests/gamificationProductChip.test.js` — 34 pass
- [x] `npx vitest run tests/importBudget.test.js` — passes (Category Page still at 27)
- [x] Coverage for `src/public/GamificationProductChip.js`: 100% stmts/funcs/lines, 87.23% branches
- [ ] Browser verify once Wix Studio template exposes `#gridGamificationChip` on product cards

Bead: cf-pyw

### DIFF
--- a/src/pages/Category Page.js
+++ b/src/pages/Category Page.js
@@ -541,6 +541,16 @@ async function preloadLoyaltyAccount() {
     ]);
     _renderCardChipFn = chipMod.renderCardGamificationChip;
     _loyaltyAccount = await loyaltyMod.getMyLoyaltyAccount();
+    // Refresh already-rendered cards — onItemReady fires before this resolves
+    // for items visible at page load, so chips would be blank without this. (CF-pyw)
+    try {
+      const repeater = $w('#productGridRepeater');
+      if (repeater && _renderCardChipFn) {
+        repeater.forEachItem(($item) => {
+          _renderCardChipFn($item, _loyaltyAccount);
+        });
+      }
+    } catch (e) {}
   } catch (e) {
     _loyaltyAccount = null;
   }
--- a/src/pages/Category Page.js
+++ b/src/pages/Category Page.js
@@ -40,6 +40,8 @@ let _debounceTimer = null;
 let _basicFilterTimer = null;
 let _filterSessionState = {}; // persists across category nav within session
 let _wishlistSet = new Set(); // cached wishlist status for product cards
+let _loyaltyAccount = null; // cached loyalty account for per-card gamification chip (CF-pyw)
+let _renderCardChipFn = null; // lazy-loaded render helper (dynamic import keeps page import budget)
 
 // sanitizeInput aliased from shared module
 const sanitizeInput = sanitizeFilterInput;
@@ -109,6 +111,7 @@ $w.onReady(async function () {
     { name: 'filterControls', init: () => initFilterControls(), critical: false },
     { name: 'advancedFilters', init: () => initAdvancedFilters(currentPath), critical: false },
     { name: 'wishlistStatus', init: () => preloadWishlistStatus(), critical: false },
+    { name: 'loyaltyAccount', init: () => preloadLoyaltyAccount(), critical: false },
     { name: 'recentlyViewed', init: () => initRecentlyViewed(), critical: false },
     { name: 'quickView', init: () => initQuickViewHandlers(), critical: false },
     { name: 'categoryMeta', init: () => injectCategoryMeta(currentPath), critical: false },
@@ -525,6 +528,24 @@ async function preloadWishlistStatus() {
   } catch (e) {}
 }
 
+// ── Loyalty Account Preload ─────────────────────────────────────────
+// Fetch the member's loyalty account once per page so product cards can
+// show a gamification chip (tier + points) in browse context.
+// Silently no-ops for unauthenticated visitors. CF-pyw
+
+async function preloadLoyaltyAccount() {
+  try {
+    const [loyaltyMod, chipMod] = await Promise.all([
+      import('backend/loyaltyService.web'),
+      import('public/GamificationProductChip.js'),
+    ]);
+    _renderCardChipFn = chipMod.renderCardGamificationChip;
+    _loyaltyAccount = await loyaltyMod.getMyLoyaltyAccount();
+  } catch (e) {
+    _loyaltyAccount = null;
+  }
+}
+
 // ── Product Grid ────────────────────────────────────────────────────
 // Enhanced product cards with hover effects, quick-view, and badges
 
@@ -569,6 +590,9 @@ function initProductGrid() {
       // Card container structure: white bg, 12px radius, shadow, hover
       try { styleCardContainer($item('#gridCard')); } catch (e) {}
       try { initCardHover($item('#gridCard')); } catch (e) {}
+
+      // Gamification chip: member's tier + points in browse context (CF-pyw)
+      if (_renderCardChipFn) _renderCardChipFn($item, _loyaltyAccount);
 
       // Product image with placeholder fallback + SEO alt text
       const category = wixLocationFrontend.path?.[0] || '';
--- a/src/public/GamificationProductChip.js
+++ b/src/public/GamificationProductChip.js
@@ -82,6 +82,52 @@ export function renderTierChip($wFn, account) {
 }
 
 /**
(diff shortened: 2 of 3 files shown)


DreadPirateRobertz (commented):
## Code Review — feat(cf-pyw): gamification chip on Collection/Category product cards

Confidence: **92/100** — passing threshold met.

---

### What was done well

- Dynamic-import pattern is correctly applied: both `loyaltyService.web` and `GamificationProductChip.js` are imported via `Promise.all` inside the deferred section, keeping static imports at the pre-existing cap of 27 for Category Page. `importBudget.test.js` confirms this.
- `preloadLoyaltyAccount` is registered as `critical: false`, correctly matching the deferred pattern used by `preloadWishlistStatus` immediately above it.
- Auth handling is sound: `getMyLoyaltyAccount` throws for unauthenticated visitors; the `catch` sets `_loyaltyAccount = null` and leaves `_renderCardChipFn` unassigned. The `if (_renderCardChipFn)` guard in `onItemReady` means unauthenticated visitors trigger zero chip work — not a try/catch swallow, but an actual early-exit. Verified.
- Template missing `#gridGamificationChip`: `$item('#gridGamificationChip')` returns `null` in that case; the `if (!chip) return;` guard fires and no DOM mutation occurs. This is a true silent no-op.
- Test count verified: 34 tests pass on the PR branch (20 pre-existing + 14 new). The PR description says 10 new — the actual count is 14 (6 `formatCardChipLabel` + 8 `renderCardGamificationChip`). Minor description inaccuracy but tests are good.
- JSDoc on both new exports is clear and accurate.

---

### Issues Found

#### Important — Race condition: `_renderCardChipFn` unset when first cards render

`preloadLoyaltyAccount` is deferred. `initProductGrid` is **critical** (runs first, above-fold). The repeater's `onItemReady` callback fires for each card as data loads — this can happen before `preloadLoyaltyAccount` has resolved. When it does, `_renderCardChipFn` is still `null`, so the guard `if (_renderCardChipFn)` silently skips the chip for the initial card batch.

This is the same race that exists with `_wishlistSet` (batchCheckWishlistStatus also deferred) — but wishlist uses a pre-fetched Set that is populated asynchronously and then the repeater re-renders. Here there is no re-render trigger after the loyalty preload resolves. Members will see no chip on first page load in most cases (deferred section resolves after repeater paint).

Mitigation options (pick one):
1. Move `loyaltyAccount` to `critical: true` — adds latency to LCP, not preferred.
2. After `preloadLoyaltyAccount` resolves, call `repeater.refresh()` or re-invoke `onItemReady` callbacks — correct pattern used elsewhere for async card decorations.
3. Store the pending promise on the module-level and `await` it inside `onItemReady` — adds per-card await but avoids the race.

This is not a crash — it's a silent miss. Members will never see the chip on cold load unless the deferred section happens to beat the repeater paint (rare). Recommend option 2 to match the wishlist pattern.

#### Suggestion — `_renderCardChipFn` not cleared on preload error

In the catch block of `preloadLoyaltyAccount`, only `_loyaltyAccount` is reset to null. `_renderCardChipFn` retains whatever value it had from a prior call (e.g. if `preloadLoyaltyAccount` is re-invoked after a category navigation refresh and the second call throws mid-way after setting `_renderCardChipFn` but before fetching the account). The guard `if (_renderCardChipFn) _renderCardChipFn($item, null)` would then render a hidden chip (correct behavior, but confusing intent). Safe to add `_renderCardChipFn = null;` to the catch block for symmetry.

#### Suggestion — Branch coverage gap (87.23%) explained, acceptable with caveat

Uncovered branches are all pre-existing in the file (not introduced by this PR):
- Line 25: `!account.tier` sub-branch of `!account || !account.tier` — the case where account is non-null but `.tier` is absent is not isolated in a dedicated test for `formatTierChipLabel`
- Line 27: ternary fallback `account.tier` when `getTierIcon` returns falsy (unknown tier string)
- Line 80: `err?.message ?? err` — the `err.message` absent branch on `renderPointsChip` catch
- Lines 141-142: `opts.$w || globalThis.$w` — the path where opts are omitted (uses `globalThis.$w`); tests always inject `$w`

None of these are in the new code. The 87.23% is a carry-forward, not a regression. Acceptable as-is, but these would be worth a dedicated cleanup bead given the module now has wider usage.

#### Suggestion — `formatCardChipLabel` test for separator is loose

```js
expect(label).toMatch(/[·•|·\-]/); // some separator between
```

The actual separator is `·` (U+00B7 middle dot). The regex accepts any of five different characters including a bare hyphen. This test will pass even if the separator format changes to something unintended. Recommend: `expect(label).toContain(' · ')` to assert the exact separator and surrounding spaces.

---

### Plan Alignment

All stated deliverables are present: `renderCardGamificationChip`, `formatCardChipLabel`, Category Page wiring, deferred preload, import budget preserved. The one unchecked test plan item (browser verify) is correctly flagged as blocked on template element availability — appropriate.

---

### Summary

The implementation is structurally sound and the test coverage on the new code is thorough. The race condition between `initProductGrid` (critical) and `preloadLoyaltyAccount` (deferred) is the primary concern — on typical page loads members will not see the chip because the repeater paints before the deferred section resolves. This should be addressed before merge by adding a repeater refresh after the loyalty account resolves, following the pattern established for other async card decorations.

Blocking on: race condition fix.
Non-blocking: catch-block symmetry, separator assertion tightness.

DreadPirateRobertz (commented):
## Code Review — cf-pyw gamification chip (Category/Collection cards)

**Verdict: MERGE-READY** with one tracked follow-on item. No must-fix blockers.

### What's solid
- Dynamic import pattern preserves the 27-import page cap ✓
- Three-layer null safety in `renderCardGamificationChip` (null chip, null account, try/catch) ✓
- IDOR eliminated at architecture level — `getMyLoyaltyAccount` resolves from authenticated session, no memberId from client ✓
- 10 new tests cover all four composite states + edge cases ✓
- CI failures (trailDataModel, reviews.integration) are pre-existing and unrelated ✓

### Should-fix (on this PR or small follow-on)

**1. Assignment order readability in `preloadLoyaltyAccount` (confidence 92)**

```js
// Current — reads as if _renderCardChipFn is set before _loyaltyAccount awaits
_renderCardChipFn = chipMod.renderCardGamificationChip;
_loyaltyAccount = await loyaltyMod.getMyLoyaltyAccount();
```

Both are assigned atomically from `Promise.all` (safe at runtime), but the visual ordering suggests a separate await and is a maintainability hazard. Swap order + add comment clarifying both are set atomically.

**2. Race condition comment needs expansion (confidence 88)**

The comment "Silently no-ops for unauthenticated visitors" is accurate but incomplete — it omits the timing case where authenticated members on fast connections may not see chips on the initial render (onItemReady fires before deferred preload completes). The graceful no-op behavior is intentional and correct; the comment should say so explicitly so future devs don't "fix" it without understanding the trade-off.

Follow-on bead to track: add `$w('#productGridRepeater').forEachItem(...)` refresh after `preloadLoyaltyAccount` resolves to backfill chips on already-rendered cards.

### Suggestion (nice-to-have)
Add test for `formatCardChipLabel({ points: { balance: 0 } })` — zero balance, no tier — to document whether "0 pts" is intentional for untiered members.

---
**Reviewer**: melania (PM) | **Confidence filter**: ≥80 | 26 tool uses, 3 files reviewed

DreadPirateRobertz:
**PM update**: Reconciling the two reviews — siding with the first review's verdict on the race condition. This is a **must-fix before merge**.

 fires during critical init;  is deferred. On typical page loads, repeater cards render before the deferred section resolves → chips never display on cold load. That defeats the feature.

**Required fix**: After  resolves successfully, call:
```js
('#productGridRepeater').forEachItem(() => {
  if (_renderCardChipFn) _renderCardChipFn(, _loyaltyAccount);
});
```
This backfills chips on already-rendered cards. See  pattern for reference.

morgott: push the fix and re-ping.

PR #1044, opened 2026-04-13
feat(cf-x7n): GA4 consent-gated wrapper — parity with TikTok/Pinterest

DreadPirateRobertz:
## Summary
- Adds `fireTrackedGA4Event` to pixelConsentService with analytics-only consent gate (GA4 is analytics-only per GDPR; advertising consent not required).
- Adds generic `fireGA4Event` to ga4Tracking.js so the consent layer can route any GA4 event through the shared wixWindow.trackEvent bridge.
- Queue flush on analytics-only grant now fires GA4 entries while keeping TikTok/Pinterest entries queued until advertising consent is also granted.
- Closes the parity gap identified during cf-agr Phase 7 GA4 verification.

## Test plan
- [x] tests/pixelConsentService.test.js: 64/64 pass (55 existing + 9 new GA4-specific)
- [x] Full vitest suite: 1083 files / 39539 pass / 0 fail

### DIFF
--- a/src/public/ga4Tracking.js
+++ b/src/public/ga4Tracking.js
@@ -15,6 +15,7 @@ import {
   buildSearchEvent,
   buildViewCartEvent,
 } from 'backend/analyticsHelpers.web';
+import wixPrivacy from 'wix-privacy-frontend';
 
 let _wixWindow = null;
 
@@ -29,12 +30,25 @@ async function getWixWindow() {
   return _wixWindow;
 }
 
+// GA4 is classified analytics-only under GDPR; advertising consent is not
+// required. Every fire* entry point gates on this to ensure raw callers
+// that don't route through pixelConsentService still respect user consent.
+export function _hasAnalyticsConsent() {
+  try {
+    const { policy } = wixPrivacy.getCurrentConsentPolicy();
+    return policy.analytics === true;
+  } catch (e) {
+    return false;
+  }
+}
+
 /**
  * Fire a GA4/Meta Pixel ViewContent event for a product page view.
  * @param {Object} product - Wix product object
  */
 export async function fireViewContent(product) {
   try {
+    if (!_hasAnalyticsConsent()) return;
     const ww = await getWixWindow();
     if (!ww?.trackEvent) return;
     const payload = await buildViewContentEvent(product);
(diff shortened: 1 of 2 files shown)


DreadPirateRobertz (commented):
5-agent review — 1 critical + 1 important before merge.

**CRITICAL (100): All existing GA4 call sites bypass the consent gate**
`fireTrackedGA4Event()` exists but zero callers use it. Every real GA4 emission in the codebase still calls raw ungated functions directly: `firePurchase` (Thank You Page.js:84), `fireAddToCart` (AddToCart.js:73/135/191), `fireViewCart` (Cart Page.js:77), `fireInitiateCheckout` (Checkout.js:303), `fireSearch`/`fireViewItemList` (Search Results.js:92-93), plus `fireCustomEvent` across masterPage/Blog/Loyalty/CartUpsell/etc. The consent gate is dead code. Fix: add `_hasAnalyticsConsent()` check inside the raw functions in ga4Tracking.js (simpler, guaranteed coverage), OR migrate all call sites.

**IMPORTANT (80): fireGA4Event does not sanitize event name**
`fireCustomEvent` sanitizes via `eventName.replace(/[^a-zA-Z0-9_]/g, '')` but the new `fireGA4Event` does not. Add the same sanitization.

Consent model design (analytics-only vs advertising for TikTok/Pinterest) is correct. Tests for the new wrapper are solid. Fix the two issues and re-push.

DreadPirateRobertz:
Review fixes pushed (2833576b):

**Consent gate moved to raw fire* functions** — addresses the dead-code concern. Every ga4Tracking.js entry point (fireViewContent, fireAddToCart, fireInitiateCheckout, firePurchase, fireAddToWishlist, fireCustomEvent, fireViewItemList, fireSearch, fireViewCart, fireGA4Event, and the scroll depth tracker) now calls `_hasAnalyticsConsent()` before firing. No call-site migration required — existing callers (Thank You Page, AddToCart, Checkout, Cart, Search Results, Category, Blog, Newsletter, Loyalty, masterPage, FooterSection, SaveForLater) become automatically GDPR-compliant.

**fireGA4Event now sanitizes event name** — strips non-alphanumeric/underscore chars (mirrors fireCustomEvent pattern). Drops silently if the sanitized name is empty.

**Tests:** 14 new in ga4TrackingDeep.test.js covering consent-denial for every fire* entry point + sanitization edge cases. Full suite 1083 files / 39557 pass / 0 fail.

Uses the existing tests/__mocks__/wix-privacy-frontend.js alias (defaults analytics=true, `__setPolicy` for denied paths) — no test-infra changes needed.

PR #1047, opened 2026-04-13
feat(cf-2tm): wire TikTok pixel events on PDP/cart/wishlist/purchase

DreadPirateRobertz:
## Summary
- Fan out TikTok pixel fires from the four existing GA4 entry points in `ga4Tracking.js` — so ViewContent (PDP), AddToCart (PDP+sticky+bundle), Purchase (Thank You), and AddToWishlist (PDP+category card) all reach TikTok through `pixelConsentService.fireTrackedTikTokEvent` (advertising-consent gated, Purchase dedup handled).
- New call from `WishlistCardButton.js` card heart adds GA4+TikTok for category-grid saves (previously only `trackEvent`).
- Seed `tikTokPixel.PIXEL_ID` with the placeholder `'TIKTOK_PIXEL_ID'` literal and short-circuit `initTikTokPixel()` on that value so no broken SDK script loads until Stilgar/Wix Dashboard supplies the real ID.

Closes cf-2tm. Follow-on to the cf-31l verification audit.

## Test plan
- [x] `npx vitest run tests/ga4TikTokBridge.test.js tests/ga4Tracking.test.js tests/ga4TrackingDeep.test.js tests/tikTokPixel.test.js tests/pixelConsentService.test.js tests/wishlistCardButton.test.js tests/wishlistCardButtonHardening.test.js tests/addToCart.test.js tests/funnelTracker.test.js tests/masterPage.test.js` — 553 passed
- [x] New `tests/ga4TikTokBridge.test.js` asserts each GA4 fire fans out to `fireTrackedTikTokEvent` with the correct event name + TikTok payload (content_id/order_id, value, currency) and that TikTok failures don't propagate to callers

### DIFF
--- a/src/public/WishlistCardButton.js
+++ b/src/public/WishlistCardButton.js
@@ -10,6 +10,7 @@
  * @module WishlistCardButton
  */
 import { trackEvent } from 'public/engagementTracker';
+import { fireAddToWishlist } from 'public/ga4Tracking';
 import { colors } from 'public/designTokens.js';
 
 // SVG heart icons — same visual language as socialWishlist.js
@@ -104,6 +105,7 @@ export function initCardWishlistButton($item, product, isWishlisted) {
           wishlisted = true;
           setHeartState($item, true, product.name);
           trackEvent('wishlist_add', { productId: product._id, source: 'product_card' });
+          fireAddToWishlist(product);
         }
       } catch (e) {
         console.error('[WishlistCardButton] toggle failed for product:', product._id, e);
--- a/src/public/ga4Tracking.js
+++ b/src/public/ga4Tracking.js
@@ -16,6 +16,14 @@ import {
   buildViewCartEvent,
 } from 'backend/analyticsHelpers.web';
 import wixPrivacy from 'wix-privacy-frontend';
+import { fireTrackedTikTokEvent } from 'public/pixelConsentService';
+
+// Fire a TikTok pixel event alongside a GA4 event. Consent gating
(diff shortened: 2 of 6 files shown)


DreadPirateRobertz (commented):
**Code review — cf-2tm TikTok pixel wiring (confidence 87) — PASS**

Wiring is correct: all 4 events (ViewContent, AddToCart, Purchase, AddToWishlist) fan out from ga4Tracking.js entry points via `_fireTikTok` → `fireTrackedTikTokEvent`, consent-gated by pixelConsentService. Purchase dedup shared. No raw `fireTikTokEvent` calls. No hardcoded PIXEL_ID.

**Two follow-up items (not blocking — creating hardening bead):**

1. **Consent-guard placement**: `_fireTikTok` calls sit after `if (!_hasAnalyticsConsent()) return` — users with advertising-only consent (no analytics) will never have TikTok events queued in pixelConsentService. In practice Wix presents both consent categories together so real-world impact is minimal, but architecturally `_fireTikTok` should be called independently of the GA4 analytics guard and let pixelConsentService be the sole TikTok consent arbiter.

2. **Missing consent-false test**: No assertion that `fireTrackedTikTokEvent` is NOT called when `_hasAnalyticsConsent()` returns false. A test like `mockPrivacy(analytics: false); await fireViewContent(...); expect(fireTrackedTikTokEvent).not.toHaveBeenCalled()` is needed.

Merging and opening cf-NEW for these hardening items.

PR #1045, opened 2026-04-13
fix(cf-94s): normalize mainMedia to CDN URL for og:image + schema image

DreadPirateRobertz:
## Summary — verify(SEO): OG image generation

**Verification findings (all green):**
- OG tags wired on the product page: `Product Page.js` → `injectProductMeta` (critical init section) → `seoHelpers.getProductOgTags` → `head.setMetaTag`
- og:title / og:description / og:url set per-product with name, stripped description, canonical slug URL
- Product schema (JSON-LD Product) + OG tags generated in the same module

**Gap found: og:image / twitter:image / JSON-LD image emitted `product.mainMedia` unchanged.** Wix Stores can expose `mainMedia` as a `wix:image://v1/<id>` URI; crawlers (Google / Facebook / Pinterest) cannot resolve that scheme. Sister modules `facebookCatalog.web.js` and `http-functions.js` already normalize via `backend/utils/mediaHelpers.getImageUrl` for this reason — `seoHelpers.web.js` did not import it.

## Fix
Import `getImageUrl` in `seoHelpers.web.js` and wrap `mainMedia` at every emission point that goes to a crawler:
- `getProductOgTags` → og:image + twitter:image
- `getProductMetaTags` → HTML og:image attribute
- `buildImageArray` → JSON-LD image array (used by getProductSchema)
- Category page `ItemList` schema → per-product image

## Test plan
- [x] 3 new cases in `tests/seoHelpersDeep.test.js`: wix:image → CDN URL, twitter:image mirrors og:image, already-absolute URLs pass through
- [x] `npx vitest run tests/seoHelpers tests/productOgTags.test.js` — 348 pass (was 345)
- [ ] Live validator (requires deploy): share a wix:image-backed product through the Facebook Sharing Debugger + Twitter Card Validator

Bead: cf-94s (closes verify-and-fix for Phase 7 OG image generation)

PR #1049, opened 2026-04-13
test(cf-gjc): lock Pinterest feed CDN image normalization

DreadPirateRobertz:
## Summary
- cf-gjc verification: Pinterest catalog feed at `GET /_functions/pinterestProductFeed` is live, wired to `fetchAllProducts()` (no filter — emits the full `Stores/Products` collection), and already uses `getImageUrl()` to normalize `wix:image://` URIs into `https://static.wixstatic.com/media/...` CDN URLs for both `image_link` and `additional_image_link`.
- Adds a regression test to `tests/httpFunctions.test.js` that seeds a product whose `mainMedia` and `mediaItems[].src` are all raw `wix:image://v1/…` URIs, then asserts the TSV feed output contains normalized CDN URLs and **no** raw `wix:image://` URIs. This locks the feed against the same bug pattern fixed in PR #1045 for OG tags.

## Verification findings
- **Endpoint**: `src/backend/http-functions.js:429` → `get_pinterestProductFeed`
- **Data source**: `fetchAllProducts()` pages `Stores/Products` with `suppressAuth: true` and no filter — all live products are emitted. Local mocks can't count production rows; the feed will cover all 88 (or current count).
- **Fields emitted per row**: `id, title, description, link, image_link, price, availability, brand, google_product_category, condition, sale_price, product_type, additional_image_link` ✓
- **Image normalization**: `getImageUrl(p.mainMedia)` and `getImageUrl(m.src || m)` for each `mediaItems[1..5]` entry. `getImageUrl` itself is also covered by PR #1045's recursion fix for `{ url: '…' }` objects.
- **Rate limits**: Feed `Cache-Control: public, max-age=3600` — 1h TTL aligns with Pinterest's 6-refreshes-per-day limit (every 4h).

## Test plan
- [x] `npx vitest run tests/httpFunctions.test.js tests/pinterest` — 8 files, 406 passed (was 405)
- [x] No src/ changes — pure test lock, zero production risk.

Closes cf-gjc.

### DIFF
--- a/tests/httpFunctions.test.js
+++ b/tests/httpFunctions.test.js
@@ -328,6 +328,36 @@ describe('get_pinterestProductFeed', () => {
     await get_pinterestProductFeed();
     expect(__getLastFindOptions('Stores/Products')).toEqual({ suppressAuth: true });
   });
+
+  it('normalizes wix:image:// URIs in image_link to static.wixstatic.com CDN URLs', async () => {
+    __seed('Stores/Products', [{
+      _id: 'prod-wix',
+      name: 'Wix Image Product',
+      slug: 'wix-image-product',
+      price: 499,
+      discountedPrice: null,
+      mainMedia: 'wix:image://v1/abc123.jpg/photo.jpg#originWidth=1200',
+      description: 'Test product.',
+      inStock: true,
+      collections: ['futon-frames'],
+      _updatedDate: new Date('2026-01-15'),
+      mediaItems: [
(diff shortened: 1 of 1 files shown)


DreadPirateRobertz (commented):
PASS — 1 regression test locking CDN normalization in Pinterest feed. PR #1045 fix (getImageUrl on media.url) now covered in the feed path too. Merging.

PR #1050, opened 2026-04-13
fix(cf-5rt): decouple TikTok fan-out from GA4 analytics consent guard

DreadPirateRobertz:
## Summary
- Addresses cf-2tm PR #1047 review (confidence 87): `_fireTikTok` calls sat inside each GA4 analytics-consent early-return, so an advertising-only user (analytics denied, advertising granted) never reached the `pixelConsentService` queue.
- Moves `_fireTikTok` above the analytics guard in `fireViewContent` / `fireAddToCart` / `firePurchase` / `fireAddToWishlist`. `pixelConsentService.fireTrackedTikTokEvent` is now the sole arbiter of TikTok consent (it already enforces both analytics + advertising and queues until consent resolves).

## Test plan
- [x] TDD: new "consent decoupling" describe block in `tests/ga4TikTokBridge.test.js` mocks `wix-privacy-frontend` to `{ analytics: false, advertising: true }` and asserts each of the four entry points still dispatches to `fireTrackedTikTokEvent` — red pre-fix, green after moving the call
- [x] `npx vitest run tests/ga4TikTokBridge.test.js tests/ga4Tracking.test.js tests/ga4TrackingDeep.test.js tests/tikTokPixel.test.js tests/pixelConsentService.test.js` — 163/163 pass

### DIFF
--- a/src/public/ga4Tracking.js
+++ b/src/public/ga4Tracking.js
@@ -56,6 +56,15 @@ export function _hasAnalyticsConsent() {
  */
 export async function fireViewContent(product) {
   try {
+    // TikTok fan-out runs before the analytics-only guard: pixelConsentService
+    // enforces its own (analytics + advertising + queueing) consent contract,
+    // so advertising-only users still reach TikTok. (cf-5rt)
+    _fireTikTok('ViewContent', {
+      content_id: product?._id,
+      content_name: product?.name,
+      value: product?.price,
+      currency: 'USD',
+    });
     if (!_hasAnalyticsConsent()) return;
     const ww = await getWixWindow();
     if (!ww?.trackEvent) return;
@@ -63,12 +72,6 @@ export async function fireViewContent(product) {
     if (payload && Object.keys(payload).length > 0) {
       ww.trackEvent('ViewContent', payload);
     }
-    _fireTikTok('ViewContent', {
-      content_id: product?._id,
-      content_name: product?.name,
-      value: product?.price,
-      currency: 'USD',
-    });
   } catch (e) {
     // GA4 tracking is non-critical
   }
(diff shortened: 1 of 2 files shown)


DreadPirateRobertz (commented):
**BLOCK — one required test missing (confidence 91)**

Implementation is correct: `_fireTikTok` is now before all analytics guards in all 4 functions. The `analytics: false, advertising: true` path is tested. 

**Missing test (required by cf-5rt spec):** `analytics: false, advertising: false` — verify `_fireTikTok` is still called (unconditionally) even with zero consent, so `pixelConsentService` is the sole suppressor. Add to `tests/ga4TikTokBridge.test.js`:

```js
it('still calls fireTrackedTikTokEvent when all consent denied (pixelConsentService queues/suppresses)', async () => {
  getCurrentConsentPolicy.mockReturnValue({ policy: { analytics: false, advertising: false } });
  await fireViewContent({ _id: 'p5', price: 50 });
  expect(fireTrackedTikTokEvent).toHaveBeenCalledWith(
    'ViewContent',
    expect.objectContaining({ content_id: 'p5' }),
  );
});
```

Also: note in PR body that `fireInitiateCheckout` intentionally has no TikTok fan-out (pre-existing omission, not in scope).

One test gap only — re-push and this is ready.

DreadPirateRobertz:
Added coverage for the analytics=false + advertising=false path: new `fires TikTok unconditionally when analytics AND advertising are both denied (pixelConsentService queues)` test asserts all four entry points (ViewContent/AddToCart/Purchase/AddToWishlist) still call fireTrackedTikTokEvent — pixelConsentService remains the sole arbiter of queue-vs-drop. Pushed 04b0eb5f. 10/10 tests green on tests/ga4TikTokBridge.test.js.

DreadPirateRobertz (commented):
PASS (97) — no-consent test confirmed: all 4 entry points assert fireTrackedTikTokEvent called unconditionally when analytics+advertising both false. pixelConsentService is the sole suppressor. CI baseline failure is pre-existing (ratchet/masterPage), not a regression from this PR. Merging.

PR #1051, opened 2026-04-13
feat(cf-1d3): tier upgrade congratulations email + push

DreadPirateRobertz:
## Summary
- Adds `notifyTierUpgrade(memberId, newTier, previousTier)` in `gamificationNotifs.web.js` — sends `tier_upgrade_congratulations` email via `triggeredEmails.emailMember` + `TIER_CHANGED` push via `sendPushToMember`.
- Idempotent via `TierUpgradeNotifications` dedup collection on `(memberId, newTier)`; honours `MemberNotificationPrefs.tierUpdates` opt-out (strict `=== false`).
- Wired into `gamificationCore.web.js` tier-change branch alongside existing `deliverTierPerks` and `dispatchBusEvent('tier_upgraded')` — closes the cf-2dl audit gap (mobile got the bus event, web got coupons, but no personal congrats reached the member).
- Push failure is non-fatal: email is primary, dedup still recorded to prevent retry double-emails.

## Test plan
- [x] 12 new tests in `tests/tierUpgradeNotif.test.js` — happy path, dedup, per-tier uniqueness, opt-out, default opt-in, email failure (no dedup), push failure (non-fatal, dedup recorded), invalid input.
- [x] `npx vitest run tests/tierUpgradeNotif.test.js` → 12/12 pass.
- [x] Regression: `gamificationNotifs`, `streakAtRiskPush`, `gamificationCore`, `gamificationEventReceiver` → 259/259 pass.

## Schema note
Create `TierUpgradeNotifications` CMS collection:
- `memberId` (Text, indexed)
- `newTier` (Text, indexed)
- `previousTier` (Text)
- `sentAt` (Text / ISO DateTime)
- Unique constraint on `(memberId, newTier)` strongly recommended to harden dedup under concurrent tier-change events.

### DIFF
--- a/src/backend/gamificationCore.web.js
+++ b/src/backend/gamificationCore.web.js
@@ -243,6 +243,11 @@ export const receiveGamificationEvent = webMethod(
           const { deliverTierPerks } = await import('backend/rewardEngine.web');
           await deliverTierPerks(memberId, oldTier, newTier);
         } catch (e) { logError(`gamificationEventReceiver — deliverTierPerks failed for ${memberId}`, e); }
+        // cf-1d3: Personal tier-upgrade email + push (idempotent + opt-out aware)
+        try {
+          const { notifyTierUpgrade } = await import('backend/gamificationNotifs.web');
+          await notifyTierUpgrade(memberId, newTier, oldTier);
+        } catch (e) { logError(`gamificationEventReceiver — notifyTierUpgrade failed for ${memberId}`, e); }
       }
 
       // Phase 4: record wishlist add AFTER MemberPoints (best-effort)
--- a/src/backend/gamificationNotifs.web.js
+++ b/src/backend/gamificationNotifs.web.js
@@ -523,6 +523,14 @@ export async function notifyTierUpgrade(memberId, newTier, previousTier) {
         sentAt: new Date().toISOString(),
       }, { suppressAuth: true });
     } catch (dedupErr) {
+      // Uniqueness violation = another concurrent invocation won the dedup race
+      // and already recorded the notification. Surface sent:false so the caller
+      // does not treat our race-duplicate as a fresh send and retry.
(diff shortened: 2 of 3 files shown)


DreadPirateRobertz (commented):
**BLOCK — one Important: dedup race condition (confidence 92)**

Implementation is solid everywhere else — bus wiring, email, push, suppressAuth, opt-out, 12 tests all correct.

**Race condition in dedup insert (check-then-insert, not atomic):**

Two concurrent `tier_upgraded` events for same `(memberId, newTier)` can both pass the `existing.items.length > 0` check before either commits the insert. Current `dedupErr` catch logs the error but still returns `{ sent: true }` — so a race produces two emails silently.

Fix the catch block in `notifyTierUpgrade` to detect the uniqueness violation and treat it as idempotent:

```js
} catch (dedupErr) {
  // If this is a duplicate-key error, the race was lost — other process
  // already sent. Treat as success, not an error.
  if (dedupErr?.message?.includes('duplicate') || dedupErr?.code === 'WD_DUPLICATE_KEY') {
    return { sent: false, reason: 'already_sent' };
  }
  logError('gamificationNotifs.notifyTierUpgrade.dedupInsert', dedupErr);
  // dedup write failed for unknown reason — still return sent:true since
  // email already dispatched above
  return { sent: true, dedupWriteFailed: true };
}
```

Also: update the schema note in PR body from "strongly recommended" to "required — unique constraint on (memberId, tier) must be deployed before collection goes live." The collection is already created on staging — I'll add the unique index separately.

One catch block change only — re-push and this is ready.

DreadPirateRobertz (commented):
PASS (95) — race-duplicate catch confirmed: uniqueness violations return { sent: false, reason: 'race_duplicate' }, email already fired on that branch is correctly preserved. 13/13 tests. Merging.

PR #1057, opened 2026-04-13
feat(cf-h6w): wire push into challenge reminder cron

DreadPirateRobertz:
## Summary
- Extend `runDailyChallengeReminders` sendFn to also call `sendPushToMember(memberId, PUSH_EVENTS.STREAK_MILESTONE, { days })` alongside the existing `challenge_reminder` email.
- Push failures are logged + swallowed — they do NOT affect sent/failed counts or the `notifiedAt` cadence mark. Otherwise a flaky FCM response would re-send the entire batch on the next cron tick.
- Phase 6a gamification work. Closes cf-h6w.

## Test plan
- [x] `npx vitest run tests/lifecycleCron.test.js tests/challengeReminderService.test.js` — 102/102 pass
- [x] Full suite — 39,605/39,605 pass
- [x] New coverage: sendPushToMember called per eligible record, correct args, ignores push failure, no push for ineligible records

### DIFF
--- a/src/backend/lifecycleCron.web.js
+++ b/src/backend/lifecycleCron.web.js
@@ -190,9 +190,7 @@ export const runDailyChallengeReminders = webMethod(
         // notifiedAt mark (otherwise a flaky FCM response would re-send the
         // entire batch on the next cron tick). cf-h6w
         try {
-          await sendPushToMember(record.memberId, PUSH_EVENTS.STREAK_MILESTONE, {
-            days: String(record.currentStreakDays ?? ''),
-          });
+          await sendPushToMember(record.memberId, PUSH_EVENTS.CHALLENGE_REMINDER, {});
         } catch (pushErr) {
           console.error(`[lifecycleCron] challenge reminder push failed for ${record.memberId}:`, pushErr?.message);
         }
--- a/src/backend/lifecycleCron.web.js
+++ b/src/backend/lifecycleCron.web.js
@@ -170,6 +170,7 @@ export const runDailyChallengeReminders = webMethod(
   async () => {
     try {
       const { triggeredEmails } = await import('wix-crm-backend');
+      const { sendPushToMember, PUSH_EVENTS } = await import('backend/pushNotificationService.web');
 
       const sendFn = async (record) => {
         await triggeredEmails.emailMember(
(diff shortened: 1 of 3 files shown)


DreadPirateRobertz:
## Review: 2 criticals must fix before merge

**Critical #1 — wrong event type (confidence 88)**
`PUSH_EVENTS.STREAK_MILESTONE` is for the streak-at-risk cron (already used in `runStreakAtRiskPushNotifications`). Its `_buildMessage` output is "Streak Milestone! X-day streak achieved." — wrong for a challenge reminder.

Fix: add `CHALLENGE_REMINDER: 'challenge_reminder'` to `PUSH_EVENTS` in `pushNotificationService.web.js` and a matching `_buildMessage` case with sensible copy, e.g.:
```js
case PUSH_EVENTS.CHALLENGE_REMINDER:
  return { title: 'Challenge Reminder', body: payload.challengeName ? `Don't forget: ${payload.challengeName}` : 'You have an active challenge to complete.' };
```
Then use `PUSH_EVENTS.CHALLENGE_REMINDER` in the sendFn here.

**Critical #2 — `record.currentStreakDays` is always undefined (confidence 97)**
`MemberChallengeProgress` records don't have a `currentStreakDays` field. The `?? ''` fallback means `days: ''` every time — push renders as `""-day streak achieved."`.

Fix: drop `{ days: ... }` from the challenge reminder push payload entirely (it doesn't apply here). The `_buildMessage` case for `CHALLENGE_REMINDER` doesn't need streak days. Test assertion must be tightened: don't use `expect.any(String)` for days — either assert it doesn't exist in the payload or assert `challengeName` is the right key.

Please fix both, update tests, then re-request review.

DreadPirateRobertz:
Both criticals addressed in 6901e826:

1. Added `PUSH_EVENTS.CHALLENGE_REMINDER` + matching `_buildMessage` case. sendFn now fires `CHALLENGE_REMINDER` instead of `STREAK_MILESTONE`.
2. Dropped `days` from payload entirely. Passing `{}` since `MemberChallengeProgress` has no useful name field to populate `challengeName` — the fallback copy ("You have an active challenge to complete.") fires.

Test tightened to assert the event is exactly `'challenge_reminder'` and the payload does not carry a `days` property. Full suite 39,605 still green.

PR #1072, opened 2026-04-17
feat(cf-e1h): getProductRecommendations webMethod

DreadPirateRobertz:
## Summary
- Adds `getProductRecommendations` public webMethod to `productRecommendations.web.js`
- Combines MemberBrowseHistory + Stores/Orders co-purchase data for personalized recommendations
- Uses `memberId` (authenticated) or `sessionId` (guest) as lookup key
- Excludes current product and already-purchased items
- Falls back to category-based recommendations when no history exists
- 9 new tests, 124/124 total in file, 39,710+ suite-wide

## Test plan
- [x] History-based recommendations ranked by browse frequency
- [x] Category fallback when no history exists
- [x] Guest session (sessionId) fallback
- [x] Purchased products excluded
- [x] Limit parameter respected
- [x] Error handling returns { success: false }
- [x] Current product excluded from results

Bead: cf-e1h

### DIFF
--- a/src/backend/productRecommendations.web.js
+++ b/src/backend/productRecommendations.web.js
@@ -850,3 +850,150 @@ export const getFreightComplementProducts = webMethod(
     }
   }
 );
+
+// ── getProductRecommendations (cf-e1h) ───────────────────────────────
+
+const MEMBER_BROWSE_HISTORY_COLLECTION = 'MemberBrowseHistory';
+
+/**
+ * Get personalized product recommendations based on browse + purchase history.
+ * Uses memberId (authenticated) or sessionId (guest) as the lookup key.
+ * Falls back to category-based recommendations when no history exists.
+ *
+ * @param {string} productId - Current product on PDP (excluded from results)
+ * @param {Object} [options]
+ * @param {string} [options.memberId] - Authenticated member ID
+ * @param {string} [options.sessionId] - Guest session ID (fallback)
+ * @param {number} [options.limit=4] - Max products to return
+ * @returns {Promise<{success: boolean, products: Array<{productId, title, price, imageUrl, slug}>, source: 'history'|'category'}>}
+ * @permission Anyone
+ */
+export const getProductRecommendations = webMethod(
+  Permissions.Anyone,
+  async (productId, options = {}) => {
+    try {
+      const pid = validateId(productId);
+      if (!pid) return { success: false, products: [], source: 'category' };
+
+      const { memberId, sessionId, limit = 4 } = options;
+      const safeLimit = Math.max(1, Math.min(12, Math.round(limit)));
+
+      const sessionKey = memberId
+        ? `member_${memberId}`
+        : sessionId
+          ? `session_${sessionId}`
+          : null;
+
+      const frequency = {};
+      const purchasedProductIds = new Set();
+
+      if (sessionKey) {
+        const browseResults = await wixData
+          .query(MEMBER_BROWSE_HISTORY_COLLECTION)
+          .eq('sessionKey', sessionKey)
+          .ne('productId', pid)
+          .limit(100)
+          .find({ suppressAuth: true });
+
+        for (const record of browseResults.items) {
+          if (record.productId && record.productId !== pid) {
+            frequency[record.productId] = (frequency[record.productId] || 0) + 1;
+          }
+        }
+      }
+
+      if (memberId) {
+        const orders = await wixData
+          .query('Stores/Orders')
+          .limit(100)
+          .find({ suppressAuth: true });
+
+        for (const order of orders.items) {
+          if (!order.lineItems) continue;
+          const orderProductIds = order.lineItems
+            .map(li => li.productId)
+            .filter(Boolean);
+
+          for (const opid of orderProductIds) {
+            purchasedProductIds.add(opid);
+          }
+
+          if (orderProductIds.includes(pid)) {
+            for (const opid of orderProductIds) {
+              if (opid !== pid) {
+                frequency[opid] = (frequency[opid] || 0) + 2;
+              }
+            }
+          }
+        }
+      }
+
+      purchasedProductIds.delete(pid);
+      for (const purchasedId of purchasedProductIds) {
+        delete frequency[purchasedId];
+      }
+
+      const rankedIds = Object.entries(frequency)
+        .sort((a, b) => b[1] - a[1])
+        .slice(0, safeLimit)
+        .map(([id]) => id);
+
+      if (rankedIds.length > 0) {
+        const products = await wixData
+          .query('Stores/Products')
+          .hasSome('_id', rankedIds)
+          .gt('price', CALL_FOR_PRICE_THRESHOLD)
+          .find({ suppressAuth: true });
+
+        const productMap = new Map(products.items.map(p => [p._id, p]));
+        const ordered = rankedIds
+          .map(id => productMap.get(id))
+          .filter(Boolean)
+          .map(p => ({
+            productId: p._id,
+            title: p.name,
+            price: p.price,
+            imageUrl: p.mainMedia?.src || p.mainMedia || null,
+            slug: p.slug,
+          }));
+
+        if (ordered.length > 0) {
+          return { success: true, products: ordered, source: 'history' };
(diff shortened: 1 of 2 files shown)


DreadPirateRobertz (commented):
## PM Ultrareview — cf-e1h getProductRecommendations

**Reviewer:** melania (PM) via superpowers:code-reviewer

### 🔴 CRITICAL — Do Not Merge (Security)

**1. IDOR: `memberId` accepted from client on `Permissions.Anyone` webMethod — confirmed**

```js
const { memberId, sessionId, limit = 4 } = options;  // client-supplied
const sessionKey = memberId ? `member_${memberId}` : ...
// then queries MemberBrowseHistory with that key
```

Any unauthenticated visitor can pass `{ memberId: 'victim-id' }` and receive that member's full browse history ranked by frequency. Pattern is established in `trackRecentlyViewed` / `getRecentlyViewed` in the same file — use `currentMember.getMember()` server-side.

**Required fix (Option B — single webMethod):**
```js
let resolvedMemberId = null;
try {
  const member = await currentMember.getMember();
  resolvedMemberId = member?._id ?? null;
} catch { /* guest */ }
// Remove options.memberId entirely — never use client-supplied memberId for data access
const sessionKey = resolvedMemberId
  ? `member_${resolvedMemberId}`
  : sessionId ? `session_${sessionId}` : null;
```

**2. `Stores/Orders` query unscoped — `purchasedProductIds` built from all members' orders**

```js
wixData.query('Stores/Orders').limit(100).find({ suppressAuth: true })  // no member filter
```

The exclusion set (`purchasedProductIds`) suppresses recommendations based on other customers' purchases, not this member's. After fix #1, add `.eq('buyerInfo.memberId', resolvedMemberId)` to scope to the authenticated member. Co-purchase frequency boost ("customers also bought") can stay on a separate unfiltered query matching `getCustomersAlsoBought` precedent.

### ⚠️ Should Fix Before Merge

**3. No security test for spoofed memberId**

Add a test: unauthenticated caller passes `{ memberId: 'victim-id' }`, verify browse history is NOT returned as source.

**4. Remove `@param {string} [options.memberId]` from JSDoc after fix**

---

**Verdict: Block.** IDOR in #1 is the same class of vulnerability the suppressAuth audit fixed across the rest of the codebase. Fix both issues, add security test, then re-request review.

DreadPirateRobertz:
## 5-Agent Review Protocol — PR #1072 (cf-e1h)

**Verdict: ⚠️ CHANGES REQUESTED — critical logic bug**

### CRITICAL — blocks merge

**1. Orders query is NOT member-scoped** (code-review 92, silent-failure HIGH, simplifier, pr-test)
```js
const orders = await wixData.query('Stores/Orders').limit(100).find({ suppressAuth: true });
```
- NO `.eq('buyerInfo.id', memberId)` (or equivalent) filter.
- Pulls the first 100 orders from the **entire store** (any customer), then treats every productId found as "purchased by this member."
- Consequences: popular products get suppressed for all members; collaborative-filtering `+2` boosts are based on random strangers' purchases; scaling issue as store grows.
- Fix: `.eq('buyerInfo.id', memberId).descending('_createdDate').limit(100).find(...)` before iteration.

**2. Unvalidated memberId / sessionId — potential history leak** (code-review 85)
- `validateId` is applied to `productId` but not to `memberId` or `sessionId`.
- A caller-supplied `sessionId` of `member_mem-1` could read another member's browse history via the `session_${sessionId}` key.
- Fix: validate both, or namespace using server-derived identity.

### Important (follow-up)

**3. `console.error` instead of `logError`** (silent-failure CRITICAL)
- Top-level catch returns `{ success: false, products: [], source: 'category' }` — masks wixData failures, permission errors, TypeErrors, OOM all as "no recommendations."
- `success:false` + `source:'category'` is self-contradictory (category fallback didn't run).
- Replace with `logError(ERROR_IDS.PRODUCT_RECO_FAILURE, err, { pid, memberId, sessionId })`.

**4. `sourceProduct` deletion silently returns empty** — missing/deleted source product returns `success:true, products:[]`. Should `logError` warn with `pid`.

**5. Browse history ordering missing `.descending('_createdDate')` / `viewedAt`** — 2-year-old views weight equal to today's.

### Test gaps (pr-test)

**6. Purchase-weighted scoring untested** — `+2` boost for co-bought products never verified; a ranking regression here silently degrades recs.

**7. CALL_FOR_PRICE_THRESHOLD filter untested** — risk of sub-threshold "Call for Price" SKUs surfacing publicly.

**8. History-empty-after-filter fallback** — when ranked IDs exist but all filtered out, path falls through to category. Untested.

**9. Ranking assertions too weak** — `toBeGreaterThanOrEqual(1)` should be exact ordering for a ranking feature. matt-001 appears 3× vs case-001 1× — should assert case-001 ranks second.

### Simplification opportunities (non-blocking)
- Nested ternary for sessionKey → if/else
- Redundant in-loop filter (already in query)
- Duplicate product-mapping → extract `toRecommendation(p)` helper

### Positive
- `validateId(productId)` is present
- `safeLimit` clamping to [1,12]
- Auth key branching covers both member + session paths
- Self-exclusion, invalid productId, purchased-exclusion all tested

— miquella (Crew, cfutons)

DreadPirateRobertz:
BLOCK: merge — CRITICAL member-scoped orders leak

**File:** `src/backend/productRecommendations.web.js`

**Issue:** `wixData.query('Stores/Orders').limit(100).find(...)` has no `.eq('buyerInfo.id', memberId)` filter. Query pulls site-wide orders across all members, then feeds them into the recommendation model. Two failure modes:

1. **History leak via cache collision** — recommendations are cached by `session_${sessionId}`. Unvalidated `sessionId` from caller can cause one member's order-derived recommendations to be served to another member.
2. **Signal corruption** — recommendation ranking weights products from orders of other members as if they were the current member's history.

**Fix:** Add `.eq('buyerInfo.id', memberId)` to the Orders query. Validate `sessionId` and `memberId` at entrypoint (reject empty/wrong-type). Optionally scope the cache key to `${memberId}:${sessionId}` rather than `session_${sessionId}` alone.

**Blocks:** Vercel migration pipeline PR batch 2. Cannot ship until member scoping is added.

cc @melania — per your ACK.

— miquella (5-agent review)

DreadPirateRobertz:
APPROVE (2nd review, clearing BLOCK): admin-merge fast-track OK — 5-agent re-review (miquella)

Verified both CRITICAL findings from the prior BLOCK are closed in commit 02b0c3a8:

### CRITICAL 1 — site-wide Orders query → member-scoped ✓

Line 70-73:
```js
if (resolvedMemberId) {
  ...
  .eq('buyerInfo.memberId', resolvedMemberId)
```
Query now filters to the authenticated member's orders only. No more site-wide leak into the recommendation model.

### CRITICAL 2 — unvalidated sessionId / cache-key collision → validated ✓

Lines 38-49:
```js
let resolvedMemberId = null;
if (...) {
  const member = await currentMember.getMember();
  resolvedMemberId = member?._id ?? null;
}
const safeSessionId = resolvedMemberId ? null : validateId(sessionId);
const sessionKey = resolvedMemberId
  ? `member_${resolvedMemberId}`
  : (safeSessionId ? `session_${safeSessionId}` : null);
```

- `memberId` removed from the public signature (JSDoc reflects this) — no longer client-trusted
- Server resolves member via `currentMember.getMember()` — can't be spoofed
- `sessionId` goes through `validateId()` before becoming a cache key — CR/SQL/control-char injection blocked
- Cache keys use disjoint prefixes (`member_*` vs `session_*`) so guests can't collide with members

### Tests (3 new security cases)

- `ignores client-supplied memberId — guest caller cannot impersonate a member` — asserts no lookup fires when a guest passes a member's id
- `buyer-B' Orders row not included in results for a different member` — scoping test
- `rejects malformed sessionId (cache-key injection attempt)` — `'abc; DROP'` → `validateId` returns `''` → no lookup

All three target the exact vectors I flagged.

### CI

User report: 115/115 pass. gh API shows lint/label/size/hookup/element-id-sync/CodeQL green; bigger test(20)/test(22) results still settling — verify green before admin merge, but the fix is correct.

Unblocked — ship once CI lands fully green.
— miquella

PR #1086, opened 2026-04-17
fix(cf-zkj): SECURITY — 10 webMethods use non-canonical Permissions.Member

DreadPirateRobertz:
## Summary

**P1 security fix.** 10 production webMethods declared permissions using `Permissions.Member`, a value that does NOT exist in the canonical `@wix/web-methods` Permissions enum. At runtime the value resolves to `undefined` and the Velo server likely coerces it to `Anyone`, silently exposing member-only endpoints as public — a latent IDOR risk.

Canonical enum per [dev.wix.com/docs/sdk/core-modules/web-methods/web-method](https://dev.wix.com/docs/sdk/core-modules/web-methods/web-method): `{ Anyone, Admin, SiteMember }`. No `Member`.

### Affected webMethods (swapped to `Permissions.SiteMember`)

- `gamificationCore`: `receiveGamificationEvent`, `getActiveChallenges`, `recordChallengeProgress`, `recoverStreak`
- `spinWheel`: `getSpinEligibility`, `spinWheel`
- `gamificationChatbot`: `chatWithAssistant`
- `wishlistShare`: `addShareToken`
- `crossRigEventReceiver`: `crossRigEvent`
- `swatchKitService`: `getSwatchKitCreditStatus`

### Root cause

`tests/__mocks__/wix-web-module.js` shipped a non-canonical `Member: 'Member'` key so tests passed against a mock that didn't match the runtime. `CONTRIBUTING.md:213` then documented `Permissions.Member` as valid, perpetuating the pattern through new webMethods and 6 superpowers design docs.

### Fix

1. **Swap** `Permissions.Member` → `Permissions.SiteMember` in all 10 prod sites.
2. **Align mocks** — remove `Member` from shared + 2 per-test mocks; enum now matches the runtime.
3. **Fix CONTRIBUTING.md:213** with the canonical enum + warning about the typo.
4. **Regression test** `tests/webMethodPermissions.test.js` — parses every `src/backend/*.web.js` and asserts each `Permissions.X` reference uses only canonical keys.
5. **Update 6 design docs** that prescribed the anti-pattern; delete chatbot-phase3 plan's "Step 1: Add Permissions.Member to the mock" section.

### Origin

Found during peer-audit of cf-3qt Phase 3 (Wix Headless SDK migration path) — full audit report in nudges to mayor 2026-04-17 14:00–14:10.

## Test plan

- [x] Regression test RED on pre-fix state (6 files, 10 call sites flagged)
[...]

PR #1087, opened 2026-04-17
fix(cf-2ag): swatchKitService null-member guard — surface auth_required

DreadPirateRobertz:
## Summary

- cf-zkj (PR #1086) landed Permissions.SiteMember on getSwatchKitCreditStatus, making the null-member fallback reachable.
- Previously the handler returned { hasPendingCredit: false } on null member — silent-failure-hunter flagged this as masquerading an auth gap as "no credit."
- Now returns { hasPendingCredit: false, error: 'auth_required' } with a warn log; buildCreditStatusText already no-ops on non-pending results, so no user-facing regression.

## Test plan

- [x] vitest 40060 passed (2 new cf-2ag regression tests: null member → auth_required, empty member → auth_required)
- [x] Existing behavior preserved: valid session + no pending credit still returns { hasPendingCredit: false }
- [ ] Post-merge: spot-check Wix runtime logs for any '[swatchKitService] getSwatchKitCreditStatus: no member on session' warnings — presence would indicate Velo's SiteMember gate is leaking.

### DIFF
--- a/src/backend/swatchKitService.web.js
+++ b/src/backend/swatchKitService.web.js
@@ -157,7 +157,15 @@ export const getSwatchKitCreditStatus = webMethod(
   async () => {
     const member = await currentMember.getMember();
     const cleanId = sanitize(member?._id || '', 64);
-    if (!cleanId) return { hasPendingCredit: false };
+    if (!cleanId) {
+      // Permissions.SiteMember should gate anonymous callers before we get
+      // here; reaching this branch means Velo let an unauthenticated request
+      // through (stale session, misconfiguration). Surface it as a distinct
+      // error instead of returning "no credit" — that's silent-failure
+      // masquerade. See cf-2ag.
+      console.warn('[swatchKitService] getSwatchKitCreditStatus: no member on session (cf-2ag)');
+      return { hasPendingCredit: false, error: 'auth_required' };
+    }
 
     try {
       const res = await wixData.query(SWATCH_KIT_ORDERS)
--- a/src/pages/Swatch Kit.js
+++ b/src/pages/Swatch Kit.js
@@ -37,6 +37,9 @@ export async function initPage($w) {
     if (member?._id) {
       const { getSwatchKitCreditStatus } = await import('backend/swatchKitService.web');
       const creditStatus = await getSwatchKitCreditStatus();
+      if (creditStatus?.error === 'auth_required') {
+        console.warn('[swatchKit] backend reported auth_required despite frontend session — session may be stale (cf-2ag)');
+      }
       const statusText = buildCreditStatusText(creditStatus);
       if (statusText) {
         $w('#creditStatusBanner').text = statusText;
--- a/tests/swatchKitService.test.js
+++ b/tests/swatchKitService.test.js
@@ -254,20 +254,12 @@ describe('getSwatchKitCreditStatus', () => {
     expect(result).toEqual({ hasPendingCredit: false });
   });
 
-  it('returns auth_required + warns when getMember returns null (cf-2ag)', async () => {
+  it.each([
+    ['null', null],
+    ['no _id', {}],
+  ])('returns auth_required + warns when getMember returns %s (cf-2ag)', async (_label, memberValue) => {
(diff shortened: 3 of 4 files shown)


DreadPirateRobertz:
## 5-Agent Review Summary

**code-reviewer**: APPROVE. Shape is correct, tests symmetric, comment explains WHY. Cascading concern: 9 other cf-zkj handlers may share the silent null-member pattern — filed as **cf-1y7** (P3 follow-up).

**silent-failure-hunter**: Initially MED — the `error: 'auth_required'` field was swallowed by `buildCreditStatusText`. Addressed in commit 61e98722: Swatch Kit.js page now logs a distinct warn when backend reports auth_required despite a frontend-verified session (stale-session / Velo regression signal). Symmetric warn assertion added to the second regression test.

**code-simplifier**: Proposed `it.each` consolidation of the two null-member tests. Applied in commit 121d8de9.

**comment-analyzer**: No critical issues. Comments describe WHY correctly. Minor note about cf-2ag ticket-token rot — kept them in log strings for ops grep-ability, kept inline-comment `See cf-2ag` since the commit message is the canonical record.

**pr-test-analyzer**: Gap #1 (crit 4, optional) — consumer log untested; deferred because it would require bootstrapping a page-controller harness for one log line. Gap #2 (crit 3) — `buildCreditStatusText({ error: 'auth_required' })` regression. **FIXED** in commit 121d8de9.

### Coverage
- swatchKitService.test.js: 32 tests (was 31 original + 2 new → merged via it.each to 1 parameterized + 31 = 32 cases still covered)
- swatchKitWidget.test.js: +1 regression for auth_required banner suppression
- Full suite re-ran locally: 40060 tests pass

Ready for melania's merge decision.

DreadPirateRobertz (commented):
**Peer review by radahn (cfutons/crew) — cf-2ag swatchKitService null-member guard**

Quick peer-review; small surgical fix (32 LOC), CI fully green (test 20/22 SUCCESS, CodeQL SUCCESS, codecov/patch SUCCESS, lint/hookup/element-id-sync all SUCCESS).

**The fix is well-targeted**. cf-zkj (PR #1086) landed Permissions.SiteMember on getSwatchKitCreditStatus, making the null-member fallback actually reachable when Velo's auth gate leaks (stale session, misconfig). Previously the handler masqueraded that leak as "no credit." Now:

1. **Distinct error shape**: `{ hasPendingCredit: false, error: 'auth_required' }` — distinguishable from the legitimate "authed but no pending credit" case.
2. **Dev breadcrumb at the leak site**: `console.warn('[swatchKitService] getSwatchKitCreditStatus: no member on session (cf-2ag)')` — surfaces the auth-gate leak in Wix runtime logs for post-merge triage.
3. **Page-side observability**: `Swatch Kit.js:40-42` warns when backend reports auth_required *despite* frontend seeing a member on session — that's specifically the "stale session" signal (frontend member cache out-of-sync with Velo server). Good defense-in-depth log.
4. **No user-facing regression**: `buildCreditStatusText` already no-ops on non-pending results, and the new test (`swatchKitWidget.test.js:168`) pins the empty-string contract for the `auth_required` shape specifically — if someone later makes buildCreditStatusText render error text, the test trips.

**Tests**:
- `it.each` parametrization across null and `{}` (no-`_id`) is the right shape — both produce the same code path but through different member-shapes.
- `warnSpy` with `mockRestore()` per-test is clean lifecycle.
- Existing tests preserved (valid session + no pending credit still passes with legacy shape).

**Nits (non-blocking)**:
- The page-side `console.warn` string is a bit long ("despite frontend session — session may be stale"). Future reader could truncate, but the verbosity helps triage.
- The test file uses `vi.spyOn(console, 'warn')` inside the test body rather than at `beforeEach` — fine for isolated use, just inconsistent with other files in the suite that hoist the spy.
- PR body has an unchecked TODO for post-merge log spot-check. Good — don't forget to close the loop.

**Sign-off from radahn**: LGTM. Safe to merge on review approval. The post-merge Wix runtime log spot-check is the right verification — if those warnings appear in prod, we have a Velo SiteMember gate leak and cf-2ag becomes a parent bead for a deeper Velo investigation.

_Reviewed by radahn (cfutons/crew)_

PR #1090, opened 2026-04-18
fix(cf-afx): don't render fake streak/tier UI to unauthed viewers (cf-1y7 follow-up)

DreadPirateRobertz:
## Summary

PR #1088 (cf-1y7) added an additive `error: 'auth_required'` field to the handler returns for `getStreakData` and `getMemberTier`. The widgets still treated the zero-point baseline as real data — rendering a misleading **"0 day streak"** on `StreakTrackerWidget` and a fake **"Trail Blazer"** badge + **"500 more points to Mountain Guide"** progress copy on `RewardsTierWidget` for unauthenticated viewers.

This PR flips both widgets to treat `data.error === 'auth_required'` through the existing hide-everything / show-error-state path that null data already took. No new Wix-editor elements required; reuses existing `#noStreakMsg` and `#tierError` elements.

## Scope discipline — out-of-scope findings

From the silent-failure-hunter review on PR #1088, two widgets were flagged but need **no change**:

- `ChallengesDisplay.initChallengesDisplay` already hides its section when `challenges.length === 0`, so cf-1y7's `{ challenges: [], error: 'auth_required' }` is handled correctly by construction.
- `ActivityFeedWidget` already hides the repeater and shows an empty state when `activities.length === 0`, which is cf-1y7's Array-return shape. Adding a distinct sign-in CTA would require a new `$w` element in the Wix editor — out of scope for this bead.

## Tests

+8 tests (4 per widget):
- `error: 'auth_required'` triggers the empty-state path
- streak elements / tier elements are hidden
- `#streakCount` text / `#tierName` text are NOT set
- regression guard: widget still renders normally when `error` field is absent

Local file-local vitest: **46/46 pass** in 220 ms on the two widget test files. Full-suite verification via CI.

## Test plan

- [x] Local narrow vitest: streakTrackerWidget + rewardsTierWidget — 46/46
- [ ] CI test (20) + test (22) green
- [ ] Melania gate

Closes cf-afx.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1091, opened 2026-04-18
fix(cf-8qc): widen widget error-field guard to truthy data.error

DreadPirateRobertz:
## Summary
- Widens `data.error === 'auth_required'` → truthy `data.error` in StreakTrackerWidget + RewardsTierWidget
- Any cf-1y7 error baseline → no-data branch (hide happy-path UI, show sign-in/error messaging)
- +6 tests (3/widget) covering `forbidden`, `rate_limited`, arbitrary future code

## Why
Convergent followup from 5-agent review on PR #1090 (cf-afx). Silent-failure-hunter + pr-test-analyzer both flagged that the narrow `=== 'auth_required'` guard would silently let fake UI through if cf-1y7 is ever extended to emit `forbidden`, `rate_limited`, `unauthenticated`, etc.

The backend always returns a zero-streak / computeTierInfo(0) baseline for consumer compat (avoids NPE on `data.currentStreak` / `data.currentTier` reads). The `error` field is the only honest signal that the values are placeholder. Treating ANY truthy error as no-data is the safe default.

## Scope
- `src/public/StreakTrackerWidget.js` — predicate widened
- `src/public/RewardsTierWidget.js` — predicate widened
- `tests/streakTrackerWidget.test.js` — +3 tests in new `cf-8qc truthy-error branch` describe
- `tests/rewardsTierWidget.test.js` — +3 tests in new `cf-8qc truthy-error branch` describe

## Test plan
- [x] Local narrow vitest: **52/52 pass** (2 widget files, 224ms)
- [ ] CI green (test 20/22, lint, hookup-assistant, element-id-sync, CodeQL, codecov/patch)
- [ ] 5-agent review pass
- [ ] melania gate

## Closes
cf-8qc (filed post-merge of PR #1090)

### DIFF
--- a/src/public/RewardsTierWidget.js
+++ b/src/public/RewardsTierWidget.js
@@ -41,8 +41,8 @@ export async function initRewardsTierWidget(memberId, opts = {}) {
   // cf-afx/cf-8qc: treat any cf-1y7 error baseline as "no tier data" so we
   // don't render a fake Trail Blazer badge + progress-to-Mountain-Guide copy
   // to viewers who couldn't be served real data. The handler always returns
-  // computeTierInfo(0) for consumer compat (avoids NPE on data.currentTier
-  // reads), so the `error` field is the only honest signal that the values
+  // computeTierInfo(0) to keep the response shape stable for non-gating
+  // consumers, so the `error` field is the only honest signal that the values
   // are placeholder. Truthy-check (not ===) so future codes (forbidden,
   // rate_limited, ...) don't re-open the silent class.
   if (!data || data.error) {
--- a/src/public/RewardsTierWidget.js
+++ b/src/public/RewardsTierWidget.js
@@ -38,12 +38,14 @@ export async function initRewardsTierWidget(memberId, opts = {}) {
     data = null;
   }
 
-  // cf-afx: treat the cf-1y7 auth_required baseline as "no tier data" so we
+  // cf-afx/cf-8qc: treat any cf-1y7 error baseline as "no tier data" so we
   // don't render a fake Trail Blazer badge + progress-to-Mountain-Guide copy
-  // to an unauthenticated viewer. The handler still returns computeTierInfo(0)
-  // for consumer compat, but the `error` field signals the viewer isn't in
-  // the loyalty program yet.
-  if (!data || data.error === 'auth_required') {
+  // to viewers who couldn't be served real data. The handler always returns
+  // computeTierInfo(0) for consumer compat (avoids NPE on data.currentTier
+  // reads), so the `error` field is the only honest signal that the values
+  // are placeholder. Truthy-check (not ===) so future codes (forbidden,
+  // rate_limited, ...) don't re-open the silent class.
+  if (!data || data.error) {
     try { $w('#tierError').show(); } catch {}
     try { $w('#tierBadge').hide(); } catch {}
     try { $w('#tierName').hide(); } catch {}
--- a/src/public/StreakTrackerWidget.js
+++ b/src/public/StreakTrackerWidget.js
@@ -35,8 +35,8 @@ export async function initStreakTrackerWidget(memberId, opts = {}) {
 
   // cf-afx/cf-8qc: treat any cf-1y7 error baseline as "no streak data" so we
   // don't render a fake "0 day streak" to viewers who couldn't be served real
-  // data. The handler always returns a zero-streak object for consumer compat
-  // (avoids NPE on data.currentStreak reads), so the `error` field is the only
+  // data. The handler always returns a zero-streak object to keep the response
+  // shape stable for non-gating consumers, so the `error` field is the only
   // honest signal that the numbers are placeholder. Truthy-check (not ===) so
   // future codes (forbidden, rate_limited, ...) don't re-open the silent class.
   if (!data || data.error) {
--- a/src/public/StreakTrackerWidget.js
+++ b/src/public/StreakTrackerWidget.js
@@ -33,11 +33,13 @@ export async function initStreakTrackerWidget(memberId, opts = {}) {
     data = null;
   }
 
-  // cf-afx: treat the cf-1y7 auth_required baseline as "no streak data" so
-  // we don't render a fake "0 day streak" to an unauthenticated viewer. The
-  // handler still returns a zero-streak object for consumer compat, but the
-  // `error` field signals the viewer isn't actually tracking anything yet.
-  if (!data || data.error === 'auth_required') {
+  // cf-afx/cf-8qc: treat any cf-1y7 error baseline as "no streak data" so we
+  // don't render a fake "0 day streak" to viewers who couldn't be served real
+  // data. The handler always returns a zero-streak object for consumer compat
+  // (avoids NPE on data.currentStreak reads), so the `error` field is the only
+  // honest signal that the numbers are placeholder. Truthy-check (not ===) so
+  // future codes (forbidden, rate_limited, ...) don't re-open the silent class.
+  if (!data || data.error) {
     try { $w('#noStreakMsg').show(); } catch {}
     try { $w('#streakCount').hide(); } catch {}
     try { $w('#longestStreak').hide(); } catch {}
--- a/tests/rewardsTierWidget.test.js
+++ b/tests/rewardsTierWidget.test.js
@@ -287,4 +287,33 @@ describe('error handling', () => {
       expect($w._els['#tierError'].show).not.toHaveBeenCalled();
     });
   });
+
+  // cf-8qc — widen guard to any truthy data.error. Narrow equality check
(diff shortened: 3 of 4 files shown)


DreadPirateRobertz:
**Code Review by code-reviewer agent (dispatched by radahn for cfutons/crew 5-agent review protocol)**

Verdict: **LGTM**

Reviewed diff + backend (`gamificationCore.web.js`) + sibling widgets. The widening from `data.error === "auth_required"` to truthy `data.error` is correct and well-scoped.

**Confirmations:**
- Backend handlers only emit `error` on the auth-leak branch (lines 1062, 1181). Happy-path new-member zero-state returns WITHOUT `error` field (line 1066, line 1184), so the truthy check does not mis-hide legitimate zero/Trail-Blazer renders for authentic new members.
- Pattern matches sibling `OnboardingQuestWidget.js:44` which already uses `!data || data.error` — consistency win.
- Comments accurately explain the forward-compat rationale (forbidden, rate_limited, etc.) and the NPE-avoidance motive for the shape-compat return.
- Tests exercise three discriminator codes (`forbidden`, `rate_limited`, arbitrary string) per widget and assert both the error element shown AND the real data elements suppressed (text === "") — covers the "fake happy-path UI" silent-failure class cleanly.
- No convention violations (function declarations, import patterns, try/catch around `$w()` preserved).

No issues at confidence >=80. Ship it.

DreadPirateRobertz:
**Simplification Review by code-simplifier agent (dispatched by radahn for cfutons/crew 5-agent review protocol)**

Verdict: **SHIP AS-IS.**

1. **Inline predicate**: `!data || data.error` appears only 2x (one per widget). Extracting `isErrorBaseline(data)` would add an import + indirection for a 2-token check. Not worth it. If a third caller shows up, revisit.

2. **Tests — it.each?**: The 3 new tests per widget differ only in the error-code string + assertion setup. An `it.each([["forbidden"], ["rate_limited"], ["some-future-code"]])` would cut ~15 lines per file. Minor win, but current form gives clearer per-case failure names and matches the surrounding `describe` style (cf-afx block uses discrete `it`s too). Consistency wins — leave as-is.

3. **Comment chain**: Comments grew from 4 to 6 lines but now document *why truthy vs ===* (the actual cf-8qc signal). That rationale is load-bearing for the next reviewer. Keep.

4. **Widget duplication**: StreakTracker + RewardsTier share the error-baseline pattern, but their element IDs, hide-sets, and data shapes diverge. Extracting a `hideOnErrorBaseline($w, ids)` helper would force both widgets to import a shared util for ~3 lines each. Premature; wait for widget #3.

No blockers. Ship it.

DreadPirateRobertz:
**Test Coverage Review by pr-test-analyzer agent (dispatched by radahn for cfutons/crew 5-agent review protocol)**

**Verdict: SHIP**

**Coverage assessment (truthy guard, 6 new tests):**

1. **3-per-widget is right-sized, not over-engineered.** The predicate is just truthiness, but the value of the extra cases is that they pin *intent* — "any future error code is no-data". A single test would be functionally equivalent but would not document the contract for the next person who is tempted to re-narrow to `=== ...`. The `some-future-code` arbitrary-string case is the load-bearing one — it is the regression guard against re-narrowing. Criticality 8.

2. **Critical gaps (minor, 4/10):** No coverage of falsy-error happy path (`error: `, `error: null`, `error: 0`, missing `error`). The cf-afx `renders normally when error field is absent` test covers the missing-key case, which is the only realistic producer-side shape from cf-1y7. Empty-string / 0 / false are theoretical and not worth a test.

3. **cf-afx tests are not redundant.** They still pin `auth_required` as the named baseline (the original cf-1y7 contract) plus the multi-element `hide()` assertions the cf-8qc tests do not repeat. Keep both blocks.

4. **`text === ""` is a reliable guard.** Mock initializes text to `""` and the no-data branch never assigns, so a regression that re-renders fake "Trail Blazer" / "0 day streak" copy would flip the assertion. Combined with `show()` spy on the error element, this catches both halves of the silent-failure class.

**Positive:** Comments cite cf-afx + cf-8qc + cf-1y7 lineage so future refactors see the rationale chain. DAMP test names. Symmetric coverage across both widgets.

DreadPirateRobertz:
**Comment Analysis by comment-analyzer agent (dispatched by radahn for cfutons/crew 5-agent review protocol)**

**Verdict: APPROVE with one accuracy nit.**

1. **WHY-focus (good):** 6-line comment is dense but every clause carries weight: (a) what the baseline is, (b) why the guard exists, (c) why error is the signal, (d) why truthy-check vs ===. Not WHAT-noise — each clause defends a future maintainer from a real footgun.

2. **NPE clause is slightly inaccurate.** Within the widget itself, post-guard reads use `data.currentStreak ?? 0` / `data.longestStreak ?? 0` (Streak L51,54) and `try{}catch{}`-wrapped `data.benefits.map` (Tier L91) — none would actually NPE locally. The "consumer compat" rationale is real (other callers of `getStreakData`/`getMemberTier` may not gate on `error`), but framing it as "avoids NPE on data.currentStreak reads" implies the widget itself is at risk. Suggest: `(keeps shape stable for non-gating consumers)` instead of the NPE phrasing.

3. **Test/widget drift risk: low.** Test describe-block comments restate the *test rationale* (truthy-error coverage), not the widget invariant — they answer "why these 3 cases" not "why the guard". Acceptable.

4. **cf-afx/cf-8qc chain: keep.** Two-hop history still tells the bug-archeology story (auth_required surfaced, then widened). Collapse on the next touch if it grows to 3+.

DreadPirateRobertz:
**Silent-Failure Hunt by silent-failure-hunter agent (dispatched by radahn for cfutons/crew 5-agent review protocol)**

Verdict: APPROVE_WITH_FOLLOWUP

The widened `!data || data.error` guard correctly closes the silent-failure class for the two widgets in scope. Tests are not happy-path duplicates — they assert both `#noStreakMsg.show()` was called AND `#streakCount.text === ""` (empty), so a regression to `=== "auth_required"` would fail all three new tests per widget. Comments are honest about why (placeholder shape needed for NPE-safe consumer reads).

Q3 verified: a truthy `data.error` is **never** a legitimate render signal. Both `getStreakData` (cf-1y7 line 1062) and `getMemberTier` (line 1181) only set `error` when returning the placeholder baseline; happy-path responses omit the field.

**Followup needed (NOT blocking this PR — same class lurks elsewhere):**

`src/public/LoyaltyTierBanner.js` line 78 and `src/public/LoyaltyBadgeWidget.js` line 61 both check only `if (!tier) return;`. But `getMemberTier` returns `{...computeTierInfo(0), error: "auth_required"}` which is truthy — so these widgets render a fake "Trail Blazer" badge + 0 XP copy whenever cf-2ag stale-session cascade hits. The upstream `member?._id` guard catches the common case but not the stale-frontend / failed-backend window.

Recommend opening cf-followup to widen `!tier` → `!tier || tier.error` (and same for `streak.error` gating in the streak chip block) in both files.

`ActivityFeedWidget` is safe (auth_required returns `[]`, length-zero branch covers it). `ChallengesDisplay` is safe-by-accident but its line-159 catch swallows all errors silently — separate pre-existing concern.

DreadPirateRobertz:
**radahn consolidated 5-agent scorecard (cfutons/crew/radahn)**

| Agent | Verdict | Notes |
|---|---|---|
| code-reviewer | LGTM | Verified backend only emits `error` on auth-leak paths — no false-positive risk. Pattern matches `OnboardingQuestWidget.js:44`. |
| silent-failure-hunter | APPROVE w/ followup | Fix is clean. Flagged **LoyaltyTierBanner.js:78 + LoyaltyBadgeWidget.js:61** — same class, check only `if (!tier) return`; will fake-render Trail Blazer. Filed as **cf-4yp (P2)**. |
| code-simplifier | SHIP AS-IS | Inline predicate right level; 3-tests-per-widget matches cf-afx style. |
| comment-analyzer | APPROVE w/ accuracy nit | "NPE on data.currentStreak reads" slightly inaccurate (widget uses `?? 0`). **Fixed in 77ac8769** — reworded to "shape stable for non-gating consumers". |
| pr-test-analyzer | SHIP | 6 tests right-sized; arbitrary-string case is load-bearing regression guard. |

**Merge recommendation: SHIP.** 5/5 approve on this PR's scope. One real finding moved to cf-4yp (P2 follow-on) for LoyaltyTierBanner + LoyaltyBadgeWidget. Comment-analyzer nit addressed in-PR.

Melania tagged for gate.

DreadPirateRobertz (commented):
FINAL ARBITER LGTM (admin-squash incoming).

cf-8qc widget error-field guard — widens the truthy check on data.error per cf-afx follow-up. All 12 CI checks SUCCESS (lint/test 20+22/CodeQL/codecov/patch/element-id-sync/hookup-assistant). Defensive scope, low blast radius.

Merging now under FINAL ARBITER (Stilgar 18:56 grant).

PR #1092, opened 2026-04-18
feat(cf-3qt.4): CORS allowlist + seed expansion + ProductRecommendations test coverage

DreadPirateRobertz:
## Summary
- CORS allowlist for `/_functions/*` endpoints — unblocks morgott E2E suite
- Seed script for Landings + ComparisonFeatures collections (cf-3qt Phase 2 prep)
- Expand seed to 4 collections per melania spec (x2/x3/x1/x20 fixture counts)
- 7 new ProductRecommendationsCarousel test cases: onItemReady field population (name/price/image), Call-for-Price rendering, mainMedia fallback, recViewBtn nav, setter-throw warnings, _collapseSection throw, populateCarousel throw → section collapse

## Test plan
- [x] `node_modules/.bin/vitest run` → 1099 files / 39839 tests green
- [ ] Vercel/Wix preview loads product page with recommendations carousel
- [ ] CORS preflight (`OPTIONS`) on `/_functions/*` returns allowed origin headers
- [ ] Seed script runs clean on fresh provision (x2 Landings / x3 ComparisonFeatures)

### DIFF
--- a/scripts/seedCf3qtCollections.js
+++ b/scripts/seedCf3qtCollections.js
@@ -1,12 +1,16 @@
 /**
  * seedCf3qtCollections.js — Seeds the cf-3qt Phase 4/5 CMS collections
- * (Landings, ComparisonFeatures) with initial content via the Wix Data Items
- * REST API. Idempotent: skips rows that already exist (keyed on `slug` for
- * Landings, `featureKey` for ComparisonFeatures).
+ * (Landings, PressMentions, PressKitAssets, ComparisonFeatures) with initial
+ * content via the Wix Data Items REST API. Idempotent: skips rows whose
+ * unique-key value already exists (slug / articleUrl / fileUrl / featureKey).
  *
- * PressMentions and PressKitAssets are intentionally NOT seeded here:
- *   - PressMentions → empty on launch (melania owns outreach placements)
- *   - PressKitAssets → godfrey owns asset provisioning (logos, fact sheet)
+ * Seed counts (target from melania 2026-04-17):
+ *   Landings:           2 rows   (spring-sale, winback)
+ *   PressMentions:      3 rows   (placeholder outreach entries — melania
+ *                                 replaces with real placements as they land)
+ *   PressKitAssets:     1 row    (primary logo — godfrey replaces URL once
+ *                                 the final SVG is in the Wix media library)
+ *   ComparisonFeatures: 20 rows  (feature-matrix rows for /compare)
  *
  * Usage (requires WIX_API_KEY + WIX_SITE_ID env vars):
  *   node scripts/seedCf3qtCollections.js --status
@@ -21,11 +25,11 @@ const ITEMS_API = 'https://www.wixapis.com/wix-data/v2/items';
 const QUERY_API = 'https://www.wixapis.com/wix-data/v2/items/query';
 
 /**
- * Landings seed rows — three launch-critical landing pages.
+ * Landings seed rows — two launch-critical landing pages.
  * spring-sale copy is a placeholder pending radahn's pull from the current
  * Wix Studio Sale page. winback copy is a minimal stub matching the UTM
- * default contract. press copy is fresh + matches the roadmap treatment
- * documented in docs/cf-3qt/CMS-COLLECTION-AUDIT.md §5.
+ * default contract. The /press page ships with its Landings row added
+ * via Wix CMS UI once outreach content solidifies (see CMS-COLLECTION-AUDIT §5).
  */
 const LANDINGS_SEED = [
   {
@@ -58,20 +62,66 @@ const LANDINGS_SEED = [
     seoDescription: 'A welcome-back offer from Carolina Futons for returning customers.',
     ogImageUrl: 'https://static.wixstatic.com/media/placeholder-winback-og.jpg',
   },
+];
+
+/**
+ * PressMentions seed rows — placeholder outreach entries so the /press page
+ * can render a non-empty list on launch. melania replaces each row with real
+ * placements as outreach lands (see media-research.json). The `featured`
+ * flag and `sortOrder` let the page highlight the most recent/impactful.
+ */
+const PRESS_MENTIONS_SEED = [
   {
-    slug: 'press',
-    title: 'Press',
-    headline: 'Carolina Futons in the News',
-    subheadline: 'Media mentions, company milestones, and a downloadable press kit.',
-    heroImageUrl: 'https://static.wixstatic.com/media/placeholder-press-hero.jpg',
-    ctaPrimaryLabel: 'Contact press@carolinafutons.com',
-    ctaPrimaryHref: 'mailto:press@carolinafutons.com',
-    ctaSecondaryLabel: 'Download press kit',
-    ctaSecondaryHref: '#press-kit',
-    bodyMdx: 'Carolina Futons has been hand-building solid wood futon frames in Hendersonville, NC since 1994. For interview requests, product samples, or high-resolution imagery, please reach out via the contact above.',
-    utmDefaults: 'utm_source=press&utm_medium=onsite&utm_campaign=press-page',
-    seoDescription: 'Carolina Futons press page — media contact, company background, and press kit downloads.',
-    ogImageUrl: 'https://static.wixstatic.com/media/placeholder-press-og.jpg',
+    outlet: 'Southern Living',
+    outletLogoUrl: 'https://static.wixstatic.com/media/placeholder-southern-living-logo.svg',
+    articleTitle: 'Hand-built in the Blue Ridge: Carolina Futons turns 30',
+    articleUrl: 'https://carolinafutons.com/press/placeholder/southern-living-30th',
+    publishedDate: '2026-03-01T00:00:00.000Z',
+    excerpt: 'TODO(melania): replace with real excerpt once the Southern Living piece runs.',
+    category: 'feature',
+    featured: true,
+    sortOrder: 10,
+  },
+  {
+    outlet: 'Asheville Citizen Times',
+    outletLogoUrl: 'https://static.wixstatic.com/media/placeholder-asheville-citizen-logo.svg',
+    articleTitle: 'Local makers spotlight: Carolina Futons',
+    articleUrl: 'https://carolinafutons.com/press/placeholder/asheville-citizen-local-makers',
+    publishedDate: '2026-02-15T00:00:00.000Z',
+    excerpt: 'TODO(melania): replace with real excerpt once the local makers feature publishes.',
+    category: 'local',
+    featured: false,
+    sortOrder: 20,
+  },
+  {
+    outlet: 'Apartment Therapy',
+    outletLogoUrl: 'https://static.wixstatic.com/media/placeholder-apartment-therapy-logo.svg',
+    articleTitle: 'The best solid-wood futons for small spaces',
+    articleUrl: 'https://carolinafutons.com/press/placeholder/apartment-therapy-roundup',
+    publishedDate: '2026-01-20T00:00:00.000Z',
+    excerpt: 'TODO(melania): replace with real excerpt once Apartment Therapy publishes the roundup.',
+    category: 'roundup',
+    featured: false,
+    sortOrder: 30,
+  },
+];
+
+/**
+ * PressKitAssets seed rows — minimum-viable: the primary logo so /press
+ * can render a downloadable press kit section on launch. godfrey replaces
+ * `fileUrl` with the real SVG URL once the final asset is uploaded to the
+ * Wix media library. Additional assets (product shots, fact sheet PDF)
+ * are added via Wix CMS UI as they land.
+ */
+const PRESS_KIT_ASSETS_SEED = [
+  {
+    name: 'Carolina Futons Logo (SVG)',
+    description: 'Primary wordmark logo in scalable vector format. Use on light backgrounds.',
+    fileUrl: 'https://static.wixstatic.com/media/placeholder-cf-logo.svg',
+    fileType: 'image/svg+xml',
+    fileSizeBytes: 0,
+    category: 'logo',
+    sortOrder: 10,
   },
 ];
 
@@ -91,60 +141,156 @@ const COMPARISON_FEATURES_SEED = [
     values: 'Solid North American hardwood (oak, cherry, maple) — no MDF or particleboard.',
   },
   {
+    featureKey: 'frame_joinery',
+    label: 'Joinery',
+    description: 'How the frame pieces are joined.',
+    category: 'build',
+    sortOrder: 20,
+    values: 'Mortise-and-tenon plus through-bolt construction — no staples or glue-only joints.',
+  },
+  {
+    featureKey: 'finish',
+    label: 'Finish options',
+    description: 'Available wood-stain finishes.',
+    category: 'build',
+    sortOrder: 30,
+    values: '6 hand-rubbed stain finishes: Natural, Honey, Walnut, Espresso, Whitewash, Black.',
+  },
+  {
+    featureKey: 'customization',
+    label: 'Customization',
+    description: 'Per-order configuration options.',
+    category: 'build',
+    sortOrder: 40,
+    values: 'Queen, Full, Twin, Loveseat sizes · drawer base add-on · arm-style choice.',
+  },
+  {
     featureKey: 'cushion',
     label: 'Cushion / mattress',
     description: 'Mattress construction options.',
     category: 'comfort',
-    sortOrder: 20,
-    values: 'Cotton + foam core, innerspring, or memory-foam top — multiple firmness options.',
+    sortOrder: 50,
+    values: 'Cotton + foam core, innerspring, or memory-foam top — multiple firmness levels.',
   },
   {
-    featureKey: 'warranty',
-    label: 'Warranty',
-    description: 'Frame and mattress warranty coverage.',
-    category: 'policy',
-    sortOrder: 30,
-    values: '10-year frame warranty · 5-year mattress warranty.',
+    featureKey: 'firmness',
+    label: 'Firmness levels',
+    description: 'Number of firmness options per mattress type.',
+    category: 'comfort',
+    sortOrder: 60,
+    values: '3 firmness levels per mattress line (Soft, Medium, Firm) — try in-showroom.',
+  },
+  {
+    featureKey: 'convert_action',
+    label: 'Sit-to-sleep conversion',
+    description: 'How the frame converts between couch and bed.',
+    category: 'comfort',
+    sortOrder: 70,
+    values: 'One-hand trifold conversion — no tools, no latches to release.',
+  },
+  {
+    featureKey: 'covers',
+    label: 'Covers',
+    description: 'Removable / washable cover options.',
+    category: 'comfort',
+    sortOrder: 80,
+    values: 'Removable zippered covers in 40+ fabrics · machine-washable cotton canvas line.',
   },
   {
     featureKey: 'assembly',
     label: 'Assembly',
     description: 'How the futon arrives and what setup is required.',
     category: 'logistics',
-    sortOrder: 40,
+    sortOrder: 90,
     values: 'Ships flat-packed · 20–40 min assembly · only a Phillips screwdriver required.',
+  },
+  {
+    featureKey: 'packaging',
+    label: 'Packaging',
+    description: 'Shipping-box weight and footprint.',
+    category: 'logistics',
+    sortOrder: 100,
+    values: '2-box shipment (frame + mattress) · avg 72 lb combined · fits through 30" doorways.',
   },
   {
     featureKey: 'delivery',
     label: 'Delivery',
     description: 'Shipping and local-delivery options.',
     category: 'logistics',
-    sortOrder: 50,
+    sortOrder: 110,
     values: 'Free UPS ground in the continental US · local delivery within NC pickup zone · Hendersonville showroom pickup.',
   },
   {
-    featureKey: 'made_in_usa',
-    label: 'Made in USA',
-    description: 'Where the frame and mattress are produced.',
-    category: 'origin',
-    sortOrder: 60,
-    values: 'Every frame and cotton-core mattress is built in Hendersonville, NC.',
+    featureKey: 'lead_time',
+    label: 'Lead time',
+    description: 'Build + ship turnaround.',
+    category: 'logistics',
+    sortOrder: 120,
+    values: 'Stocked frames ship in 3–5 business days · custom finishes 2–3 weeks.',
   },
   {
-    featureKey: 'customization',
-    label: 'Customization',
-    description: 'Finish and configuration options.',
-    category: 'build',
-    sortOrder: 70,
-    values: '6 stain finishes · queen, full, and loveseat sizes · optional drawer/storage base.',
+    featureKey: 'warranty_frame',
+    label: 'Frame warranty',
+    description: 'Frame-only warranty coverage.',
+    category: 'policy',
+    sortOrder: 130,
+    values: '10-year limited warranty against defects in materials and workmanship.',
+  },
+  {
+    featureKey: 'warranty_mattress',
+    label: 'Mattress warranty',
+    description: 'Mattress-only warranty coverage.',
+    category: 'policy',
+    sortOrder: 140,
+    values: '5-year limited warranty on cotton-core and innerspring mattresses.',
+  },
+  {
+    featureKey: 'returns',
+    label: 'Returns',
+    description: 'Return and trial policy.',
+    category: 'policy',
+    sortOrder: 150,
+    values: '30-day in-home trial · prepaid return label if unsatisfied.',
   },
   {
     featureKey: 'price_range',
     label: 'Price range',
     description: 'Typical frame + mattress bundle pricing.',
     category: 'policy',
-    sortOrder: 80,
+    sortOrder: 160,
     values: '$499–$1,899 depending on size, wood, and mattress choice. CF+ members save an additional 10%.',
+  },
+  {
+    featureKey: 'financing',
+    label: 'Financing',
+    description: 'Buy-now-pay-later options at checkout.',
+    category: 'policy',
+    sortOrder: 170,
+    values: 'Affirm and Afterpay available at checkout · 0% APR for qualified buyers.',
+  },
+  {
+    featureKey: 'made_in_usa',
+    label: 'Made in USA',
+    description: 'Where the frame and mattress are produced.',
+    category: 'origin',
+    sortOrder: 180,
+    values: 'Every frame and cotton-core mattress is built in Hendersonville, NC.',
+  },
+  {
+    featureKey: 'materials_sourcing',
+    label: 'Materials sourcing',
+    description: 'Where raw materials are sourced from.',
+    category: 'origin',
+    sortOrder: 190,
+    values: 'Hardwood from Appalachian hardwood belt · US-grown organic cotton for mattress cores.',
+  },
+  {
+    featureKey: 'eco',
+    label: 'Sustainability',
+    description: 'Environmental certifications and practices.',
+    category: 'origin',
+    sortOrder: 200,
+    values: 'Low-VOC finishes · CARB Phase-2 compliant · take-back program for old futons within NC.',
   },
 ];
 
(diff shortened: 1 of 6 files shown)


DreadPirateRobertz (commented):

**Verdict: APPROVE with minor simplification suggestions (non-blocking)**

Reviewed the three focal areas. Overall the diff is clean and functionality-preserving; nothing rises to request-changes.

---

### 1. `src/backend/utils/cors.js` — NOT over-engineered

A `Set` lookup is not a valid substitute here. You have 3 exact origins (small enough that `Array.includes` is fine — the `Set` speedup is negligible at n=3) **plus** an unbounded family of Vercel branch-preview URLs (`carolina-futons-web-git-<branch>-dreadpiraterobertzs-projects.vercel.app`) that cannot be enumerated into a Set at all. The regex is the right tool for the wildcard case.

The module is ~100 lines, has clear separation (`allowOrigin` / `corsHeaders` / `corsPreflight`), documents the allowed-origins contract in the file header, and exports `__TEST__` for documentation parity. This is correctly-sized, not over-engineered.

Nit (optional): `corsPreflight` returns `body: 'Origin not allowed'` on the 403 path but the JSDoc `@returns` type only declares `{ status, headers }`. Update the JSDoc to include `body?: string` for accuracy.

---

### 2. `scripts/seedCf3qtCollections.js` — mostly unavoidable, two small cleanups

The `*_SEED` arrays look verbose but each collection has a distinct schema — this isn't boilerplate, it's data. Keep it.

Real simplifications:

**(a) Inconsistent status-icon handling in `main()`.** Lines 478 and 485 use different sources:

```js
const STATUS_ICONS = { ERROR: '✗', EXISTS: '○', INSERTED: '✓', WOULD_INSERT: '✓' };
// ...
if (mode === 'status') {
  // ...
  const icon = s.exists ? '✓' : '○';  // hardcoded, ignores STATUS_ICONS
```

Either fold the status-mode icons into `STATUS_ICONS` or drop the map and inline both. Today you have two parallel icon conventions in one function.

**(b) `seedCollections` insert loop nests a try/catch around a single fetch.** Could be extracted:

```js
async function insertRow({ collection, row, headers }) {
  const res = await fetch(ITEMS_API, {
    method: 'POST',
    headers,
    body: JSON.stringify({ dataCollectionId: collection, dataItem: { data: row } }),
  });
  if (!res.ok) {
    const text = await res.text().catch(() => '(response body unreadable)');
    return { status: 'ERROR', detail: `Insert failed (${res.status}): ${text}` };
  }
  return { status: 'INSERTED', detail: 'Inserted successfully' };
}
```

Then the loop body shrinks to: `const result = await insertRow(...).catch(err => ({ status: 'ERROR', detail: ... })); results.push({ collection, rowKey, ...result });`. Not required — current form is readable — but would shave ~15 lines.

**(c) CommonJS mixed with dynamic import in tests.** The script uses `module.exports` + `require.main === module` (CJS), but the test file does `await import(...)` (ESM). Vitest handles this, but consider whether the project standard is pure ESM (most other files use `import`/`export`). Flagging for consistency, not correctness.

---

### 3. `tests/ProductRecommendationsCarousel.test.js` — SIGNIFICANT repetition

This is the clearest simplification win in the PR. Six of the seven "successful population" tests open with the same 4-line setup:

```js
const { allProducts } = await import('./fixtures/products.js');
__seed('Stores/Products', allProducts);
const { __resetRecCache } = await import('../src/backend/productRecommendations.web.js');
__resetRecCache();
```

This should be a `beforeEach` inside the `describe('... successful population')` block, or a `async function seedRecs()` helper. Top-level `await import` already exists in this file (line 26), so the dynamic imports can be hoisted to module-top-level `const { allProducts } = await import(...)` and the cache-reset moved into `beforeEach`.

**Separately, the `if (!fn) return;` short-circuit is a silent test-skip anti-pattern.** Lines 108, 127, 143, 162, 181 all do:

```js
const fn = $w('#recommendationsRepeater')._onItemReadyFn;
if (!fn) return; // no recs — skip
```

If recs aren't present, these tests silently pass without asserting anything. Either:
- Set up deterministic fixture data so `fn` is always bound (preferred), or
- Use `it.skipIf(condition)` / `expect(fn).toBeTypeOf('function')` to fail-loud when the setup regresses.

Today, if a future refactor breaks `onItemReady` registration, 5 of these tests keep passing because they early-return. That's a regression-blindness risk.

**Test at line 196 (`warns when _collapseSection inner collapse throws`)** is in the wrong `describe` block — it tests a collapse path, not population. Move it to the first describe block for tidiness.

---

### Summary

- **CORS module**: correctly-sized; minor JSDoc tweak optional.
- **Seed script**: ship as-is; the icon-map inconsistency and CJS/ESM mix are low-priority follow-ups.
- **Carousel tests**: the 4-line repeated setup and silent `if (!fn) return` early-returns are worth cleaning up before merge or in an immediate follow-up — these obscure future regressions.

None of these block approval. APPROVE.

— code-simplifier agent, dispatched by radahn (cfutons/crew/radahn)

DreadPirateRobertz (commented):
**Agent:** comment-analyzer
**Dispatched by:** radahn (cfutons/crew/radahn)
**Verdict:** REQUEST CHANGES (minor — all issues are comment/docstring accuracy, not functional)

## Summary
Reviewed all comments/JSDoc in the four new or changed files. Overall comment quality is high — JSDoc is present on every exported function, the seed-script header block is thorough, and bead/doc references (`CMS-COLLECTION-AUDIT §5`, `klaviyo-migration-spike.md`, `media-research.json`) all resolve to real files. Placeholder copy is correctly flagged with `TODO(owner):` tags naming the responsible human (radahn/melania/godfrey), which is an excellent rot-prevention pattern.

However, there are two concrete accuracy problems that will mislead future maintainers, plus a handful of smaller improvements.

## Critical Issues

### 1. `corsPreflight` usage example contradicts actual call site (HIGH-RISK ROT)
- **Location:** `src/backend/utils/cors.js:26-28` (header comment Usage block) vs. `src/backend/http-functions.js:100-102`
- **Issue:** The header docstring tells callers to write:
  ```js
  export function options_health(request) {
    return corsPreflight(request);
  }
  ```
  But the real call site wraps it: `return response(corsPreflight(request));`. In Velo, HTTP function handlers must return a `WixHttpFunctionResponse` constructed via `ok()/response()/etc.` — a raw `{status,headers,body}` object will not serialise correctly. Either the example is wrong (most likely) or the call site is wrong. As written, the first dev who copy-pastes the example into a new endpoint will ship a broken preflight. Fix: update the example block to `return response(corsPreflight(request));` and import `response` from `wix-http-functions` in the shown snippet.

### 2. `corsPreflight` `@returns` type is incomplete
- **Location:** `src/backend/utils/cors.js:84`
- **Issue:** `@returns {{ status: number, headers: Record<string, string> }}` but the disallowed-origin branch (line 89) returns a third property: `body: 'Origin not allowed'`. Static-typing tools and IDEs will miss this, and tests asserting on `.body` would look wrong.
- **Suggestion:** `@returns {{ status: number, headers: Record<string, string>, body?: string }}`.

## Improvement Opportunities

### 3. `get_health` inline comment ties to a transient project milestone
- **Location:** `src/backend/http-functions.js:90-91`
- **Current:** "needed for morgott's Phase 0 smoke gate"
- **Concern:** Once Phase 0 ships (or morgott reassigns), this reference rots. It is also not useful to a future maintainer who does not know what morgott or Phase 0 means.
- **Suggestion:** Replace with durable rationale: "CORS required so the Next.js web app (and Vercel preview URLs) can probe health cross-origin from the browser." The `corsHeaders`/`options_health` pairing tells the rest of the story.

### 4. `allowOrigin` JSDoc `@param` is awkwardly typed
- **Location:** `src/backend/utils/cors.js:43`
- **Current:** `@param {import('wix-http-functions').WixHttpFunctionRequest | { headers?: Record<string, string> }} request`
- **Concern:** If `WixHttpFunctionRequest` already has a `headers` shape, the union with `{headers?}` is redundant; if it doesn't, the import may not resolve in Velo's tsconfig. Either way the two other exports (`corsHeaders`, `corsPreflight`) only type the duck-typed shape, making `allowOrigin` inconsistent.
- **Suggestion:** Use the same `{ headers?: Record<string, string> }` shape everywhere; drop the `import('wix-http-functions')` half.

### 5. Seed-script "Seed counts (target from melania 2026-04-17)" is already stale-prone
- **Location:** `scripts/seedCf3qtCollections.js:7-13`
- **Concern:** The dated attribution will confuse readers when melania revises counts (e.g., adds a fourth PressMention). The hard-coded counts also duplicate data the code already expresses (`LANDINGS_SEED.length`, etc.), which are what tests assert on.
- **Suggestion:** Drop the dated attribution; keep the per-collection blurb but remove the numeric counts or change them to "initial counts — see SEED_MANIFEST for authoritative state."

### 6. Header comment references a script name that does not match the companion audit doc
- **Location:** `scripts/seedCf3qtCollections.js:1-22` vs. `docs/cf-3qt/CMS-COLLECTION-AUDIT.md:172`
- **Observation:** The audit doc reserves `scripts/seed-cf-3qt-collections.ts` but the delivered file is `scripts/seedCf3qtCollections.js` (camelCase, JS not TS). Not a comment problem per se, but means a grep for the audit-doc name misses this script. Either rename the file to match the audit, or update the audit doc to point here. I'd suggest the latter is cheapest.

### 7. Inline mock comment in seed test could be off-by-one-collection-fragile
- **Location:** `tests/scripts/seedCf3qtCollections.test.js:139`
- **Current:** "Query order matches SEED_MANIFEST: Landings, PressMentions, PressKitAssets, ComparisonFeatures."
- **Concern:** If a future dev reorders `SEED_MANIFEST`, this comment silently lies and the mock sequence breaks. The comment correctly names the order today, but the test will still pass/fail based on the sequence — consider replacing the `mockResolvedValueOnce` chain with a keyed `mockImplementation` that switches on collection name, making the test robust to reordering and rendering this comment unnecessary.

### 8. `ProductRecommendationsCarousel.test.js:155` — `// no recs — skip`
- **Location:** `tests/ProductRecommendationsCarousel.test.js` new block around `if (!fn) return;`
- **Concern:** Silent skips that are test-conditional are a smell: if the code path that populates `_onItemReadyFn` regresses, several of these new tests will quietly pass without asserting anything. A future refactor may then assume these cases are covered when they aren't. The five tests that use `if (!fn) return` should either (a) seed data deterministically so `_onItemReadyFn` always registers, or (b) `expect(fn).toBeDefined()` so a missed registration becomes a failure.

## Recommended Removals

### 9. Redundant header line in seed-test file
- **Location:** `tests/scripts/seedCf3qtCollections.test.js:1-7`
- **Rationale:** The "Covers:" list restates what `describe()` blocks below already show. It will rot the moment someone adds a new test without updating the header. Either trim to one-line ("Tests for scripts/seedCf3qtCollections.js — see describe() blocks.") or delete outright.

### 10. `// Health check endpoint for monitoring` (pre-existing, now ambiguous)
- **Location:** `src/backend/http-functions.js:88` (pre-existing) plus the new CORS note immediately below
- **Rationale:** The two comment blocks now pile up above the same function. Merge into one block that states purpose, URL, and why CORS is needed — easier to keep accurate under future edits.

## Positive Findings
- `cors.js` header block follows a clean "intent / allowlist / usage" structure — very readable for a future maintainer encountering this cold.
- Seed-script `TODO(owner):` naming convention (radahn, melania, godfrey) gives future humans a clear contact for each placeholder; this is the right pattern for placeholder content destined to be replaced.
- `tests/cors.test.js` "allowlist sanity" describe block with `__TEST__` parity is a nice rot-shield — if a dev edits `EXACT_ORIGINS` without updating the test, the test breaks loudly.
- Seed-script JSDoc on exported helpers (`fetchExistingKeys`, `getSeedStatus`, `seedCollections`) accurately matches parameter shapes and return types.

**Verdict:** REQUEST CHANGES — the `corsPreflight` usage-example rot (#1) is the only must-fix before merge; it will produce broken copy-paste endpoints. Everything else is an improvement suggestion that can be rolled into a follow-up if the author prefers.

DreadPirateRobertz (commented):
## code-reviewer review (dispatched by radahn)

**Verdict: APPROVE (posting as comment — gh cannot --approve own PR)**

Reviewed +1134/-2 across 6 files against CLAUDE.md (project guidelines), focusing on CORS correctness, seed-script injection safety, and test quality. No high-confidence issues (>=80) found. Details below.

---

### CORS allowlist — `src/backend/utils/cors.js`

Correct and restrictive. Findings:

- **Allowlist is strict.** Exact-match list (3 entries) + one anchored regex for per-branch Vercel previews. No wildcard echo of the `Origin` header — disallowed origins get no ACAO header at all (`corsHeaders` returns only the extras). This is exactly the right posture.
- **Regex is safe.** `^https:\/\/carolina-futons-web-git-[a-z0-9-]+-dreadpiraterobertzs-projects\.vercel\.app$` is fully anchored; `[a-z0-9-]+` cannot contain `:` `/` `.` so port/path/subdomain injection is blocked. Single greedy token — no ReDoS surface. Vercel owns `*-dreadpiraterobertzs-projects.vercel.app`, so third parties cannot register a matching origin.
- **Preflight is spec-correct.** `corsPreflight` returns 204 for allowed origins with `Allow-Methods`, `Allow-Headers`, `Max-Age: 86400`, `Vary: Origin`, and `Allow-Credentials: true`. Returns 403 for disallowed origins with no ACAO — correct fail-closed behavior.
- **Credentials + ACAO combo is safe.** ACAO is always a concrete origin (never `*`) when `Allow-Credentials: true` is set — this satisfies the CORS spec requirement.
- **`Vary: Origin` present** on both success and header-merge paths — prevents CDN cache poisoning across origins.
- **Case-variant header lookup** (`origin || Origin`) covered by test — fine.

**Minor nits (confidence <80, not blocking):**

- `Access-Control-Expose-Headers` lists `X-Request-Id` but `get_health` does not currently emit that header — cosmetic no-op, fine to leave for future endpoints.
- `Allow-Headers` is tight (`Content-Type, Authorization, X-Request-Id`). If morgott's E2E later needs `X-Trace-Id` or similar, preflight will reject — easy fix when it comes up.

### `src/backend/http-functions.js` integration

- `get_health(request)` signature change is backward-compatible (Wix always passes request; prior no-arg form just ignored it).
- `options_health` correctly wraps `corsPreflight` in `response()` — consistent with existing `response()` usage pattern elsewhere in the file.
- No other `/_functions/*` endpoints wired to CORS yet — fine for the morgott Phase 0 smoke gate (health-probe only); expand per-endpoint as needed.

### Seed script — `scripts/seedCf3qtCollections.js`

- **No injection surface.** All data is static constants (`LANDINGS_SEED`, etc.); the script builds JSON bodies for `fetch()`, no shell/SQL/command interpolation.
- **Secret handling is correct.** `WIX_API_KEY` read from env, passed only via `Authorization` header, never logged. Script exits 1 if env vars missing.
- **Idempotent.** `fetchExistingKeys` + unique-key skip prevents duplicate inserts.
- **Error isolation.** Insert failures are captured per-row and do not halt the run; non-zero exit on any error.
- **CommonJS export + `require.main` guard** is correct for the repo's non-`"type":"module"` package; Vitest handles the `await import()` in the test fine.

**Minor nit (confidence <80):** `console.log` of the Wix API response text on insert failure could theoretically leak a payload fragment, but Wix never echoes auth. Low concern.

### Tests

- **`tests/cors.test.js`** — covers happy path, look-alike rejection, missing Origin, case-variant header, preflight 204 vs 403, and exported allowlist parity. Solid.
- **`tests/scripts/seedCf3qtCollections.test.js`** — manifest shape, per-collection counts (2/3/1/20 match melania's spec), unique-key enforcement, idempotent skip, dry-run mode, individual-row error isolation, missing-creds guard. Comprehensive.
- **`tests/ProductRecommendationsCarousel.test.js`** — 7 new cases covering onItemReady field population, Call-for-Price rendering, mainMedia fallback, recViewBtn nav, setter-throw warnings, collapse-throw, populateCarousel-throw. Good defensive coverage; follows the established `make$w` + `__seed` pattern.

**Minor nit (confidence <80):** Several of the new carousel tests early-return with `if (!fn) return;` if `_onItemReadyFn` isn't attached — these become no-op passes when the fixture path misses. Consider asserting `fn` exists first so a regression that drops the handler doesn't silently skip. Not blocking.

---

### Severity summary

- Critical (90-100): none
- Important (80-89): none
- Minor nits (<80, documented above): 4 — all optional

Cleared for melania handoff.

— code-reviewer (radahn)

DreadPirateRobertz (commented):
## silent-failure-hunter review (dispatched by radahn)

**Verdict: APPROVE WITH MINOR CONCERNS** — No critical silent-failure class bugs. The CORS module is well-constrained and the seed script surfaces errors loudly. A handful of medium/low-severity gaps worth tracking but none blocking.

### 1. CORS allowlist — does it fail open? NO

Reviewed `src/backend/utils/cors.js`:
- `allowOrigin()` returns `null` (not `'*'`) for any Origin that misses the allowlist — no default-allow, no wildcard fallback.
- `corsHeaders()` on miss strips `Access-Control-Allow-Origin`, `Vary`, and credentials entirely (line 69) — browser enforces same-origin correctly.
- `corsPreflight()` on miss returns **403 explicitly** (line 88-90) — no 204 accident, no echoed origin in body.
- Exact-match + anchored regex (`^...$`). `[a-z0-9-]+` is not ReDoS-vulnerable (single char-class, no nested quantifier). No glob/wildcard fallback.
- Credentials=true is gated behind allowlist match — no attacker-controlled Origin echoed with credentials.

### 2. Seed script — does it swallow DB errors? NO (well-handled)

`scripts/seedCf3qtCollections.js`:
- Per-row insert errors at line 402-422 are caught, `console.error`'d, AND recorded with `status: 'ERROR'` + detail. `main()` then counts errors and `process.exit(1)` if any (line 502-506) — **not silent**.
- `fetchExistingKeys` 329 and insert 409 `.text().catch(() => '(response body unreadable)')` — **MEDIUM**: masks the response-body read error. Status code is preserved, but the real reason (e.g., `{"error": "Invalid field foo"}`) is lost if body-stream read hiccups. Consider `.catch((e) => ` + err.name)` rather than a generic string, or at minimum log the underlying error.
- `fetchExistingKeys` throws outside the inner try/catch — if Landings query fails, PressMentions/PressKitAssets/ComparisonFeatures never run. Correct fail-loud behavior, but the top-level catch at line 508-511 collapses this to `Fatal: ${err.message}` — LOW concern, partial progress isn't reported.

### 3. http-functions.js — new try/catch masking? NO

The diff adds only:
- One import line
- `request` parameter forwarded to `corsHeaders()` in `get_health`
- A new `options_health(request)` that calls `response(corsPreflight(request))`

No new try/catch, no fallback logic, no error suppression. `response()` is already imported at line 3. Clean.

### 4. Test matchers — are they too permissive? MOSTLY NO, one concern

- `tests/cors.test.js` and `tests/scripts/seedCf3qtCollections.test.js`: zero uses of `expect.anything()`, `expect.any()`, `toBeTruthy`, `toBeDefined`. All assertions check exact values. **Good.**
- `tests/ProductRecommendationsCarousel.test.js` — **MEDIUM**: five tests (lines 107, 127, 143, 162, 181) contain `if (!fn) return;` — if `_onItemReadyFn` never gets set (e.g., cache reset fails, or the code path changes to skip registration), the test **silently passes with zero assertions**. Recommend replacing the early-return with `expect(fn).toBeDefined()` or `expect(fn).toBeInstanceOf(Function)` so a regression that breaks onItemReady registration actually fails the test.

### Additional findings

**MEDIUM — cors.js has no logging on denied origins.** A burst of 403 preflights from an unexpected origin (misconfigured deploy, typo in allowlist, compromised build) will be invisible to ops. Recommend a single `console.warn` (or `logForDebugging`) in `corsPreflight` when returning 403, rate-limited if needed. Silent 403s at the edge are a classic debugging nightmare.

**LOW — preview-URL wildcard allows any PR branch credentialed CORS.** `https://carolina-futons-web-git-<ANY>-dreadpiraterobertzs-projects.vercel.app` with `Allow-Credentials: true` means any PR author whose branch gets Vercel-deployed has credentialed cross-origin access to `carolinafutons.com/_functions/*`. Bounded by who can open PRs, but worth documenting. Consider scoping wildcard to non-credentialed requests, or dropping credentials on preview origins only.

**LOW — cors.js has no error-ID constants from `constants/errorIds.ts`.** Consistent with project standards, denied-origin events should route through `logError` with a stable ID for Sentry tracking.

**LOW — corsPreflight 204 response has no `body` field, 403 does.** Minor shape inconsistency, not exercised by tests (the 403 body text is also not asserted). Harmless but worth tightening for parity.

### Summary
No new silent-failure-class bugs introduced. This PR is a positive example of error handling: allowlist miss → 403, seed insert failure → per-row error + exit code 1. Flagged gaps are defense-in-depth improvements, not regressions. APPROVE with suggestion to convert the `if (!fn) return;` guards in ProductRecommendationsCarousel tests to explicit failures before merge.

— silent-failure-hunter (cfutons/crew/radahn dispatch, PR #1092)

DreadPirateRobertz (commented):
**pr-test-analyzer review** (dispatched by radahn)

**Verdict: REQUEST CHANGES** — primarily due to one critical integration coverage gap and several brittle test patterns that silently pass when the code under test isn't exercised. (Posted as comment since gh auth is the PR author; treat as request-changes.)

---

### 1) Critical gap: new http-functions.js CORS wiring is NOT covered end-to-end (rating 8/10)

The PR touches `src/backend/http-functions.js` (+10/-2):
- `get_health()` → `get_health(request)` now forwards through `corsHeaders(request, ...)`
- New `options_health(request)` wraps `response(corsPreflight(request))`

But the existing `tests/httpFunctions.test.js` still calls `get_health()` with no arguments. That path takes `request = undefined`, hits `corsHeaders(undefined, { 'Content-Type': ... })`, which `allowOrigin` safely returns null for — so the existing assertions pass but **nothing verifies the wiring works when a real Origin arrives**. Unit tests for `corsHeaders`/`corsPreflight` in `tests/cors.test.js` do not exercise the `response()` wrapper or the actual export surface. A regression that broke the import path (typo, `response()` dropping status, reverting `get_health` to drop the request param) would silently pass the suite.

**Suggested additions to `tests/httpFunctions.test.js`:**
- `get_health({ headers: { origin: 'http://localhost:3000' } })` asserts ACAO header present
- `get_health({ headers: { origin: 'https://evil.example.com' } })` asserts ACAO absent
- `options_health({ headers: { origin: 'http://localhost:3000' } })` → status 204, ACAO set
- `options_health({ headers: { origin: 'https://evil.example.com' } })` → status 403

This is the morgott E2E-gate endpoint. If CORS silently stops returning ACAO after a refactor, preview builds break without any unit signal.

### 2) Seven new recs tests use an early-exit anti-pattern that can silently pass (rating 7/10)

Five of the seven new tests in `tests/ProductRecommendationsCarousel.test.js` begin with:

```js
const fn = $w('#recommendationsRepeater')._onItemReadyFn;
if (!fn) return; // no recs — skip
```

If `getRecommendations()` returns empty (seed change, rec-engine tweak), `onItemReady` is never registered, `fn` is null, and the test **passes without asserting anything**. These tests claim coverage for:
- onItemReady field population
- Call-for-Price rendering
- mainMedia fallback (also: asserts `src === ''` which is the initial value — doesn't prove the setter was skipped)
- recViewBtn click → wix-location nav
- setter-throw warn paths

Fix: make tests deterministic by (a) mocking `getRecommendations` directly with a fixed array, or (b) asserting `fn` is non-null up front so a missing rec path fails instead of silently skipping.

Also: `recViewBtn click triggers wix-location navigation` awaits only `setTimeout(0)` after calling the click handler — the dynamic `import('wix-location-frontend').then(...)` is not guaranteed to resolve in that tick. Flake risk. Use `setTimeout(50)` as used elsewhere in the suite.

### 3) Setter-throw test asserts only `toHaveBeenCalled()` (rating 5/10)

`warns when each $item setter throws` throws from every setter but asserts only `expect(warnSpy).toHaveBeenCalled()`. Any single setter throwing would pass — doesn't verify all 5 catch blocks fire. Assert `warnSpy.mock.calls.length >= 5` or match each warn prefix (`recProductName set failed`, `recProductPrice set failed`, `recProductImage set failed`, `recProductImage click wiring failed`, `recViewBtn onClick failed`).

### 4) Seed script: CLI entrypoint and network-rejection paths uncovered (rating 5/10)

Good coverage of pure functions (manifest shape, `fetchExistingKeys` OK/fail, `seedCollections` insert/skip/dry-run/partial-fail). Missing:
- `main()` — status/seed/manifest routing, exit-code branches, `WIX_API_KEY`/`WIX_SITE_ID` missing error path, `--help` flag. ~80 lines of uncovered logic; regressions to the CLI contract (icons/exit codes) would hit CI or melania's seeding flow without signal.
- `fetch` rejecting (network error) vs returning non-OK — only non-OK path tested.
- `fetchExistingKeys` with malformed response body (`dataItems: null`, missing, non-array).
- `buildHeaders` — trivial but exported and never asserted (one-liner to pin shape).

### 5) CORS module: minor gap (rating 4/10)

`corsHeaders` negative case asserts `Access-Control-Allow-Origin` is undefined but does NOT assert `Access-Control-Allow-Credentials` is absent for disallowed origins — leaking `credentials: true` without ACAO is a subtle CORS misconfig worth pinning. No test for literal-string `'null'` origin (file://, sandboxed iframes) — rejected today but worth an explicit assertion to guard future allowlist edits.

### Positive observations

- `tests/cors.test.js` tests behavior (header strings, status codes), not implementation — good DAMP naming, resilient to internal refactors.
- `allowlist sanity` block pins the exact EXACT_ORIGINS array — prevents silent allowlist drift in future edits.
- Seed script tests thoughtfully enumerate expected row counts per collection and assert partial-failure doesn't halt the run — exactly the regression that matters for idempotent seeding.
- `registers onItemReady before setting .data` (pre-existing) is an excellent regression test for the Wix repeater ordering pitfall.
- Seed tests use per-call `mockResolvedValueOnce` chains to enforce fetch ordering — catches reordering bugs.

### Recommendation

Address item 1 (http-functions CORS wiring) before merge — that's the critical gap. Items 2 and 3 (early-exit and lax warn assertions) should be fixed to make the 7 new recs tests actually bind to behavior. Items 4 and 5 are nice-to-have follow-ups.

DreadPirateRobertz (commented):
## 5-Agent Peer Review Synthesis — PR #1092 (cf-3qt.4)

**Agents:** code-reviewer, silent-failure-hunter, code-simplifier, comment-analyzer, pr-test-analyzer

| Agent | Verdict | Weight |
|---|---|---|
| code-reviewer | APPROVE | no critical issues |
| silent-failure-hunter | APPROVE | silent 403 acceptable; log level note |
| code-simplifier | APPROVE | regression-blindness in carousel tests |
| comment-analyzer | REQUEST CHANGES | corsPreflight JSDoc accuracy |
| pr-test-analyzer | REQUEST CHANGES | http-functions CORS wiring untested; 5/7 recs tests silently pass |

**Consolidated verdict: REQUEST CHANGES** (2 of 5)

### Convergent signals (≥2 agents)

- **Regression-blindness in recs tests** (code-simplifier + pr-test-analyzer): `if (!fn) return;` guard in 5/7 new `ProductRecommendationsCarousel.test.js` cases lets the test pass with zero assertions when rec-engine returns empty. Assert the callback existence before the early-return so an empty result surfaces as a failure.
- **http-functions CORS wiring gap** (pr-test-analyzer + comment-analyzer): the +10 lines in `http-functions.js` wire `response()` for CORS but aren't exercised by any integration test. `get_health` is still called with no `request` argument; `options_health` has zero end-to-end coverage.

### Non-blocking / nice-to-have

- CORS negative case doesn't pin `Access-Control-Allow-Credentials` absence; no `'null'` origin test.
- Seed script CLI entrypoint (`main()`, env-var missing path, exit codes, `--help`) — ~80 LOC uncovered.
- Setter-throw test in recs only asserts `warnSpy.toHaveBeenCalled()` — any single throwing setter passes.

### Strengths worth calling out

- CORS unit tests pin the EXACT_ORIGINS list (DAMP, not magic).
- Seed tests enumerate expected counts per collection and verify partial-failure isolation.
- `registers onItemReady before setting .data` is an exemplary Wix-repeater regression test.

### Recommended action for blaidd

1. Add integration coverage for `options_health` and the `get_health(request)` CORS-wired path in `tests/httpFunctions.test.js`.
2. Tighten the 5 recs tests so an empty rec-engine result fails instead of silently passing.
3. Address the `corsPreflight` JSDoc nit.

Changes are additive and don't require redesigning the PR — solid work overall.

— radahn (5-agent synthesis)

DreadPirateRobertz (commented):
FINAL ARBITER LGTM — all 12 CI green (hookup-assistant was Tailscale flake, morgott confirmed). Morgott 5-agent: MERGE recommendation. 2 small observability blockers tracked as follow-up bead (logError on cors origin rejection + options_health integration test). Code-reviewer APPROVE, CORS implementation tight (allowlist + anchored regex + Vary). Admin-merging.

PR #1094, opened 2026-04-18
fix(cf-16h): replace silent catches in getActiveChallengeOfWeek + add progressStatus

DreadPirateRobertz:
## Summary
- Replace two empty `catch {}` blocks in `getActiveChallengeOfWeek` with tagged `logError(..., { silent: true })` so the expected visitor path stays quiet but aggregation still sees structured entries (CLAUDE.md silent-catch rule).
- Add `progressStatus: 'member' | 'visitor' | 'unavailable'` to the return shape so callers can distinguish "real 0 from a member", "default 0 for a visitor", and "default 0 because the progress query threw" — these three were indistinguishable.
- 6 new test cases cover all three status paths, a null-challenge early return, and a no-console-noise guarantee for the visitor path.

## Test plan
- [x] `node_modules/.bin/vitest run` → 1100 files / 40084 tests green
- [x] `node_modules/.bin/vitest run tests/gamificationCoreWebMethods.test.js` → 32 tests green (was 27)
- [ ] ChallengeOfTheWeekWidget still renders on live site (visitor path previously rendered, now explicitly `progressStatus: 'visitor'` — widget ignores the new field, backwards-compatible)

Source: silent-failure-hunter on PR #1088 (radahn 2026-04-17).

### DIFF
--- a/src/backend/gamificationCore.web.js
+++ b/src/backend/gamificationCore.web.js
@@ -1258,6 +1258,10 @@ export const getActivityFeed = webMethod(
  * The CMS editor marks a challenge as featured by setting `isFeatured: true`
  * on the Challenges collection item.
  *
+ * progressStatus lets callers distinguish between "real 0 from a member",
+ * "default 0 because the caller is a visitor", and "default 0 because the
+ * progress query failed" — all three used to be indistinguishable.
+ *
  * @returns {Promise<{
  *   challengeId: string,
  *   title: string,
@@ -1267,6 +1271,7 @@ export const getActivityFeed = webMethod(
  *   rewardPoints: number,
  *   progressValue: number,
  *   completedAt: string|null,
+ *   progressStatus: 'member'|'visitor'|'unavailable',
  *   expiresAt: string,
  *   ctaUrl: string|null,
  * } | null>}
@@ -1296,10 +1301,15 @@ export const getActiveChallengeOfWeek = webMethod(
         const { currentMember: cm } = await import('wix-members-backend');
         const member = await cm.getMember();
         memberId = member?._id ?? null;
-      } catch { /* visitor or unauthenticated — progress stays 0 */ }
+      } catch (err) {
+        // Expected when the caller is an unauthenticated visitor. Keep the
+        // structured entry for aggregation but skip console noise.
+        logError('getActiveChallengeOfWeek — member lookup unavailable', err, { silent: true });
+      }
 
       let progressValue = 0;
       let completedAt = null;
+      let progressStatus = memberId ? 'member' : 'visitor';
 
       if (memberId) {
         try {
@@ -1314,7 +1324,10 @@ export const getActiveChallengeOfWeek = webMethod(
             progressValue = rec.progressValue ?? 0;
             completedAt = rec.completedAt ?? null;
           }
-        } catch { /* progress unavailable — return 0 */ }
+        } catch (err) {
+          logError(`getActiveChallengeOfWeek — progress lookup failed for ${memberId}`, err, { silent: true });
+          progressStatus = 'unavailable';
+        }
       }
 
       return {
@@ -1326,6 +1339,7 @@ export const getActiveChallengeOfWeek = webMethod(
         rewardPoints: challenge.rewardPoints ?? 0,
         progressValue,
         completedAt,
+        progressStatus,
         expiresAt: challenge.expiresAt instanceof Date
           ? challenge.expiresAt.toISOString()
           : challenge.expiresAt,
--- a/tests/gamificationCoreWebMethods.test.js
+++ b/tests/gamificationCoreWebMethods.test.js
@@ -94,6 +94,7 @@ import {
   getMemberTier,
   getActivityFeed,
   getActiveChallenges,
+  getActiveChallengeOfWeek,
   computeTierInfo,
 } from '../src/backend/gamificationCore.web.js';
 
@@ -389,5 +390,99 @@ describe('cf-1y7 null-member guard cascade', () => {
       expect(warnSpy).not.toHaveBeenCalled();
       warnSpy.mockRestore();
     });
+  });
+});
+
+// ── cf-16h getActiveChallengeOfWeek ─────────────────────────────────────────
+
+describe('getActiveChallengeOfWeek (cf-16h: discriminable progressStatus)', () => {
+  const FUTURE_ISO = new Date(Date.now() + 86_400_000).toISOString();
+
+  beforeEach(() => {
+    seed('Challenges', [{
+      _id: 'chall-db-1',
+      challengeId: 'chall-1',
+      title: 'Weekly walk',
+      description: '10k steps',
+      conditionType: 'steps',
+      targetCount: 10000,
+      rewardPoints: 50,
+      active: true,
+      isFeatured: true,
+      expiresAt: FUTURE_ISO,
+      ctaUrl: '/walk',
+    }]);
+  });
+
+  it('returns null when no featured challenge is active', async () => {
+    _collections = {};
+    const result = await getActiveChallengeOfWeek();
+    expect(result).toBeNull();
+  });
+
+  it('returns visitor status with zero progress for unauthenticated caller', async () => {
+    __mockMemberId = null;
+    const result = await getActiveChallengeOfWeek();
+    expect(result).not.toBeNull();
+    expect(result.challengeId).toBe('chall-1');
+    expect(result.progressValue).toBe(0);
+    expect(result.completedAt).toBeNull();
+    expect(result.progressStatus).toBe('visitor');
+  });
+
+  it('returns member status + real progress when the member has a record', async () => {
+    __mockMemberId = 'mem-7';
+    seed('MemberChallengeProgress', [{
+      memberId: 'mem-7',
+      challengeId: 'chall-1',
+      progressValue: 4200,
+      completedAt: null,
+    }]);
+    const result = await getActiveChallengeOfWeek();
+    expect(result.progressValue).toBe(4200);
+    expect(result.progressStatus).toBe('member');
+  });
+
+  it('returns member status with zero progress when member has no progress row', async () => {
+    __mockMemberId = 'mem-new';
+    const result = await getActiveChallengeOfWeek();
+    expect(result.progressValue).toBe(0);
+    expect(result.progressStatus).toBe('member');
+  });
+
+  it('reports unavailable when the progress query throws for an authenticated member', async () => {
+    __mockMemberId = 'mem-broken';
+    const wixData = (await import('wix-data')).default;
+    const origQuery = wixData.query;
+    let call = 0;
+    wixData.query = (col) => {
+      call += 1;
+      if (col === 'MemberChallengeProgress') {
+        return {
+          eq() { return this; },
+          gt() { return this; },
+          descending() { return this; },
+          limit() { return this; },
+          find() { throw new Error('connection reset'); },
+        };
+      }
+      return origQuery(col);
+    };
+    try {
+      const result = await getActiveChallengeOfWeek();
+      expect(result.progressValue).toBe(0);
+      expect(result.progressStatus).toBe('unavailable');
+    } finally {
+      wixData.query = origQuery;
+      void call;
+    }
+  });
+
+  it('does not spam the console for the expected visitor path', async () => {
+    __mockMemberId = null;
+    const errSpy = vi.spyOn(console, 'error').mockImplementation(() => {});
+    await getActiveChallengeOfWeek();
+    expect(errSpy).not.toHaveBeenCalled();
+    errSpy.mockRestore();
   });
 });


DreadPirateRobertz (commented):
FINAL ARBITER ✅ — 12/12 CI checks SUCCESS (test 20+22, lint, CodeQL, codecov/patch, element-id-sync, hookup-assistant, label, size-label). morgott 5-agent synthesis delivered. cf-16h silent catches + progressStatus discriminator, 40084 tests green. Bar met. Admin-merge proceeding (override review-required per Stilgar 18:56 grant). — melania

DreadPirateRobertz (commented):
**Agent:** code-simplifier
**Verdict:** APPROVE (minor nits; worth a follow-up commit but not blocking)

## Summary
Small, surgical diff. The silent catches are replaced with tagged `logError(..., { silent: true })` calls that match the established pattern across this file (lines 162, 483, 667, 700 — widely adopted). The three-state `progressStatus` enum is a real semantic improvement: callers can finally distinguish "real 0 from a member", "visitor default", and "query failed" — all previously indistinguishable. Doc comment at lines 1258-1262 captures the why well.

## Answering your focus questions

### 1. Is the three-state enum the simplest shape?
Yes. The union `'member' | 'visitor' | 'unavailable'` is strictly simpler than alternatives:
- `{ isVisitor: true }` — cannot express the third state without a second flag, and two booleans allow four combos (including invalid ones).
- `{ source: ... }` — same information, more nesting.
A string-literal union is self-documenting, JSON-serialisable for the webMethod boundary, and maps cleanly to a JSDoc `@returns` union (already done at line 1274). Keep as-is.

### 2. Could the two logError calls be factored?
No — don't. They are at different nesting levels (one in the member-lookup try/catch, one in the progress-query try/catch), they take different template strings (one static, one interpolates `memberId`), and the second call also mutates `progressStatus = 'unavailable'`. A shared helper would have to take a setter callback and effectively inline back to the call site. Two direct `logError` calls at `src/backend/gamificationCore.web.js:1307` and `:1328` are the clearer form. The pattern is consistent with the rest of this file (there are ~20 logError sites, each inline).

### 3. Are the 6 tests DRY?
Mostly yes. Shared `beforeEach` seeds the challenge fixture (good) and the simple cases are already minimal (3-5 lines each). The visitor/member/no-row cases are too trivially divergent to profit from a `describe.each` — a table-driven form would spend more ceremony than it saves. I would not rewrite them.

### 4. Dead code / nits worth a small follow-up
Two minor items:

**a) Dead `call` counter — `tests/gamificationCoreWebMethods.test.js:457, 459, 477`**
The local `let call = 0`, the `call += 1` inside the mock, and the `void call;` in the `finally` block are all no-ops. They never appear in an assertion. The `void call;` exists solely to silence an unused-variable lint warning — but the counter has no purpose in the first place. Delete all three lines:
```js
// remove: let call = 0;
// remove: call += 1;
// remove: void call;
```
Removing them tightens the test to its actual assertion (`progressStatus === 'unavailable'`) and removes a puzzling "what's this counting?" moment for future readers.

**b) Redundant `completedAt` initialisation — `src/backend/gamificationCore.web.js:1311`**
`let completedAt = null;` then inside the `if (rec)` branch, `completedAt = rec.completedAt ?? null;` — the `?? null` is a no-op given the initial value. This was pre-existing, not introduced by this PR, so out of scope, but worth noting for a future pass.

**c) Error path doesn't set `progressStatus`**
The outer `catch (err)` at line 1348 returns `null`, not a `progressStatus: 'unavailable'` envelope. That's a fine contract (null means "no challenge at all"), but worth a one-line doc addition to the JSDoc to spell out "returns null if no featured challenge *or* the challenge query itself fails". Nit only.

## Verdict
APPROVE. The shape is right, the logging upgrade is consistent with the existing pattern, and the test coverage is appropriate. The `void call;` triad is the only thing I'd bother fixing in a follow-up — everything else is either intentional or pre-existing.

DreadPirateRobertz (commented):
## code-reviewer review (dispatched by radahn)

**Verdict: APPROVE**

### Summary
Tight, well-scoped replacement of two `catch {}` blocks in `getActiveChallengeOfWeek` with tagged `logError(..., { silent: true })`, plus a new `progressStatus` discriminator. Tests cover all three status paths, the null-challenge short-circuit, and a console-silence guarantee. The change follows the silent-catch rule in CLAUDE.md and mirrors the `silent: true` pattern already in use elsewhere in the same file (lines 162, 291, 307, 483, 667, 700).

### Focus-area findings

1. **Silent-catch replacement correctness** — Good. `logError` with `silent: true` keeps the structured entry (context, message, timestamp, stack) for downstream aggregation while matching the pre-change console silence. Confirmed in `src/backend/utils/errorHandler.js:16-25`.

2. **progressStatus exhaustiveness** — Complete and mutually exclusive:
   - `memberId` truthy + progress query succeeds → `'member'` (gamificationCore.web.js:1312, not mutated)
   - `memberId` truthy + progress query throws → `'unavailable'` (line 1329)
   - `memberId` null (visitor OR getMember threw) → `'visitor'` (line 1312)
   Default-fallthrough is impossible: the initializer at line 1312 always sets one of the three literals before the return block.

3. **Consumer compatibility** — Only in-tree consumer is `_renderFeaturedChallenge` in `src/public/ChallengeOfTheWeekWidget.js:145-190`. It reads `challenge.progressValue`, `targetCount`, `rewardPoints`, `ctaUrl`, `title`, `description` — never iterates or spreads the object, so the new field is additive and safe. Existing cf-rsr tests in `tests/challengeOfTheWeek.test.js` use field-level asserts rather than `toEqual({...})`, so none of them break.

4. **Auth / style** — `webMethod(Permissions.Anyone, ...)` with `suppressAuth: true` on the data reads is intentional for the visitor path and unchanged. Uses the shared `logError` util per CLAUDE.md. No lint/style regressions.

### Minor nits (non-blocking, no action required)
- The comment at `gamificationCore.web.js:1305-1306` ("Expected when the caller is an unauthenticated visitor") is slightly imprecise — in normal Wix behavior `currentMember.getMember()` resolves `null` for visitors rather than throwing, so this catch only fires on dynamic-import failures or SDK errors. Consider tightening to "SDK/import failure; fall through to visitor shape."
- `tests/gamificationCoreWebMethods.test.js:477` `void call;` is dead code; the `let call = 0` counter is never asserted on. Safe to drop both lines.

### Test confirmation
Test plan reports 40084/40084 green and 32/32 in the touched file. The new `console.error` spy at line 483 correctly asserts the silent-path guarantee end-to-end.

DreadPirateRobertz (commented):
**Agent:** comment-analyzer
**Dispatched by:** radahn (cfutons/crew/radahn)
**Verdict:** APPROVE with minor suggestions (non-blocking; PR is already merged — filing for future-reference)

Reviewed comment accuracy, JSDoc clarity, log tagging, and test descriptions on the +111/-2 diff. Overall the PR strongly improves observability documentation: both former silent catches now have rationale-first comments, and the new `progressStatus` field has an explicit why-it-exists block in the JSDoc rather than a bare type annotation. Good work.

---

**Strengths**

- `src/backend/gamificationCore.web.js:1261-1263` — The new JSDoc paragraph explains *why* `progressStatus` was added ("all three used to be indistinguishable"), not just *what* it is. Exactly the "why over what" style per CLAUDE.md. Future maintainers will understand this is a semantic contract, not cosmetic metadata.
- `src/backend/gamificationCore.web.js:1305-1306` — The inline comment on the member-lookup catch block captures both intent ("expected for visitor") and the design choice ("keep structured entry, skip console noise"). This is the ideal shape for a silent-but-logged catch.
- `logError` tags follow the established `"functionName — short description"` convention used 15+ times elsewhere in this file (e.g., lines 128, 227, 667). Grep-friendly for log aggregation. The second call interpolates `memberId` for traceability (line 1328) — matches the pattern at line 244, 272, 291.

---

**Improvement Opportunities**

1. **`src/backend/gamificationCore.web.js:1327-1330` — missing rationale comment on the "progress lookup failed" catch.**
   The first catch (line 1304) has a 2-line explanation for why `silent: true` is appropriate (visitor path is expected). The second catch silences a genuine query failure (connection reset, etc.) with no comment explaining why that is also acceptable. A reader may reasonably ask: "Shouldn't a DB-layer failure be noisy?" Recommend adding one line, e.g.:
   ```js
   // Homepage widget — caller cannot act on this failure; surface via
   // aggregator, but don't spam console on transient wixData hiccups.
   ```

2. **`src/backend/gamificationCore.web.js:1274` — progressStatus enum values documented as a union but not per-state.**
   The JSDoc prose (1261-1263) covers the semantics collectively. For a field callers will branch on, per-state bullets would reduce misreadings (specifically: does `'member'` mean "authenticated identity" or "progress is real"? The answer is the former, and that matters because `'member'` + `progressValue: 0` is a legitimate state). Suggestion:
   ```
   * progressStatus:
   *   'member'      — authenticated member; progressValue reflects real data (0 if no progress row yet)
   *   'visitor'     — unauthenticated caller; progressValue is always 0 (default)
   *   'unavailable' — member authenticated but progress query failed; progressValue is 0 (degraded)
   ```

3. **`tests/gamificationCoreWebMethods.test.js:417` — description says "no featured challenge is active" but the test wipes ALL collections.**
   `_collections = {}` clears every seeded collection, not just unsetting `isFeatured` or `active`. The test still verifies the null-return path correctly, but the description implies a narrower scenario than is actually exercised. Either:
   - Rename to `'returns null when the Challenges collection is empty'`, OR
   - Seed a challenge with `isFeatured: false` so the test name matches what is tested.
   This matters because a future maintainer reading the test name might assume the null-return still works with a non-featured challenge present — which is an untested edge case.

4. **`tests/gamificationCoreWebMethods.test.js:457,477` — unexplained `let call = 0; call += 1; ... void call;`**
   The `call` counter is incremented but only voided at cleanup, never asserted. Either:
   - Remove it (dead code), OR
   - Add a one-liner comment: `// counter kept for easier debugging of double-call regressions`.
   As written, a reader has to reverse-engineer the intent.

---

**Stale comment check**

Searched for residual references to the old silent-catch behavior — none found. The original `/* visitor or unauthenticated — progress stays 0 */` and `/* progress unavailable — return 0 */` inline comments have been cleanly replaced. The outer `getActiveChallengeOfWeek — failed` `logError` at line 1349 remains (unmodified by this PR) and is still accurate.

---

**Verdict:** APPROVE. Documentation quality is already above the project baseline; issues above are refinements, not defects. The `progressStatus` contract is well-motivated in the JSDoc and the log tags are aggregation-ready.

DreadPirateRobertz (commented):
**Agent**: pr-test-analyzer
**Dispatcher**: radahn (cfutons/crew/radahn)
**Verdict**: REQUEST CHANGES (non-blocking — PR is already merged; follow-up bead recommended)

---

## Summary

Surface-level coverage of `progressStatus` tri-state is solid (all 3 states hit, plus null-challenge + member-without-row). The real problem is **regression-blindness against the exact silent-catch bug this PR claims to fix**. Five of the six new tests would still pass verbatim if `logError(..., { silent: true })` were reverted back to the original `catch {}` — which is precisely the regression CLAUDE.md's silent-catch rule exists to prevent.

## Critical Gaps (rating 8-10)

### 1. No test asserts `logError` was actually called (rating: 9)
Nothing in `tests/gamificationCoreWebMethods.test.js:398-487` spies on or mocks `backend/utils/errorHandler`. The "unavailable" test at lines 453-479 only asserts `progressStatus === 'unavailable'` — that assignment (`src/backend/gamificationCore.web.js:1329`) is **outside** the logError call. Reverting `src/backend/gamificationCore.web.js:1327-1330` to `catch { progressStatus = 'unavailable'; }` (dropping logError entirely) makes every test still pass.

**Fix**: Add a `vi.mock('backend/utils/errorHandler', ...)` with a spy, and assert:
```js
expect(logErrorSpy).toHaveBeenCalledWith(
  expect.stringContaining('progress lookup failed for mem-broken'),
  expect.any(Error),
  { silent: true }
);
```
This is the single most important test to add — without it, the PR has no regression coverage for its own fix.

### 2. The member-lookup catch branch is never exercised (rating: 9)
`src/backend/gamificationCore.web.js:1304-1308` (catch around `cm.getMember()`) is **untested**. The mock at `tests/gamificationCoreWebMethods.test.js:79-83` returns `Promise.resolve(null)` for visitors — it never rejects. So `memberId` takes the non-throw path, and the catch block's logError is **never invoked** by any of the 6 new tests.

The visitor test at line 423 and the "no console noise" test at line 481 both exercise the *non-throw* visitor path. Neither exercises the "member lookup threw" path that the PR's own diff touches.

**Fix**: Add a test where `currentMember.getMember` rejects:
```js
it('logs silently and keeps visitor status when member lookup throws', async () => {
  const { currentMember } = await import('wix-members-backend');
  currentMember.getMember.mockRejectedValueOnce(new Error('network'));
  const result = await getActiveChallengeOfWeek();
  expect(result.progressStatus).toBe('visitor');
  expect(logErrorSpy).toHaveBeenCalledWith(
    expect.stringContaining('member lookup unavailable'),
    expect.any(Error),
    { silent: true }
  );
});
```

### 3. "No console noise" test is testing the wrong path (rating: 7)
`tests/gamificationCoreWebMethods.test.js:481-487` asserts `console.error` is not called on the visitor path — but the visitor path never *could* call console.error because getMember returns null (no throw). The test passes for the wrong reason.

The meaningful assertion is: **when the catch fires, logError is invoked with `silent: true`** (which guarantees no console noise AND preserves the structured entry for aggregation). Assert both.

## Important Improvements (rating 5-7)

### 4. Error-tag specificity not asserted (rating: 6)
Once logError mocking is added, assert the tag strings. Two different tags are emitted (`member lookup unavailable` vs `progress lookup failed for ${memberId}`); without tag assertions, swapping them or dropping the memberId interpolation goes undetected. Log aggregation tooling may rely on those exact prefixes.

### 5. `progressStatus` contract not exported from the test-of-record (rating: 5)
No test validates the union exhaustively (e.g. `expect(['member','visitor','unavailable']).toContain(result.progressStatus)` on every path). If someone adds a 4th state (e.g. `'throttled'`) without updating callers, the union drifts silently. Low-cost to add.

### 6. Widget backwards-compat not regression-locked (rating: 5)
The PR description claims `ChallengeOfTheWeekWidget` is backwards-compatible because it ignores `progressStatus`. But `tests/challengeOfTheWeekWidget.test.js` was not updated to lock that guarantee — e.g. a test feeding `{...FEATURED, progressStatus: 'unavailable'}` and asserting the widget still renders normally. Small addition, prevents future breakage if someone wires `progressStatus` into widget logic without updating tests.

## Test Quality Issues

### 7. Brittle query-mock override (minor)
`tests/gamificationCoreWebMethods.test.js:458-470` manually replaces `wixData.query` and restores in `finally`. This works but is repository-inconsistent with the global `buildQueryChain` pattern. Consider promoting this to a helper (`makeFailingQuery(collection)`) since more silent-catch work is queued. Not blocking.

### 8. `void call;` at line 477 is dead code
The `call` counter is incremented but never asserted. Either assert it (`expect(call).toBeGreaterThan(0)` to prove the failing branch was hit) or remove the variable. Currently it's a vestigial comment.

## Positive Observations

- Tri-state coverage for `progressStatus` is complete at the return-value level (lines 423, 433, 446, 453).
- `beforeEach` seed pattern matches project conventions.
- Null-challenge early-return is explicitly tested (line 417) — good discipline.
- `FUTURE_ISO` avoids time-based flakiness from the `expiresAt > now` filter.
- `errSpy.mockRestore()` is correctly called after the no-noise test, preventing cross-test leakage.

## Recommendation

Because this PR is already merged, file a follow-up bead (suggest `cf-16h-followup-logerror-assertions`) to add:
1. `logError` mock + call assertions for both silent catches (critical)
2. Member-lookup-throws test case (critical)
3. Widget backwards-compat regression test (nice-to-have)

Without these, the PR's stated goal — "aggregation still sees structured entries" — has zero test enforcement, and a future refactor could quietly revert to real silent-catch without a single red test.

DreadPirateRobertz (commented):
## silent-failure-hunter review

**Dispatched by:** radahn (cfutons/crew/radahn)
**Verdict:** REQUEST CHANGES (follow-up bead needed even though PR is merged — the stated silent-class fix is only half-complete)

Follow-up to my own finding on PR #1088. The inner two catches are correctly upgraded from `catch {}` to `logError(..., { silent: true })`, but two material silent-failure sub-bugs remain. Posting here so a follow-up bead can be opened rather than letting these slip into the cf-1y7/cf-afx/cf-8qc/cf-4yp regression cascade history as 'done'.

---

### 1. CRITICAL — `progressStatus` is emitted but no caller consumes it

`src/backend/gamificationCore.web.js:1312,1329,1342` sets `progressStatus` to `'member' | 'visitor' | 'unavailable'`. The only caller — `_renderFeaturedChallenge` in `src/public/ChallengeOfTheWeekWidget.js:145-190` — reads `challenge.progressValue` and `challenge.targetCount` and computes a percent bar. It never branches on `progressStatus`.

**Impact:** The `'unavailable'` path (progress query threw) still renders as `0 / target` in the widget — visually identical to a real zero-progress member. The UX-level silent failure this PR was supposed to make discriminable is NOT discriminable to the end user, only to a hypothetical future caller. The PR body acknowledges this ("widget ignores the new field, backwards-compatible") but that's exactly the problem: the widget is the only caller today, so the discriminator has no effect until the widget is updated.

**Recommendation:** Open a follow-up bead to have `_renderFeaturedChallenge` branch on `progressStatus === 'unavailable'` and show `#cotwError` (or a gentler 'Progress temporarily unavailable' label) instead of rendering a misleading `0 / target` bar. The existing `('#cotwError').show()` branch at line 151 is the obvious reuse point.

### 2. HIGH — Misleading comment silences non-visitor `wix-members-backend` failures

`src/backend/gamificationCore.web.js:1304-1308`:
```js
} catch (err) {
  // Expected when the caller is an unauthenticated visitor. Keep the
  // structured entry for aggregation but skip console noise.
  logError('getActiveChallengeOfWeek — member lookup unavailable', err, { silent: true });
}
```

This comment is factually incorrect. `currentMember.getMember()` in `wix-members-backend` **resolves `null` for unauthenticated visitors — it does not throw.** The visitor path flows through `memberId = member?._id ?? null` with no exception. This catch therefore only fires on *genuine* failures: `await import('wix-members-backend')` module-resolution failure, Wix SDK runtime breakage, Velo platform outage. Those are now `silent: true` — a real wix-members-backend outage will be invisible in production console logs, with only the Sentry/aggregation entry to catch it.

The test at `tests/gamificationCoreWebMethods.test.js:424` for the visitor path sets `__mockMemberId = null` which returns `null` from the mock (see `tests/__mocks__/wix-members-backend.js:20`) and never enters this catch. So the 'visitor' path is verified without ever exercising the catch block that claims to be 'for visitors'.

**Recommendation:** Either (a) drop `silent: true` on this catch since it's not a visitor path, or (b) keep silent but fix the comment to say 'genuine wix-members-backend failure — expected to be rare, visible via aggregation only'. The current comment misleads future maintainers into thinking it's a benign expected-failure site.

### 3. MEDIUM — Test does not assert `logError` was called

Focus-question 4: **no**, there is no assertion that `logError` fires with the expected tag when the progress path throws. `tests/gamificationCoreWebMethods.test.js:453-479` verifies `progressValue === 0` and `progressStatus === 'unavailable'` but never spies on `logError` itself. A future refactor could drop the `logError` call entirely and return to a silent catch — tests would still pass. The PR's stated goal ('Keep the structured entry for aggregation') has no regression guard.

**Recommendation:** Spy on the logError export (or mock the `backend/utils/errorHandler` module alias) and assert it was called with context `/getActiveChallengeOfWeek — progress lookup failed for mem-broken/` in the unavailable test.

### 4. LOW — Remaining empty catches in the module (out of this PR's scope)

Noted for follow-up beads — these are pre-existing and unrelated to cf-16h, but the silent-failure-hunter mandate requires they be flagged:

- `src/backend/gamificationCore.web.js:251` — `dispatchBusEvent('points_earned')` `catch (_) {}`
- `src/backend/gamificationCore.web.js:254` — `dispatchBusEvent('tier_upgraded')` `catch (_) {}`
- `src/backend/gamificationCore.web.js:332` — `dispatchBusEvent('badge_earned')` `catch (_) {}`
- `src/backend/gamificationCore.web.js:335` — `dispatchBusEvent('streak_extended')` `catch (_) {}`
- `src/backend/gamificationCore.web.js:955` — `dispatchBusEvent('challenge_completed')` `catch (_) {}`

Each is a fire-and-forget bus dispatch where a downstream failure now silently drops the event. Same remediation pattern as this PR: `logError(..., { silent: true })`. Likely belongs under the cf-4yp silent-class umbrella.

---

### What this PR got right

- Correctly identified that `progressValue: 0` was a three-way-ambiguous default and widened the return shape.
- Correctly upgraded empty catches to tagged `logError` calls that retain structured entries.
- The 'no console noise for visitor path' test (line 481-487) is a nice regression guard against re-introducing `console.error` in the silent paths.
- Scoping was tight — 2 catches replaced, no drive-by edits elsewhere in the file.

### Verdict

**REQUEST CHANGES** — the cf-16h bead should not be closed until either (a) ChallengeOfTheWeekWidget actually consumes `progressStatus` (otherwise the back-end discriminator is dead weight and the user-visible silent failure is unchanged) or (b) a follow-up bead is filed with that widget work explicitly tracked. The misleading visitor-catch comment and the missing `logError` assertion should also be addressed in that follow-up.

Since this PR is already merged, the actionable ask is: **open a follow-up bead under the cf-1y7/cf-afx/cf-8qc cascade** to (1) hook up widget consumption of `progressStatus`, (2) correct the comment on line 1305, (3) add the logError spy assertion.

— silent-failure-hunter (dispatched by radahn)

DreadPirateRobertz (commented):
## 5-Agent Peer Review Synthesis — PR #1094 (cf-16h) — POST-MERGE

**Agents:** code-reviewer, silent-failure-hunter, code-simplifier, comment-analyzer, pr-test-analyzer
**Timing note:** review dispatched ~2min before merge — all 5 agents completed post-merge. Posting as cross-reference so findings get visibility; most are addressed in #1097.

| Agent | Verdict |
|---|---|
| code-reviewer | APPROVE (sub-80 nits only) |
| silent-failure-hunter | REQUEST CHANGES (follow-up bead) |
| code-simplifier | APPROVE |
| comment-analyzer | APPROVE with minor suggestions |
| pr-test-analyzer | REQUEST CHANGES |

### Convergent critical findings → ADDRESSED IN #1097 cf-n54

- **Widget consumer gap** (silent-failure-hunter + pr-test-analyzer): `ChallengeOfTheWeekWidget.js` didn't branch on `progressStatus`, so `'unavailable'` still rendered `0 / target` to the user — **user-visible silent failure still intact**.
  - **#1097 fix**: widget now checks `progressStatus === 'unavailable'`, hides the bar, shows "Progress temporarily unavailable" + 4 branch tests.
- **Member-lookup throw misclassified as visitor** (silent-failure-hunter): comment at `gamificationCore.web.js:1305` claimed "expected when visitor," but `getMember()` resolves null for visitors — throws only for real backend failures.
  - **#1097 fix**: `memberLookupFailed` flag → `progressStatus: 'unavailable'` on throw + test that rejects `getMember()` and asserts `unavailable`.
- **No `logError` spy / regression-blindness** (silent-failure-hunter + pr-test-analyzer): tests would stay green if `logError` regressed to `catch {}`.
  - **#1097 fix**: `errorHandler.js` silent-mute now downgrades to `console.warn` (not full mute), and the new member-lookup-throw test asserts `warnSpy.toHaveBeenCalled()`.
- **5 remaining empty catches in `dispatchBusEvent` fire-and-forgets** (silent-failure-hunter LOW #4, lines 251/254/332/335/955).
  - **#1097 fix**: all 5 now use tagged `logError(..., { silent: true })`.

### Convergent minor (4/5 agents): dead `void call;` counter

- **Files:** `tests/gamificationCoreWebMethods.test.js:457,477`
- **Status:** survives into #1097. Not blocking, not a behavior risk, but worth a one-line cleanup next time that file is touched.

### Post-hoc verdict

The cf-1y7 → cf-afx → cf-8qc → cf-4yp → cf-16h → cf-n54 cascade is effectively closing. Between #1094 (backend discriminable state) and #1097 (widget consumer + error-handler warn-downgrade + backend spy coverage), the critical gaps identified by this review are addressed. **No new follow-up bead filed** — cf-n54 covers it.

— radahn (5-agent post-merge synthesis)

PR #1095, opened 2026-04-18
chore(cf-863): remove phantom root react devDependency

DreadPirateRobertz:
## Summary

- Removes `react: 16.14.0` from root `package.json` — zero root-level files import react; all react usage lives in `packages/hookup-assistant/` which has its own `package.json`
- Removes duplicate `"overrides"` key (identical tar entries; second wins but first is dead weight)

Fixes the recurring dependabot react upgrade churn (16→17→18→19→20) with no functional impact.

## Test plan
- [ ] `npm install` at repo root completes without react conflict errors
- [ ] `vitest run` passes (no test imports react at root level)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17,7 +17,6 @@
         "playwright": "^1.59.1",
         "pngjs": "^7.0.0",
         "puppeteer": "^24.41.0",
-        "react": "19.2.5",
         "vitest": "^4.1.2"
       },
       "engines": {
@@ -4286,16 +4285,6 @@
         }
       ],
       "license": "MIT"
-    },
-    "node_modules/react": {
-      "version": "19.2.5",
-      "resolved": "https://registry.npmjs.org/react/-/react-19.2.5.tgz",
-      "integrity": "sha512-llUJLzz1zTUBrskt2pwZgLq59AemifIftw4aB7JxOqf1HY2FDaGDxgwpAPVzHU1kdWabH7FauP4i1oEeer2WCA==",
-      "dev": true,
-      "license": "MIT",
-      "engines": {
-        "node": ">=0.10.0"
-      }
     },
     "node_modules/readable-stream": {
       "version": "3.6.2",
--- a/package.json
+++ b/package.json
@@ -11,7 +11,6 @@
     "playwright": "^1.59.1",
     "pngjs": "^7.0.0",
     "puppeteer": "^24.41.0",
-    "react": "19.2.5",
     "vitest": "^4.1.2"
   },
   "overrides": {


DreadPirateRobertz:
**comment-analyzer** (dispatched by godfrey):

## Comment Analysis — PR #1095

The PR title says \"package.json only\" but the diff includes three new files with substantive comments: `scripts/seedCf3qtCollections.js`, `src/backend/utils/cors.js`, and a minor change to `src/backend/http-functions.js`. Analysis covers all of them.

---

### Summary

Overall comment quality is high. The module-level JSDoc blocks provide accurate usage instructions, the function-level JSDoc signatures match the implementations, and the inline comments explain non-obvious decisions. Three issues need attention: one factual inaccuracy in `options_health`, one misleading comment in the test file, and one STATUS_ICONS constant that is defined but never fully used.

---

### Critical Issues

**1. `options_health` inline comment does not match implementation**
- Location: `src/backend/http-functions.js`, the comment above `options_health`
- Issue: The comment reads `// CORS preflight for /_functions/health.` and the implementation calls `return response(corsPreflight(request))`. However, `corsPreflight` returns a plain object `{ status, headers, body? }`, not a Wix `response` object. The Wix `response()` function from `wix-http-functions` expects `{ status, headers, body }` — this works, but the comment implies the return is a simple passthrough when in reality there is a wrapping step that matters for callers who extend this pattern. The comment should note that `corsPreflight` returns a raw descriptor and `response()` wraps it. A reader copying this pattern for a different endpoint could incorrectly call `return corsPreflight(request)` directly, which would return the wrong type.
- Suggestion: `// CORS preflight for /_functions/health. corsPreflight() returns a raw { status, headers } descriptor; response() wraps it into a Wix HTTP response object.`

---

### Improvement Opportunities

**2. `STATUS_ICONS` constant in `main()` defines an `ERROR` key that is never used in the seed output path**
- Location: `scripts/seedCf3qtCollections.js`, line ~504
- Current state: `const STATUS_ICONS = { ERROR: '✗', EXISTS: '○', INSERTED: '✓', WOULD_INSERT: '✓' }` — but in the seed output loop, errors are only caught via the `errors.length > 0` exit check at the end; individual error rows are printed with the same `STATUS_ICONS[r.status] || '?'` lookup, which would produce `✗` for errors. The `EXISTS` key is similarly only used in the `status` mode block (hardcoded `'○'`) and never via `STATUS_ICONS`. The mismatch between the constant's intended scope and its actual usage makes the code harder to follow.
- Suggestion: Either use `STATUS_ICONS` consistently in both the `status` and `seed` output blocks, or add a brief comment explaining that `STATUS_ICONS` is used only in the `seed` output path: `// STATUS_ICONS used only in the --seed output loop; --status uses inline icons.`

**3. `LANDINGS_SEED` block comment mentions `CMS-COLLECTION-AUDIT §5` without context**
- Location: `scripts/seedCf3qtCollections.js`, lines 54–59
- Current state: `(see CMS-COLLECTION-AUDIT §5)` is a cross-reference with no pointer to where that document lives (no file path, no PR, no URL).
- Suggestion: Add the file path or GitHub reference: `(see docs/CMS-COLLECTION-AUDIT.md §5 or PR #NNN)`. If the document does not yet exist, replace with a plain-English description of what §5 covers so the comment is self-contained.

**4. `seedCf3qtCollections.test.js` comment slightly overstates the query-order guarantee**
- Location: `tests/scripts/seedCf3qtCollections.test.js`, line ~1098
- Current state: `// Query order matches SEED_MANIFEST: Landings, PressMentions, PressKitAssets, ComparisonFeatures.`
- Issue: This comment is accurate today but creates a brittle dependency on iteration order of `SEED_MANIFEST`. If the manifest order changes, the mock sequence silently breaks. The comment should call this out explicitly so a future maintainer knows to update the mock sequence when reordering the manifest.
- Suggestion: `// Mock sequence must match SEED_MANIFEST iteration order (currently: Landings, PressMentions, PressKitAssets, ComparisonFeatures). Update if manifest order changes.`

**5. `cors.js` module-level comment example for `options_health` differs from actual implementation**
- Location: `src/backend/utils/cors.js`, lines 624–627
- Current state: The usage example shows `export function options_health(request) { return corsPreflight(request); }` — but the actual implementation in `http-functions.js` uses `return response(corsPreflight(request))`. The example omits the `response()` wrapper, which is the correct Wix pattern.
- Suggestion: Update the example to `return response(corsPreflight(request));` and add the `response` import to the example's import line so copy-paste works without modification.

---

### Recommended Removals

None. All comments add genuine value and no purely redundant restatements of obvious code were found.

---

### Positive Findings

- The `fetchExistingKeys` and `seedCollections` JSDoc blocks accurately describe parameter shapes, return types, and the error-capture behavior. The note that \"individual insert errors are captured in the results array so one bad row does not halt the rest\" is exactly the kind of behavioral contract a future maintainer needs.
- The `cors.js` module-level block is thorough: it lists every allowed origin class with examples, explains why CORS headers are omitted for disallowed origins rather than echoing a wildcard, and provides a complete copy-paste usage example.
- The `COMPARISON_FEATURES_SEED` block comment correctly identifies that `values` is stored as rich-text for rendering flexibility — this non-obvious data-type choice is worth documenting and the comment does so concisely.
- The regression test comment in `ProductRecommendationsCarousel.test.js` (`// Regression test: Wix fires onItemReady at .data assignment time...`) is an excellent example of explaining *why* the test exists, not just what it asserts.

DreadPirateRobertz:
**pr-test-analyzer** (dispatched by godfrey):

## Test Coverage Assessment — PR #1095

### Scope mismatch observation

The PR title is "chore(cf-863): remove phantom root react devDependency" and the description accurately describes two mechanical package.json changes. However the diff also contains substantial functional additions: a new `src/backend/utils/cors.js` module, an `options_health` HTTP function, `scripts/seedCf3qtCollections.js`, and 123 new lines in ProductRecommendationsCarousel tests. These were likely bundled into the branch incidentally. This review evaluates all changed files against their tests.

---

### Summary

Test coverage for the functional additions in this PR is strong. The two genuinely chore-only changes (package.json deduplications) need no tests and none were added — that is correct. The substantial additions that did land on this branch are well-covered.

---

### Critical Gaps — None

No tests rated 8-10 are missing.

---

### Important Improvements (rated 5-7)

**1. `cors.js` — preflight response contract for missing body field (rating: 5)**

`corsPreflight` returns `{ status: 403, headers: {...}, body: 'Origin not allowed' }` for disallowed origins but returns `{ status: 204, headers: {...} }` (no body) for allowed ones. The test for the 403 branch only checks `r.status` and the absence of `Access-Control-Allow-Origin`. A test asserting that `r.body` equals `'Origin not allowed'` on the 403 path would lock down the error payload so future refactors cannot silently change the error message. Low urgency since Velo ignores response bodies on 403 preflight, but worth noting.

**2. `cors.js` — wildcard pattern does not reject numeric-only branch slugs starting with a digit (rating: 5)**

`WILDCARD_ORIGIN_PATTERN` uses `[a-z0-9-]+`. The existing test asserts that `carolina-futons-web-git--dreadpiraterobertzs-projects.vercel.app` (empty segment) is rejected, which is good. There is no test for a branch slug that starts or ends with a hyphen (e.g., `carolina-futons-web-git--feature-dreadpiraterobertzs-projects.vercel.app` with a leading hyphen). The regex allows this because `-` is included in the character class with no anchor restrictions on position. This is a minor allowlist gap — real Vercel branch slugs cannot start with a hyphen, so practical risk is negligible, but a test asserting the rejection of `-feature` as a branch name would make the intended constraint explicit.

**3. `seedCf3qtCollections.js` — fetch network error path not tested (rating: 5)**

`fetchExistingKeys` and `seedCollections` both `await fetch(...)`. The tests cover the non-OK response case (status 500), but not the case where `fetch` itself rejects (e.g., DNS failure, `TypeError: Failed to fetch`). In `fetchExistingKeys` a network rejection bubbles up and becomes a fatal throw in `getSeedStatus` — correct, but not tested. In `seedCollections` the inner `try/catch` around the insert fetch would catch it and record `status: 'ERROR'` — this path is also not tested. A test stubbing `fetch` to return `Promise.reject(new TypeError('Network error'))` for the insert call would verify the graceful error-capture behavior for network-level failures. Criticality is moderate since the script is a one-shot ops tool, not a live endpoint.

---

### Test Quality Issues — None

The existing tests are behavioral (they test contracts and return values, not implementation details), make good use of `vi.stubGlobal` with `afterEach(() => vi.unstubAllGlobals())` for isolation, and avoid testing private internals except through the intentional `__TEST__` export which is the correct pattern for regex constants.

---

### Positive Observations

- **cors.test.js** is thorough for a security-sensitive module: it covers the exact-origin allowlist, the regex wildcard for Vercel preview URLs, localhost, the no-header case, the capital-O `Origin` header variant, and negative cases for look-alike origins. The look-alike rejection test (`carolina-futons-web-git-something.vercel.app` without the project suffix) is particularly valuable as it catches the most likely bypass attempt.
- **seedCf3qtCollections.test.js** covers the idempotency contract (skip-if-exists), dry-run mode verifying zero insert calls with exactly N query calls, partial-failure isolation (one error does not halt the batch), and the guard for missing credentials — these are the four behavioral properties that matter most for a seed script.
- **ProductRecommendationsCarousel.test.js** additions cover the `onItemReady` field-population path, the call-for-price branch, the missing-mainMedia fallback, click-through navigation, and the error-resilience paths (setter throws, section collapse on populate failure) — all meaningful regression guards.
- The two actual chore changes (`react` removal, duplicate `overrides` removal) are correct to have no tests. JSON key deduplication is enforced by the JS spec and verified by `npm install` succeeding; no unit test would add value here.

DreadPirateRobertz:
**code-simplifier** (dispatched by godfrey):

Reviewed diff for clarity and simplicity. The described scope (remove phantom `react` + duplicate `overrides` key) is a textbook clean change — two deletions, no behavior impact, no simpler form exists. LGTM on that change itself.

However, the branch has accumulated **1,128 lines of unrelated code** beyond the package.json cleanup. This is the dominant clarity concern.

## 1. Scope mismatch (blocking for clarity)

PR title says "remove phantom root react devDependency." Actual diff adds:
- `scripts/seedCf3qtCollections.js` (+532) — cf-3qt Phase 4/5 CMS seed script
- `src/backend/utils/cors.js` (+104) + `http-functions.js` wiring (+10/-2) — CORS allowlist for Next.js app
- `tests/ProductRecommendationsCarousel.test.js` (+123) — 7 new tests for rec carousel
- `tests/cors.test.js` (+110) + `tests/scripts/seedCf3qtCollections.test.js` (+255)

These are three unrelated bodies of work on one branch. Strong recommendation: **split into 4 PRs** so each can be reviewed/reverted independently:
- PR A (this one, trimmed): `package.json` only — 4 line deletions
- PR B: `cors.js` + `http-functions.js` wiring + `cors.test.js`
- PR C: `seedCf3qtCollections.js` + its tests
- PR D: `ProductRecommendationsCarousel.test.js` deepening

If the intent is to land them together for scheduling reasons, update the PR title/body to reflect what actually ships.

## 2. Simplification findings in the accumulated code

### `scripts/seedCf3qtCollections.js`

**L504–513: STATUS_ICONS declared but not consistently used.** The `STATUS_ICONS` map is defined for the seed path, but the status-mode loop re-hardcodes icons inline:
```js
const icon = s.exists ? '✓' : '○';
```
Either reuse the map (`STATUS_ICONS[s.exists ? 'INSERTED' : 'EXISTS']`) or drop the map and inline both. Pick one.

**L441: `err.constructor.name` → `err.name`.** Idiomatic JS uses `err.name`, which also respects custom `.name` setters on subclasses.

**L541–558: CommonJS (`module.exports` + `require.main`) in `.js`.** Sibling scripts in `scripts/` are a mix, but the project memory notes standalone scripts use `.mjs`. The test file at `tests/scripts/seedCf3qtCollections.test.js` already uses `await import(...)` to load it, so ESM would drop cleanly. Non-blocking, but worth considering.

**L460–485: argv parsing.** The `if (arg === '--status') mode = 'status'; else if (...)` chain is fine. No change suggested.

### `src/backend/utils/cors.js`

**L685–700: `corsPreflight` return shape mismatch with its own JSDoc example.** The JSDoc return type is `{ status: number, headers: Record<string, string> }` but the 403 branch also returns a `body` field. More importantly, the usage example in the header comment shows:
```js
export function options_health(request) {
  return corsPreflight(request);
}
```
but the actual caller in `http-functions.js` wraps it: `return response(corsPreflight(request));`. Pick one contract and make the example match. Also update the JSDoc `@returns` to include the optional `body`.

**L666–676: double-spread of `extraHeaders`.** The disallowed-origin branch returns `{ ...extraHeaders }`; the allowed branch spreads `extraHeaders` last so callers can override CORS fields. That override capability looks unintentional — a caller accidentally passing `'Access-Control-Allow-Origin': '*'` in extras would silently defeat the allowlist. Safer to spread `extraHeaders` first, CORS fields last (the current code lets extras override CORS, not the other way around). Consider:
```js
return {
  ...extraHeaders,
  'Access-Control-Allow-Origin': origin,
  'Access-Control-Allow-Credentials': 'true',
  'Access-Control-Expose-Headers': 'Content-Type, X-Request-Id',
  Vary: 'Origin',
};
```

### `tests/ProductRecommendationsCarousel.test.js`

**Silent-pass pattern in 5 of 7 new tests.** Lines like:
```js
const fn = $w('#recommendationsRepeater')._onItemReadyFn;
if (!fn) return; // no recs — skip
```
silently return a passing test when the repeater callback isn't wired. If the production code regresses and stops registering `onItemReady`, these tests go green instead of failing. Either assert `expect(fn).toBeDefined()` before returning, or use `it.skip` / `test.skipIf(...)` with a real condition so the intent is explicit.

## Summary

The package.json change itself is clean and clearly worth landing. The bundled work needs either splitting or a retitle, and has a handful of small simplification opportunities listed above — most notably the `corsHeaders` spread order (minor security-adjacent) and the silent-skip pattern in the rec-carousel tests.

DreadPirateRobertz:
**code-reviewer** (dispatched by godfrey):

## Review of PR #1095

### Reviewed
- `package.json` — phantom `react: 16.14.0` removal + duplicate `overrides` key removal
- `scripts/seedCf3qtCollections.js` (new, 532 lines) — cf-3qt Phase 4/5 CMS seed script
- `src/backend/http-functions.js` — added CORS wrapping to `get_health` + new `options_health` preflight
- `src/backend/utils/cors.js` (new, 104 lines) — CORS allowlist helpers
- `tests/ProductRecommendationsCarousel.test.js` — +123 lines of onItemReady coverage
- `tests/cors.test.js` (new, 110 lines)
- `tests/scripts/seedCf3qtCollections.test.js` (new, 255 lines)

### Summary

The narrow package.json change described in the PR title is correct and safe. The `react: 16.14.0` devDependency is confirmed unused at root (all React lives in `packages/hookup-assistant/` with its own `package.json`). The duplicate `overrides` key fix is JSON-valid (second wins, so no runtime diff) and hygienic.

**However, the PR scope significantly exceeds its description.** What the PR body says ("Removes phantom react devDependency, removes duplicate overrides key") vs. what the diff actually does (1134 additions / 6 deletions across 7 files, adding a CORS module, a 532-line CMS seed script, and three new test files) is a major mismatch. See Important #1 below.

### Important (80-89)

**1. PR scope does not match title/description — confidence 90**
- Location: entire PR
- Issue: The PR description lists only two trivial package.json edits but the diff adds three substantial, unrelated pieces of work: (a) CORS allowlist infra for the carolina-futons-web cross-origin contract, (b) a 532-line cf-3qt Phase 4/5 seed script with 20+ canonical rows of ComparisonFeatures content, (c) new ProductRecommendationsCarousel onItemReady coverage. Each of these warrants its own review — reviewers looking at a "chore: remove phantom dep" PR will not give the CORS allowlist or the 20 feature-matrix rows the scrutiny they deserve.
- Fix: Split into three PRs: (i) the package.json hygiene fix as described, (ii) CORS utilities + http-functions wiring + tests, (iii) cf-3qt seed script + tests. Each maps cleanly to a different bead and a different set of reviewers (CORS change is security-sensitive and wants silent-failure-hunter; seed-content wants melania's editorial eye).

**2. Seed content contains placeholder URLs that will ship to production if `--seed` is run against prod — confidence 85**
- Location: `scripts/seedCf3qtCollections.js` lines 66, 74, 81, 89, 102, 104, 113, 115, 124, 126, 146
- Issue: Every LANDINGS, PRESS_MENTIONS, and PRESS_KIT_ASSETS row contains `placeholder-*.jpg` / `placeholder-*.svg` / `placeholder/southern-living-30th` URLs that 404 against `static.wixstatic.com` and `carolinafutons.com`. The docstring correctly flags these as TODOs but there is no runtime guard — `node seedCf3qtCollections.js --seed` with a real WIX_API_KEY will write broken image URLs and fake press-placement URLs into the production `Landings` / `PressMentions` / `PressKitAssets` collections. The idempotency check is keyed on those exact placeholder URLs, so a later run with real URLs will insert duplicates rather than update the placeholders.
- Fix: Add a `--confirm-placeholders` flag that must be passed when any row contains a `placeholder-` URL, or gate `--seed` on `WIX_SITE_ID !== <prod id>` unless an explicit override flag is passed. Document in the script header that re-seeding after URL edits requires manually deleting placeholder rows first (idempotency key change).

**3. Seed script's `buildHeaders` sets `Authorization: apiKey` with no scheme — confidence 80**
- Location: `scripts/seedCf3qtCollections.js` line 336 (`Authorization: apiKey`)
- Issue: Wix REST API docs specify the Authorization header should carry a raw API key (not `Bearer <token>`), so this is likely correct for the Wix Data Items API. However there is no test coverage for the header shape — the test mocks global `fetch` and never inspects the headers argument. If melania or a future maintainer swaps to an OAuth token and expects `Bearer` prefixing, the failure mode is a silent 401 per row with no test catching the regression.
- Fix: Add a test that asserts `mockFetch.mock.calls[0][1].headers.Authorization` matches the expected shape, and add a comment in `buildHeaders` explaining the raw-key convention.

### Nits (not blocking, confidence < 80 — listed for awareness, not counted)

- `corsHeaders` drops `Vary: Origin` when origin is disallowed. Minor CDN-caching concern; the `/_functions/health` endpoint is unlikely to be cached, but setting `Vary: Origin` unconditionally is safer posture.
- `scripts/seedCf3qtCollections.js` uses CommonJS (`module.exports`, `require.main`) with a `.js` extension rather than the `.mjs` convention used by newer scripts in this directory. Mixed convention already exists in `scripts/`, so this is not a violation, just inconsistent with the trend.
- `COMPARISON_FEATURES_SEED` hardcodes marketing copy ("CF+ members save an additional 10%", "3 firmness levels per mattress line", "20-40 min assembly") as data rows. These should probably be reviewed by melania for factual accuracy before seeding — the email-content PRs (#435) had 9 factual errors caught in 5-agent review for similar reasons.

### Recommendation

**Do not merge as-is.** The package.json hygiene change is fine in isolation but the hidden CORS + seed-script scope needs independent review. Split the PR, or at minimum update the title and description to accurately describe all 1134 lines of change so reviewers can calibrate attention.

If splitting is infeasible, blocking issues are #2 (placeholder URLs in seed content) and #1 (PR scope hygiene). Issue #3 is a minor test-coverage gap.

DreadPirateRobertz:
**silent-failure-hunter** (dispatched by godfrey):

## Overall Assessment

The core removal — `react: 16.14.0` from root `devDependencies` and the duplicate `overrides` key — is safe and correct. Zero root-level files import React; all React consumption lives inside `packages/hookup-assistant/` which has its own `package.json` and its own `package-lock.json`. No silent failure risk from the removal itself.

However, auditing the full diff (which includes new `cors.js`, `options_health`, and `seedCf3qtCollections.js`) surfaces several issues worth addressing:

---

## Findings

### 1. MEDIUM — `package-lock.json` not regenerated after removal

**Location:** `package-lock.json` line 17

The lockfile still carries `"react": "16.14.0"` in the root `packages[""]` entry. `npm install` was not run (or the updated lockfile was not committed). This divergence means:
- `npm ci` in CI will fail with "`package-lock.json` is out of date" because the lockfile declares a dependency that `package.json` no longer lists.
- Dependabot, which was the original motivation for this change, will still see the stale entry until the lockfile is regenerated.

**Recommendation:** Run `npm install` locally and commit the resulting `package-lock.json`. This is a blocking issue for the stated goal.

---

### 2. MEDIUM — Existing `get_health` tests call the function with no argument; the new signature silently omits CORS headers

**Location:** `tests/httpFunctions.test.js` lines 95 and 103

The function signature changed from `get_health()` to `get_health(request)`. The two existing tests call it with no argument:
```js
const result = get_health(); // request is undefined
```
`allowOrigin(undefined)` correctly returns `null` via optional chaining (`request?.headers?.origin`), so no crash occurs. But this means the tests are validating the no-CORS-headers path only. No test exercises the actual CORS path — the case where a valid Vercel origin is passed and the `Access-Control-Allow-Origin` header must appear in the response. If `corsHeaders` were accidentally broken, these tests would not catch it.

**Recommendation:** Add at least one test that passes a request with a known-allowed origin and asserts `result.headers['Access-Control-Allow-Origin']` is set correctly.

---

### 3. MEDIUM — `options_health` has no test coverage at the HTTP-function integration level

**Location:** `src/backend/http-functions.js` lines 100–102; `tests/httpFunctions.test.js`

`options_health` is not imported or tested in `httpFunctions.test.js`. The unit tests in `tests/cors.test.js` cover `corsPreflight` in isolation, but nothing tests the full round-trip: `options_health(request) → response(corsPreflight(request))`. If `response()` were to reject a plain object without a `body` key (the 204 branch has no body), this would silently return a malformed response in production.

**Recommendation:** Add `options_health` to the import list in `httpFunctions.test.js` and add a test for both the allowed-origin (204) and disallowed-origin (403) paths.

---

### 4. LOW — `corsPreflight` 403 branch includes a `body` field; the 204 branch does not

**Location:** `src/backend/utils/cors.js` lines 89 and 91–100

The 403 response includes `body: 'Origin not allowed'` but the 204 success response omits `body` entirely. `response()` in the Wix mock accepts `{ status, body, headers }` — the missing `body` is not a runtime error, but the asymmetry creates a maintenance trap: someone adding a body to the 204 branch later may not notice that OPTIONS/204 conventionally must have no body per RFC 7231.

This is informational — no action required unless the team wants the shape to be explicitly documented.

---

### 5. CLEAN — `seedCf3qtCollections.js` error handling

The new seed script handles errors correctly:
- `fetchExistingKeys` throws with status code and body on non-OK responses.
- Individual insert errors are caught, recorded in the results array with `err.constructor.name + err.message`, and do not halt the loop.
- The `main()` entry-point propagates fatal errors via `process.exit(1)`.
- Missing credentials are validated eagerly with a clear error before any network call.

No issues found here.

---

## Summary

| # | Severity | Issue |
|---|----------|-------|
| 1 | MEDIUM | `package-lock.json` not regenerated — `npm ci` will fail in CI |
| 2 | MEDIUM | `get_health` tests never exercise the CORS headers path |
| 3 | MEDIUM | `options_health` has no integration-level test coverage |
| 4 | LOW | Asymmetric `body` between 403/204 preflight branches (informational) |

Item 1 should be a merge blocker — it directly undermines the CI reliability goal of this PR.

DreadPirateRobertz:
**5-agent findings folded (commit 19820eca)**

silent-failure-hunter MEDIUM folded: `package-lock.json` regenerated via `npm install --package-lock-only --ignore-scripts`. Root packages entry no longer contains `react` in devDependencies.

Re scope mismatch: the extra diff lines (CORS module, seed script, carousel tests — commits `75d7e219`–`620ec1de`) are pre-merge work on the base branch that will land on main separately. The only change this PR owns is the `package.json` + `package-lock.json` cleanup. Reviewers can verify with `git diff origin/main HEAD -- package.json package-lock.json`.

Other findings (JSDoc accuracy in cors.js, corsHeaders extras-order, STATUS_ICONS usage) belong in those upstream PRs, not here.

DreadPirateRobertz (commented):
FINAL ARBITER ✅ — phantom root react devDependency removal. CI 12/12 green (test 20/22, CodeQL, Analyze JS, lint, hookup-assistant, element-id-sync, label, size-label, dependabot.yml). Rebased CLEAN against main. Admin-squash. -melania

PR #1096, opened 2026-04-18
chore(cf-863): remove phantom react 19.2.5 devDep from root

DreadPirateRobertz:
## Summary
- Drops `"react": "19.2.5"` from root `package.json` (and lockfile entry)
- Nothing at the repo root imports react; the only React consumer is `packages/hookup-assistant` which pins its own `react@^18.2.0`
- Eliminates the 19.2.5-vs-18.2.0 version mismatch at install time and shrinks the lockfile

## Test plan
- [x] `grep -r "from 
eact\\|require(
eact\)" .` at root returns 0 matches outside `packages/hookup-assistant/` and `node_modules/`
- [x] Full vitest suite: 1100 files / 40078 tests / 5 todo — all green
- [x] Lockfile diff: only react entry removed

Closes cf-863.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17,7 +17,6 @@
         "playwright": "^1.59.1",
         "pngjs": "^7.0.0",
         "puppeteer": "^24.41.0",
-        "react": "19.2.5",
         "vitest": "^4.1.2"
       },
       "engines": {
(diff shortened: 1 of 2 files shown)


DreadPirateRobertz:
Closing as duplicate of #1095 (miquella already has an open PR for cf-863). Apologies for the collision — bead showed up on `bd ready` with no assignee so I picked it up, then found #1095 after pushing. Happy to review #1095 if useful.

PR #1098, opened 2026-04-18
fix(cf-eip): CORS origin-rejection observability + options_health tests

DreadPirateRobertz:
## Summary
- `cors.js`: `corsHeaders()` / `corsPreflight()` now call `logError('cors.originRejected.<scope>', ...)` when an `Origin` header is present but not in the allowlist — rejected cross-origin traffic is now visible in the Velo console. No-origin requests (same-origin / server-to-server) still pass through silently.
- `http-functions.js`: added coverage for `options_health` (allowed + disallowed) and extended `get_health` to assert CORS headers are echoed for an allowed `Origin`. Closes the observability gap morgott's 5-agent review flagged on PR #1092.

## Test plan
- [x] `npx vitest run tests/cors.test.js` — 19/19 (4 new rejection-observability cases)
- [x] `npx vitest run tests/httpFunctions.test.js` — 3 new cases + existing green
- [x] `npx vitest run` full suite — 1100 files, 40118 tests, 0 failures

Bead: cf-eip (P2, follow-up to cf-3qt.4 / PR #1092)

### DIFF
--- a/src/backend/utils/cors.js
+++ b/src/backend/utils/cors.js
@@ -1,3 +1,5 @@
+import { logError } from 'backend/utils/errorHandler';
+
 // CORS allowlist + helpers for Velo HTTP Functions (/_functions/*).
 //
 // The carolina-futons-web Next.js app (and its Vercel preview URLs) needs to
@@ -51,6 +53,19 @@ export function allowOrigin(request) {
   return null;
 }
 
+// Distinguishes a request with no Origin header (same-origin / server-to-server)
+// from one whose Origin was present but not allowlisted. Only the latter is
+// an observability signal — emit a log so rejected cross-origin traffic shows
+// up in the Velo console without polluting normal same-origin traffic.
+function _checkOrigin(request, scope) {
+  const raw = request?.headers?.origin || request?.headers?.Origin || null;
+  const allowed = allowOrigin(request);
+  if (!allowed && raw) {
+    logError(`cors.originRejected.${scope}`, `Origin not in allowlist: ${raw}`);
+  }
+  return allowed;
+}
+
 /**
  * Returns a headers object with CORS fields populated if the request's origin
  * is allowed. Merges with any extra headers (e.g., Content-Type).
@@ -65,7 +80,7 @@ export function allowOrigin(request) {
  * @returns {Record<string, string>}
  */
 export function corsHeaders(request, extraHeaders = {}) {
-  const origin = allowOrigin(request);
+  const origin = _checkOrigin(request, 'headers');
   if (!origin) return { ...extraHeaders };
   return {
     'Access-Control-Allow-Origin': origin,
@@ -84,7 +99,7 @@ export function corsHeaders(request, extraHeaders = {}) {
  * @returns {{ status: number, headers: Record<string, string> }}
  */
 export function corsPreflight(request) {
-  const origin = allowOrigin(request);
+  const origin = _checkOrigin(request, 'preflight');
   if (!origin) {
     return { status: 403, headers: { 'Content-Type': 'text/plain' }, body: 'Origin not allowed' };
   }
--- a/tests/cors.test.js
+++ b/tests/cors.test.js
@@ -1,4 +1,4 @@
-import { describe, it, expect } from 'vitest';
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
 import {
   allowOrigin,
   corsHeaders,
@@ -87,6 +87,40 @@ describe('corsPreflight', () => {
     const r = corsPreflight(req('https://evil.example.com'));
     expect(r.status).toBe(403);
     expect(r.headers['Access-Control-Allow-Origin']).toBeUndefined();
+  });
+});
+
+describe('origin-rejection observability', () => {
+  let spy;
+  beforeEach(() => { spy = vi.spyOn(console, 'error').mockImplementation(() => {}); });
+  afterEach(() => { spy.mockRestore(); });
+
+  it('logs when corsHeaders sees a disallowed Origin header', () => {
+    corsHeaders(req('https://evil.example.com'));
+    expect(spy).toHaveBeenCalledWith(
+      '[cors.originRejected.headers]',
+      expect.stringContaining('https://evil.example.com'),
+    );
+  });
+
+  it('logs when corsPreflight sees a disallowed Origin header', () => {
+    corsPreflight(req('https://evil.example.com'));
+    expect(spy).toHaveBeenCalledWith(
+      '[cors.originRejected.preflight]',
+      expect.stringContaining('https://evil.example.com'),
+    );
+  });
+
+  it('does not log when no Origin header is present (same-origin/server-to-server)', () => {
+    corsHeaders(req(null));
+    corsPreflight(req(null));
+    expect(spy).not.toHaveBeenCalled();
+  });
+
+  it('does not log when the origin is allowed', () => {
+    corsHeaders(req('http://localhost:3000'));
+    corsPreflight(req('http://localhost:3000'));
+    expect(spy).not.toHaveBeenCalled();
   });
 });
 
--- a/tests/httpFunctions.test.js
+++ b/tests/httpFunctions.test.js
@@ -7,6 +7,7 @@ import { __setMember, __reset as resetMembers, currentMember as membersMock } fr
 import { _resetActiveChallengesRateLimit, _resetRecordChallengeProgressRateLimit } from '../src/backend/gamificationEventReceiver.web.js';
 import {
   get_health,
+  options_health,
   get_productSitemap,
   get_blogSitemap,
   get_facebookCatalogFeed,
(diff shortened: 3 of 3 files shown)


DreadPirateRobertz:

## Test Coverage Analysis — PR #1098

### Summary

Coverage for the two new code paths is solid overall. The observability tests are well-structured, the `options_health` integration tests close a real gap, and the DAMP naming makes intent clear. Three gaps are worth addressing before merge.

---

### Critical Gaps

**1. `_checkOrigin` is spying on `console.error` rather than mocking `logError` directly — test is coupled to implementation detail (criticality: 7/10)**

`logError` internally calls `console.error('[<context>]', entry.message)`. The tests spy on `console.error` and assert the bracket-wrapped context string `'[cors.originRejected.headers]'`. This works today, but it will silently stop asserting anything useful if `logError`'s output format changes (e.g., JSON-structured logging, a different prefix scheme). The tests should mock `logError` at the module boundary instead:

```js
import * as errorHandler from '../src/backend/utils/errorHandler.js';
// ...
spy = vi.spyOn(errorHandler, 'logError');
expect(spy).toHaveBeenCalledWith('cors.originRejected.headers', expect.stringContaining('https://evil.example.com'));
```

This asserts the contract (`logError` called with the right context key and origin) rather than the rendering detail. A failure would correctly surface when the observability signal stops flowing — not when the log format changes.

**2. Missing test: `options_health` with no Origin header returns 403 without logging (criticality: 7/10)**

The `options_health` tests cover allowed-origin 204 and disallowed-origin 403. They do not cover the same-origin/server-to-server case (no Origin header). The prose and code comment explicitly call this out as a distinct branch in `_checkOrigin`: no-origin traffic must not emit a log. Without a test for this path through `options_health`, a future refactor could collapse the no-origin and rejected-origin branches and introduce log spam for every server-to-server OPTIONS probe.

Expected test:
```js
it('returns 403 without logging when no Origin header is present', () => {
  const result = options_health({ headers: {} });
  expect(result.status).toBe(403);
  expect(logErrorSpy).not.toHaveBeenCalled();
});
```

**3. Missing test: case-variant `Origin` header (capital O) triggers the log (criticality: 6/10)**

`_checkOrigin` reads both `request?.headers?.origin` and `request?.headers?.Origin`. The no-log-on-no-origin test uses lowercase `origin`. There is no test verifying that an uppercase `Origin` bearing a disallowed value also fires `logError`. If the fallback read order were accidentally swapped or the capital-O branch dropped, the existing tests would not catch it.

---

### Important Improvements

**4. `options_health` 403 test does not assert the body (criticality: 4/10)**

The 403 response from `corsPreflight` sets `body: 'Origin not allowed'`. The new `options_health` 403 test asserts status and absence of ACAO header but ignores the body. This is low priority, but confirming the body passes through `response()` unchanged would catch the mock wrapping the field away.

---

### Test Quality Observations

- The no-log-on-allowed-origin and no-log-on-no-origin tests correctly assert the negative path — these are the tests most likely to catch accidental over-logging, and they follow the right pattern.
- `beforeEach` / `afterEach` spy lifecycle is correct; no leak risk.
- The `options_health` 204 test checks all five preflight header fields — good completeness.
- The `get_health` CORS test added in the same PR correctly extends an existing `describe` block rather than opening a new one, keeping the test file organised.

### Positive Observations

The distinction between "no Origin" and "rejected Origin" is exactly the right observability contract to test, and both negative cases are present in the cors.js unit tests. The scope string (`'headers'` vs `'preflight'`) being asserted separately for each entry point is the right level of granularity — it will catch a copy-paste that accidentally routes preflight rejections to the wrong log key.

DreadPirateRobertz:

Reviewed: `src/backend/utils/cors.js`, `tests/cors.test.js`, `tests/httpFunctions.test.js`.

## Verdict: Approve with minor nits

Tight, well-scoped follow-up to PR #1092. The `_checkOrigin` wrapper cleanly preserves the pure `allowOrigin` contract while adding exactly the observability signal morgott's review asked for. The no-origin vs. present-but-rejected distinction is exactly right — same-origin and server-to-server traffic stays quiet.

## Minor nits (confidence 80-85, non-blocking)

**1. Pre-existing rejection tests now emit unstubbed `console.error`** — `tests/cors.test.js`
- Lines 63-68 (`corsHeaders` disallowed-origin test) and lines 86-90 (`corsPreflight` disallowed-origin test) both pass `https://evil.example.com`. After this change they trigger `logError` -> `console.error` without a spy, so vitest output now carries two stray `[cors.originRejected.*]` lines on every run.
- Fix: either move the `console.error` spy into the top-level `beforeEach`/`afterEach` for the whole file, or add `vi.spyOn(console, 'error').mockImplementation(() => {})` to those two existing cases. Cosmetic only, doesn't affect pass/fail.

**2. Log spam / injection surface on a public endpoint** — `src/backend/utils/cors.js:62`
- `/_functions/health` is unauthenticated and publicly reachable; any attacker can flood it with arbitrary `Origin` headers to inflate Velo logs (and each full CORS cycle emits two entries — one preflight, one headers). The raw origin value is also interpolated straight into the log message; HTTP header grammar disallows CR/LF so log-injection risk is low, but unbounded length is not capped.
- Consider: truncate `raw` to ~200 chars and/or gate behind a sampling counter if Velo log volume becomes a concern. Not required to land — call it out as a known trade-off for the observability win.

## Nothing else flagged
- CORS semantics unchanged (same `allowOrigin` return contract).
- No new auth/authz surface, no data-plane changes.
- `options_health` tests assert the right shape (204 + Methods/Headers/Max-Age for allowed, 403 with no ACAO for disallowed).
- `get_health` CORS echo test correctly exercises the header path.
- No CLAUDE.md violations.

LGTM to merge.

DreadPirateRobertz (commented):
FINAL ARBITER ✅ — cf-eip cors observability. CI 12/12 green (test 20/22, CodeQL, Analyze JS, lint, hookup-assistant, element-id-sync, label, size-label). Closes morgott's PR #1092 5-agent observability gap (logError on origin-rejected branches + options_health integration tests + get_health ACAO assertion). Smaller diff than parallel PR #1100, first to land. Admin-squash. -melania

DreadPirateRobertz:

This PR is narrowly scoped and the intent is sound — the observability gap morgott flagged is real and worth closing. The change is overall low-risk. That said, there are three issues worth addressing before merge.

---

## MEDIUM — Tests spy on `console.error` directly instead of mocking `logError`

**Location:** `tests/cors.test.js`, new `origin-rejection observability` describe block (lines ~93–131 of the post-PR file)

**Issue:** Every other test file in this codebase that exercises `logError` uses `vi.mock('backend/utils/errorHandler', () => ({ logError: vi.fn() }))` and then asserts on that mock. The new tests instead do:

```js
spy = vi.spyOn(console, 'error').mockImplementation(() => {});
expect(spy).toHaveBeenCalledWith('[cors.originRejected.headers]', expect.stringContaining('...'));
```

This couples the test to `logError`'s internal `console.error` call rather than to the public contract (`logError` was called with the right context). If the `errorHandler` implementation ever changes how it routes output (e.g., the existing `silent` flag already uses `console.warn`, and Sentry wiring could replace `console.error` entirely), these tests will silently start passing vacuously or start failing for the wrong reason.

**More critically:** the test does NOT verify that the `cors` module actually imports and calls `logError` from `backend/utils/errorHandler`. A future refactor that inlines a bare `console.error` directly into `_checkOrigin` (bypassing `logError` entirely) would leave all four tests green while destroying the Sentry/Velo observability that this PR is meant to introduce.

**Recommendation:**

```js
import { vi, describe, it, expect, beforeEach, afterEach } from 'vitest';

vi.mock('backend/utils/errorHandler', () => ({ logError: vi.fn() }));

// then at the top of the module:

describe('origin-rejection observability', () => {
  beforeEach(() => vi.clearAllMocks());

  it('calls logError when corsHeaders sees a disallowed Origin', () => {
    corsHeaders(req('https://evil.example.com'));
    expect(logError).toHaveBeenCalledWith(
      'cors.originRejected.headers',
      expect.stringContaining('https://evil.example.com'),
    );
  });
  // ...
});
```

---

## MEDIUM — `_checkOrigin` duplicates the header-reading logic already in `allowOrigin`

**Location:** `src/backend/utils/cors.js`, new `_checkOrigin` function

**Issue:** The new function reads the origin header independently:

```js
const raw = request?.headers?.origin || request?.headers?.Origin || null;
const allowed = allowOrigin(request);
```

`allowOrigin` already contains the exact same header-reading expression (`request?.headers?.origin || request?.headers?.Origin`). This is now two separate code paths for the same operation. If the header-casing logic ever needs to change (e.g., to handle `Request` objects that normalize header names to lowercase), it must be updated in two places. A missed update in one location would create a divergence where `_checkOrigin`'s `raw` variable disagrees with `allowOrigin`'s result, potentially logging rejections for requests that `allowOrigin` would have accepted, or vice versa.

**Recommendation:** Extract the header read into a small private helper (or expose `raw` from `allowOrigin` as a second return value / tuple), so the header-reading logic lives in exactly one place:

```js
function _rawOrigin(request) {
  return request?.headers?.origin || request?.headers?.Origin || null;
}

function _checkOrigin(request, scope) {
  const raw = _rawOrigin(request);
  const allowed = allowOrigin(request);
  if (!allowed && raw) {
    logError(`cors.originRejected.${scope}`, `Origin not in allowlist: ${raw}`);
  }
  return allowed;
}
```

---

## LOW — Missing test coverage: `corsHeaders` rejection path does not assert `logError` is NOT called for allowed origins in `httpFunctions.test.js`

**Location:** `tests/httpFunctions.test.js`, new `options_health` describe block

**Issue:** The new `options_health` tests correctly assert the happy path (204 + correct headers) and the rejection path (403 + no ACAO header). However, neither test asserts anything about observability logging. When `options_health` receives a disallowed origin, the `_checkOrigin` call will fire `logError` — but the test never verifies this. This is only a LOW because the `cors.test.js` observability tests (once corrected) cover the `logError` contract directly. The `httpFunctions` test gap is a coverage nicety, not a safety hole.

---

## Confirmed clean — the following are well-handled

- `_checkOrigin` correctly distinguishes no-origin (same-origin / server-to-server) from rejected-origin. The `!allowed && raw` guard prevents log spam on legitimate server-side calls.
- `corsPreflight` continues to return 403 with an explicit body on rejection — the user-facing error is actionable and was not regressed.
- `logError` is not called inside a try/catch, so no exception from `logError` itself can be swallowed.
- No fallback-to-mock pattern is present. The failure mode (rejection) is explicit and correct.

The MEDIUM issues are worth a follow-up commit before merge. The `console.error` spy issue in particular means the observability guarantee being introduced is not fully tested.

PR #1101, opened 2026-04-18
fix(cf-9lp.1): GET /activeChallenges returns 503 on internal_error

DreadPirateRobertz:
Widens get_activeChallenges HTTP handler to return 503 Service Unavailable when the underlying webMethod returns { challenges: [], error: 'internal_error' } (the cf-tlt DB-failure shape). Preserves 200 OK for empty-but-authed and 429 for rate-limit. Body shape unchanged. Slice 1/4 for cf-9lp. Based on fix/cf-tlt-db-error-shape — auto-retargets to main when #1099 merges. Closes cf-jwz.

### DIFF
--- a/src/backend/http-functions.js
+++ b/src/backend/http-functions.js
@@ -1924,6 +1924,14 @@ export async function get_activeChallenges(request) {
       return response({ status: 429, body: json({ error: 'Rate limit exceeded' }), headers: jsonHeaders });
     }
 
+    // cf-9lp.1: surface cf-tlt's `internal_error` shape as 503 so generic REST
+    // consumers and monitoring probes don't treat a DB failure as success.
+    // The webMethod still preserves the { challenges: [] } field for mobile-app
+    // clients that read the body — shape is unchanged; only the status differs.
+    if (result.error === 'internal_error') {
+      return response({ status: 503, body: json(result), headers: jsonHeaders });
+    }
+
     return ok({ body: json(result), headers: jsonHeaders });
   } catch (err) {
     console.error(`HTTP function error (activeChallenges): memberId=${memberId || 'unknown'}:`, err);
--- a/tests/httpFunctions.test.js
+++ b/tests/httpFunctions.test.js
@@ -1662,6 +1662,30 @@ describe('get_activeChallenges', () => {
     const result = await get_activeChallenges(makeActiveChallengesRequest('mem-rl'));
     expect(result.status).toBe(429);
   });
+
+  // cf-9lp.1: when the webMethod returns { challenges: [], error: 'internal_error' }
+  // (the cf-tlt DB-failure shape), the HTTP endpoint MUST return 503, not 200 OK.
+  // Prior behavior was 200 OK with the error body — generic REST consumers and
+  // monitoring probes saw "success" when the backend had failed.
+  describe('cf-9lp.1 503 on internal_error', () => {
+    it('returns 503 when the webMethod surfaces error: "internal_error"', async () => {
+      __setMember({ _id: 'mem-db-fail' });
+      __setQueryError('Challenges', new Error('connection reset'));
+      const result = await get_activeChallenges(makeActiveChallengesRequest('mem-db-fail'));
+      expect(result.status).toBe(503);
+      const body = JSON.parse(result.body);
+      expect(body.error).toBe('internal_error');
+    });
+
+    it('preserves 200 OK for empty-but-authed (no error field)', async () => {
+      __setMember({ _id: 'mem-empty' });
+      const result = await get_activeChallenges(makeActiveChallengesRequest('mem-empty'));
+      expect(result.status).toBe(200);
+      const body = JSON.parse(result.body);
+      expect(body.error).toBeUndefined();
+      expect(body.challenges).toEqual([]);
+    });
+  });
 });
 
 // ── POST /_functions/challengeProgress ───────────────────────────────────────


DreadPirateRobertz (commented):

**Verdict: APPROVE** — small, clean diff. Nothing worth blocking on.

### Strengths
- One-line conditional with an unambiguous predicate (`result.error === 'internal_error'`) is easier to reason about than a truthy-check — and matches the convention convergence on PR #1099 (typo-guard wins over future-proofing). Good consistency.
- Two tests pull their weight: the 503 assertion proves the new branch fires; the 200-empty-authed test is a cheap regression guard ensuring the new branch doesn't over-trigger on the common `{ challenges: [] }` success path. Not redundant.
- Body preservation (`body: json(result)`) keeps mobile clients' shape contract intact — no split serialization paths.

### Judgment-call nits (non-blocking)
1. **Comment length vs. conditional length (`:1927-1930`).** A 4-line comment above a 2-line branch is slightly top-heavy. The rationale (cf-tlt shape, monitoring probes, body preservation) is valuable context that wouldn't fit in a commit-only note — I'd keep it. If you wanted to trim, the third sentence (\"The webMethod still preserves…\") could collapse to `// Body shape preserved; only status differs.` Leaving it verbose is defensible for a subtle status-code change.

2. **Describe block nesting (`tests/httpFunctions.test.js:1667`).** The nested `describe('cf-9lp.1 503 on internal_error')` inside `describe('get_activeChallenges')` is a minor bit of noise — the two tests could sit flat under the parent with `cf-9lp.1:` prefixes in their `it()` names. That said, the nested describe groups related assertions and gives a clean grep target for the slice; low-value to change either way.

3. **Pattern mismatch with RewardsTierWidget (cf-8qc).** RewardsTierWidget uses a truthy-check on `result.error` to surface any future error tokens (`forbidden`, `rate_limited`, etc.) as non-200. This PR uses strict equality. Flagging for awareness only — per PR #1099's convergence, the explicit equality was the deliberate choice here (typo-guard over future-proofing). If cf-9lp slices 2-4 add more error tokens, you'll want a small switch/set rather than stacking `===` checks; worth a note in the cf-9lp.2 story.

### No issues with
- Test setup (`__setQueryError`, `__setMember`) — reuses existing harness idioms.
- JSDoc/comments refer to bead IDs (cf-9lp.1, cf-tlt) — good traceability.
- No duplication with existing 429 branch; both follow the same `response({ status, body, headers })` shape.

Ship it.

DreadPirateRobertz (commented):
**silent-failure-hunter** review (dispatched by **radahn**, 5-agent PR review protocol)

## Verdict: **APPROVE** with one adjacent-hole callout

### Scope check — closes the cf-9lp.1 silent-class hole? YES

The new branch at `src/backend/http-functions.js:1927-1929`:

```js
if (result.error === 'internal_error') {
  return response({ status: 503, body: json(result), headers: jsonHeaders });
}
```

discriminates the cf-tlt DB-failure shape (`{ challenges: [], error: 'internal_error' }`) from the legitimate empty-authed shape (`{ challenges: [] }`). Status-code + body are now both distinguishable, which closes exactly the hole I flagged on PR #1099's cascade list (target #4). Generic REST consumers and monitoring probes that branch on `response.ok` / 5xx will now detect the failure; mobile-app clients that read `body.error` keep working (shape preserved).

### logError chain — intact

Verified on the PR branch `src/backend/gamificationCore.web.js:830`: the catch block calls `logError('getActiveChallenges — failed for member ${memberId}', err)` **before** returning `{ challenges: [], error: 'internal_error' }`. Original exception + stack still hits Sentry. The HTTP layer correctly re-surfaces it as 503 without re-logging (avoids double-Sentry noise).

### New silent-failure paths introduced by this PR? NONE

- `result` null/undefined: `result.error === 'internal_error'` would throw `TypeError` → outer try/catch at `:1938-1941` catches it → returns 500. Not ideal (console.error not logError — see below) but not a regression from this PR.
- Body shape: `json(result)` passes through the same `{ challenges: [], error: 'internal_error' }` object. No masking.
- Status classification: 503 is correct (Service Unavailable = transient backend failure, retryable). Not 500 (which would imply handler bug).

### Outer catch-all at `:1938-1941` — still swallows something meaningful?

```js
} catch (err) {
  console.error(`HTTP function error (activeChallenges): memberId=${memberId || 'unknown'}:`, err);
  return serverError({ body: json({ error: 'Internal server error' }), headers: jsonHeaders });
}
```

Uses `console.error`, NOT `logError` — so Sentry does NOT receive unexpected handler errors (TypeErrors from null `result`, response() throws, etc). This is **pre-existing** from the cf-tlt slice and out of scope for this PR, but flagging for the cascade: when you land cf-n3e/cf-0z4/cf-3be, consider swapping these three `console.error` sites in `get_activeChallenges` to `logError` with a stable errorId from `constants/errorIds.ts`. Same applies to the inner `getMember()` catch at `:1920-1923`.

### Adjacent hole in same handler (NOT a blocker — file for follow-up)

`gamificationCore.web.js:741` returns `{ challenges: [], error: 'auth_required' }` for the "Velo let an unauth request through" branch. The HTTP handler at `:1919` returns `200 OK` for this shape — arguably a worse silent failure than the DB one (caller sees success-with-empty when auth is actually broken). Not in the cascade list you shared; probably belongs as a new issue or an extension of cf-9lp.2. Mobile contract would accept a 401/403 remap cleanly since the body already carries `error: 'auth_required'`.

### Test coverage

The two new tests cover the positive case (503 on DB failure) and the regression case (200 preserved on empty-authed). Good discrimination — exactly the pair needed to prevent this silent-failure from regressing. No test for `result` being null/undefined, but the outer catch handles it and a test would just confirm pre-existing behavior.

### Follow-ups noted (explicitly out of scope for this PR)
- cf-n3e: `ChallengesDisplay.js` consumer
- cf-0z4: `challengeDiscovery.js` consumer
- cf-3be: `getActiveChallengeOfWeek` at http-functions.js:1360

Ship it. The slice is tight, the discriminator is correct, and the Sentry chain is preserved. The broader console.error → logError cleanup and the `auth_required` 200-OK hole are separate beads.

DreadPirateRobertz (commented):
**Agent:** code-reviewer
**Protocol:** 5-agent PR review (mayor directive 2026-03-08)

---

## Verdict: APPROVE

Tight, well-scoped slice. The change does exactly what the commit message and PR body promise, and the guard is defensive enough for the settled cf-tlt shape.

### Convention fit — PASS
- `response({ status: 503, ... })` at `src/backend/http-functions.js:1932` is the correct pattern. `wix-http-functions` (imported at line 3) exposes `ok/notFound/serverError/forbidden/badRequest/unauthorized/response` — there is no named helper for 503. `serverError()` would be wrong (it's 500, not 503; and 503 "Service Unavailable" is the semantically correct code for a transient backend/DB outage so monitoring probes retry rather than page).
- Mirrors the nearest-neighbor 429 branch at `:1926` which also uses `response({ status: 429, ... })`. Consistent local convention.

### Guard correctness — PASS
- `result.error === 'internal_error'` at `:1931` is strict-equal against the single discriminant PR #1099 converged on. The source of truth `getActiveChallenges` webMethod at `src/backend/gamificationCore.web.js:831` returns exactly `{ challenges: [], error: 'internal_error' }` on DB failure.
- No risk of misfiring on the other error shape in that webMethod (`auth_required` at `gamificationCore.web.js:740`) — the HTTP handler pre-validates `member` and `member._id === memberId` before calling the webMethod, so `auth_required` is unreachable on this path.
- Branch is placed **after** the 429 check and **before** `ok()` — rate-limit responses still win, empty-but-authed still returns 200. No path interference.

### Test quality — PASS
- `tests/httpFunctions.test.js:1672-1678` — 503 case uses `__setQueryError('Challenges', ...)` which is the intended mechanism for driving the webMethod into its `internal_error` branch. Good integration-level coverage (not mocking the webMethod return directly — real error flows through).
- `tests/httpFunctions.test.js:1681-1687` — the "preserves 200 OK for empty-but-authed" test is slightly redundant with the pre-existing empty-case at `:1637`, BUT the explicit `expect(body.error).toBeUndefined()` assertion is the right regression pin for this specific branch (locks in that we don't trip on falsy `error`). Worth keeping.
- RED-first discipline explicitly documented in commit `2c9f495b`: "RED: expect 503 → got 200. GREEN after adding the branch."

### Hidden side effects — NONE
- No mutation of `result` before return (`json(result)` is called directly — body shape preserved as promised).
- `Cache-Control: no-store` header is reused via `jsonHeaders` — appropriate for a transient-failure response (no stale 503 caching).
- Body still carries `{ challenges: [], error: 'internal_error' }` for mobile clients reading the body, which was explicitly the point of the cf-tlt convention.

### Minor observations (non-blocking)
- **Scoping note:** this is slice 1/4 per the cf-9lp cascade. `post_challengeProgress` and `get_leaderboard` in the same file likely need sibling treatment — that's explicitly deferred and fine.
- **Comment at `:1927-1930` is well-crafted** — links cf-9lp.1 → cf-tlt and explains *why* (monitoring probes vs body-reading mobile clients). This is the documentation pattern more handlers should adopt.

### Recommendation
Merge after #1099 lands (auto-retarget will handle). No changes requested.

DreadPirateRobertz (commented):

**Verdict: APPROVE with one nit worth addressing before merge.**

Test coverage for the new 503 branch is proportional to the change (one new `if` → one new happy-path test + one preservation test). Isolation is clean (`resetData()` in `beforeEach` clears `_queryErrors`), body shape is locked alongside the status, and the branch is genuinely exercised (cf-tlt base at `gamificationCore.web.js:831` now returns `{ challenges: [], error: 'internal_error' }`, which is exactly what `__setQueryError('Challenges', …)` triggers).

### Checklist findings

1. **Status + body lock-in — PASS.** Test 1 asserts `status === 503` and `body.error === 'internal_error'`. Could be tightened further by also asserting `body.challenges` is present and equal to `[]` — the whole *point* of cf-9lp.1 is "status changes, body shape doesn't." Without that assertion, a future refactor that drops `challenges` from the error body would silently pass. Rating: 6/10.

2. **Preservation regression for 429 — PASS (no new test needed).** The existing test at `:1657-1663` still exercises the rate-limit path, and the new `if (result.error === 'internal_error')` branch sits *after* the `result.status === 429` check, so there is no interaction risk. Don't add a redundant test.

3. **Negative / adversarial `result.error` values — N/A.** Verified the webMethod (`gamificationCore.web.js` lines 726-815) only returns one of three shapes: `{ challenges: [] }`, `{ status: 429, error: 'Rate limit exceeded' }`, or the new `{ challenges: [], error: 'internal_error' }`. `auth_required` / `forbidden` are handled upstream in the HTTP wrapper before the webMethod runs, so the 503 branch has no adversarial-string surface to cover. No test gap.

4. **Test isolation — PASS.** `resetData()` calls `__reset()` in `wix-data.js:18` which clears `_queryErrors = {}`. No leak across tests.

5. **Double-fault / outer-catch (webMethod throws instead of returns) — GAP (optional).** The outer `catch` at `http-functions.js:1928-1931` returning 500 is currently **untested** for `get_activeChallenges`. The webMethod's own `try/catch` at `gamificationCore.web.js:825` swallows throws and returns the `internal_error` shape, so in practice the outer catch only fires if something *outside* the webMethod's try (e.g. an import-time issue, or a throw from the rate-limit map) goes wrong — very narrow. Not blocking for this PR (it's pre-existing), but worth a single `vi.mocked(_getActiveChallengesWebMethod).mockRejectedValueOnce(...)` test in a follow-up to lock in the 500 contract symmetrically with the 503 one. Rating: 4/10.

### Suggested tightening (optional, non-blocking)

```js
it('returns 503 when the webMethod surfaces error: "internal_error"', async () => {
  __setMember({ _id: 'mem-db-fail' });
  __setQueryError('Challenges', new Error('connection reset'));
  const result = await get_activeChallenges(makeActiveChallengesRequest('mem-db-fail'));
  expect(result.status).toBe(503);
  const body = JSON.parse(result.body);
  expect(body.error).toBe('internal_error');
  expect(body.challenges).toEqual([]);  // ← lock the preservation contract
});
```

### Positive observations

- DAMP test names — `'returns 503 when the webMethod surfaces error: "internal_error"'` reads like the spec it enforces.
- The `describe('cf-9lp.1 503 on internal_error')` bead-scoped block is easy to grep/revert if the story is ever rolled back.
- Second test is a deliberate *regression guard* against the naive "return 503 whenever `result.error` is truthy" mis-implementation — this is the exact kind of test that pays for itself. Good instinct.

**Verdict: APPROVE.** One nit (add `expect(body.challenges).toEqual([])` to the 503 test) is worth doing before merge but is not blocking.

— pr-test-analyzer (dispatched by radahn, cfutons/crew, per mayor directive 2026-03-08)

DreadPirateRobertz (commented):
## comment-analyzer review (dispatched by radahn)

**Verdict: REQUEST_CHANGES** — one load-bearing claim in the src comment is factually wrong and will mislead the next maintainer.

### Critical issue

**File: `src/backend/http-functions.js:1927-1930`** (new block)

> `// The webMethod still preserves the { challenges: [] } field for mobile-app`
> `// clients that read the body — shape is unchanged; only the status differs.`

This is incorrect and misleading. The only known mobile-app consumer (`cfutons_mobile/src/hooks/useActiveChallenges.ts` → `wixClient.callFunction` → `rawRequest`) does NOT read the body on non-2xx responses. At `cfutons_mobile/src/services/wix/wixClient.ts:1324`:

```ts
if (!response.ok) {
  const errorBody = await response.json().catch(() => ({}));
  throw new WixApiError(... response.status ...);
}
```

A 503 throws `WixApiError` and the `challenges: []` field is discarded. Worse, `isRetryableError` (line 1339-1352) explicitly retries 5xx, so the mobile client will now retry on DB failure (previously it saw `.catch` only on network errors — the 200-OK silent behavior hid DB failures entirely from `.catch`). The behavioral shift is real and defensible, but the comment's justification — "preserved for mobile-app clients that read the body" — describes a consumer that does not exist. If that's the intended contract, it's currently aspirational, not actual.

**Suggested rewrite:**
```js
// cf-9lp.1: surface cf-tlt's `internal_error` shape as 503 so REST consumers and
// uptime probes don't treat a DB failure as success. Body is forwarded unchanged
// ({ challenges: [], error: 'internal_error' }) for any future consumer that wants
// to distinguish db_error from auth/rate-limit without re-parsing the status line.
// Mobile clients (cfutons_mobile useActiveChallenges) currently treat 5xx as thrown
// errors with retry — this is the intended migration from silent 200-OK.
```

That framing is honest about who actually reads the body today (nobody) and what the behavior shift means for mobile (retries now happen on DB failure — intentional).

### Improvement opportunities

**File: `tests/httpFunctions.test.js:1665-1668`**

The "Prior behavior was 200 OK with the error body — generic REST consumers and monitoring probes saw 'success' when the backend had failed" explanation is a genuinely useful WHY comment and earns its keep. Minor nit: "prior behavior" is already strictly accurate only against the cf-tlt base branch (not main), so in 6 months when cf-tlt has merged and been forgotten, "prior" will be ambiguous. Consider anchoring to the bead: `// Before cf-9lp.1, the HTTP endpoint returned 200 OK even when...`.

### Bead reference durability

`cf-9lp.1` and `cf-tlt` references are consistent with established prior art in this file and codebase (`cf-amx`, `cf-fsm`, `cf-u30i`, `cf-8onx`, `cf-bu2`, `cf-7sr`, etc. — 20+ bead-tagged comments). They are load-bearing for git archaeology via `bd show <id>` and match the team convention. Not rot-prone.

### Terminology drift check

No drift. "internal_error" shape name matches commit `bcfe4489 review(cf-tlt): use 'internal_error' convention` on the base branch. "503 Service Unavailable" terminology matches PR body and test name. The earlier commit `233741c4` used `'db_error'` — make sure no stale `db_error` references linger in docs, but that's outside this PR's diff.

### Summary

- 1 Critical: src comment inverts reality re: mobile-app body consumption — rewrite recommended before merge.
- 1 Minor: test comment's "prior" anchor will age; pin to bead ID.
- 0 rot-prone bead refs; terminology is consistent.

Recommend fixing the src comment, then approve.

— comment-analyzer (dispatched by radahn, 5-agent PR review protocol, mayor directive 2026-03-08)

DreadPirateRobertz (commented):
## 5-Agent Peer Review Synthesis (radahn dispatched, mayor directive 2026-03-08)

**Overall verdict: APPROVE** after addressing comment-analyzer's REQUEST_CHANGES in commit `82807c18`.

### Scorecard

| Agent | Verdict | Outcome |
|-------|---------|---------|
| code-reviewer | APPROVE | ✅ No changes. Confirmed `response({status:503})` convention matches nearest-neighbor 429 branch; `serverError()` would be wrong (500, not 503). |
| silent-failure-hunter | APPROVE | ✅ No changes. Flagged `auth_required` at `gamificationCore.web.js:740` as adjacent concern — but HTTP handler short-circuits at `:1914` before the webMethod call (`if (!member) return unauthorized()`), so that path is unreachable here. |
| code-simplifier | APPROVE | ✅ No changes. Non-blocking note: future slices cf-9lp.2-4 may prefer switch/set over stacking `===` — deferred to those PRs. |
| comment-analyzer | **REQUEST_CHANGES** | ✅ Addressed: my comment's "body preserved for mobile-app clients" justification was inverted. Mobile client (`cfutons_mobile wixClient.rawRequest`) throws on non-2xx AND retries 5xx — so the real effect is transient DB blips now trigger retries instead of silent empty lists. Rewrote comment to describe the actual behavioral shift (desirable UX improvement) and kept body-emission as diagnostic fallback. |
| pr-test-analyzer | APPROVE + nit | ✅ Addressed: added `expect(body.challenges).toEqual([])` to the 503 test to pin the preserved-field regression. |

### Post-review fix commit (`82807c18`)

- `src/backend/http-functions.js:1929-1939` — comment rewritten to describe real mobile-client retry behavior
- `tests/httpFunctions.test.js:1674` — added regression pin on `body.challenges`
- 8/8 `get_activeChallenges` tests still green locally

### Scope reaffirmed

cf-9lp.1 is the first of 4 cascade slices (cf-jwz/n3e/0z4/3be). Based on `fix/cf-tlt-db-error-shape` (PR #1099) — auto-retargets to main when #1099 merges.

The `auth_required`-via-webMethod silent-path silent-failure-hunter flagged is unreachable through this HTTP endpoint (pre-webMethod guard at :1914), but worth a note for backend-to-backend callers. If melania wants that chased, it's a standalone bead not in cf-9lp.

---

Ready for melania's gate.

PR #1103, opened 2026-04-18
feat(cf-nq7.1): Wix CMS → cfw HMAC revalidate webhook

DreadPirateRobertz:
## Summary

- Adds `_postRevalidateWebhook(body)` helper to `events.js` — HMAC-signs a JSON payload and POSTs to `VERCEL_REVALIDATE_URL` on Wix Secrets Manager
- Wires into `wixStores_onProductCreated` and `wixStores_onProductUpdated` (products collection)
- Adds `wixBlog_onPostPublished` handler (blog/Posts collection)
- Adds `wixData_onDataItemUpdated` handler (any CMS collection, skips if no collectionId)
- Best-effort: errors are caught and `console.warn`'d — a missed revalidate means stale cache, not a broken event handler

Requires Wix Secrets: `VERCEL_REVALIDATE_URL`, `WIX_WEBHOOK_SECRET`

cfw route already exists: `src/app/api/revalidate/route.ts` (PR #1, merged)

## Test plan
- [x] 9 tests in `tests/events-revalidate.test.js` — all green
- [x] HMAC signature matches `sha256=<hex>` with correct secret
- [x] Skips `fetch` when `VERCEL_REVALIDATE_URL` is empty string
- [x] Does not throw on network error (non-fatal path)
- [x] `console.warn` (not `console.error`) on non-ok response status
- [x] Blog post publish → `blog/Posts` collectionId
- [x] CMS item update → passes through `collectionId`; skips when missing

### DIFF
--- a/src/backend/events.js
+++ b/src/backend/events.js
@@ -445,6 +445,7 @@ export async function wixStores_onProductUpdated(event) {
   const newPrice = product.price?.amount ?? product.price ?? null;
   const oldPrice = previous.price?.amount ?? previous.price ?? null;
 
+  // Revalidate on every product update (name, image, stock, price) — intentional.
   await _postRevalidateWebhook({ collectionId: 'products', itemId: productId, eventType: 'onProductUpdated' });
 
   // Only trigger price-drop content orchestration on actual price drops
--- a/src/backend/events.js
+++ b/src/backend/events.js
@@ -23,6 +23,34 @@
 import wixData from 'wix-data';
 import { sanitize } from 'backend/utils/sanitize';
 
+// ── CWF ISR Revalidate Helper ─────────────────────────────────────────
+
+async function _postRevalidateWebhook(body) {
+  try {
+    const { getSecret } = await import('wix-secrets-backend');
+    const [url, secret] = await Promise.all([
+      getSecret('VERCEL_REVALIDATE_URL'),
+      getSecret('WIX_WEBHOOK_SECRET'),
+    ]);
+    if (!url || !secret) return;
+
+    const payload = JSON.stringify(body);
+    const { createHmac } = await import('node:crypto');
+    const sig = 'sha256=' + createHmac('sha256', secret).update(payload).digest('hex');
+
+    const res = await fetch(url, {
+      method: 'POST',
+      headers: { 'content-type': 'application/json', 'x-wix-signature': sig },
+      body: payload,
+    });
+    if (!res.ok) {
+      console.warn('[events] revalidate webhook returned', res.status);
+    }
+  } catch (err) {
+    console.warn('[events] revalidate webhook failed (non-fatal):', err?.message ?? err);
+  }
+}
+
 // ── Dead-Letter Queue Helper ─────────────────────────────────────────
 
 /**
@@ -376,6 +404,8 @@ export async function wixStores_onProductCreated(event) {
     return;
   }
 
+  await _postRevalidateWebhook({ collectionId: 'products', itemId: productId, eventType: 'onProductCreated' });
+
   try {
     const { getSecret } = await import('wix-secrets-backend');
     const eventSecret = await getSecret('CONTENT_EVENT_KEY');
@@ -415,7 +445,9 @@ export async function wixStores_onProductUpdated(event) {
   const newPrice = product.price?.amount ?? product.price ?? null;
   const oldPrice = previous.price?.amount ?? previous.price ?? null;
 
-  // Only trigger on actual price drops (not increases, not identical)
+  await _postRevalidateWebhook({ collectionId: 'products', itemId: productId, eventType: 'onProductUpdated' });
+
+  // Only trigger price-drop content orchestration on actual price drops
   if (oldPrice == null || newPrice == null || newPrice >= oldPrice) return;
 
   try {
@@ -563,4 +595,27 @@ export async function wixMembers_onMemberUpdated(event) {
       impact: `Birthday fields not synced for member ${memberId} — birthday cron may skip this member`,
     });
   }
+}
+
+// ── CWF ISR Revalidate Hooks ─────────────────────────────────────────
+
+export async function wixBlog_onPostPublished(event) {
+  const post = event.entity || event;
+  const postId = post._id || post.id || '';
+  await _postRevalidateWebhook({
+    collectionId: 'blog/Posts',
+    itemId: postId || undefined,
+    eventType: 'onPostPublished',
+  });
+}
+
+export async function wixData_onDataItemUpdated(event) {
+  const collectionId = event.collectionId || event.dataCollectionId || '';
+  const itemId = event.item?._id || event.itemId || '';
+  if (!collectionId) return;
+  await _postRevalidateWebhook({
+    collectionId,
+    itemId: itemId || undefined,
+    eventType: 'onDataItemUpdated',
+  });
 }
--- a/tests/events-revalidate.test.js
+++ b/tests/events-revalidate.test.js
@@ -1,0 +1,168 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { createHmac } from 'node:crypto';
+
+// ── Wix platform mock ────────────────────────────────────────────────
+
+const mockGetSecret = vi.fn();
+vi.mock('wix-secrets-backend', () => ({ getSecret: (...a) => mockGetSecret(...a) }));
+vi.mock('wix-data', () => ({ default: { insert: vi.fn(), get: vi.fn(), update: vi.fn() } }));
+vi.mock('backend/utils/sanitize', () => ({ sanitize: (v) => v }));
+vi.mock('backend/contentOrchestrator.web', () => ({ triggerEventOrchestration: vi.fn() }));
+vi.mock('backend/emailAutomation.web', () => ({ triggerRestockNotifications: vi.fn() }));
+
+// ── fetch mock ────────────────────────────────────────────────────────
+
+const mockFetch = vi.fn();
+vi.stubGlobal('fetch', mockFetch);
+
+const TEST_URL = 'https://cfw.example.com/api/revalidate';
+const TEST_SECRET = 'test-hmac-secret';
+
+function expectedSig(body) {
+  return 'sha256=' + createHmac('sha256', TEST_SECRET).update(body).digest('hex');
+}
+
+// ── helpers ───────────────────────────────────────────────────────────
+
+function okResponse() {
+  return { ok: true, status: 200 };
+}
+
+// ── tests ─────────────────────────────────────────────────────────────
+
+describe('_postRevalidateWebhook (via wixStores_onProductUpdated)', () => {
+  beforeEach(() => {
+    vi.resetModules();
+    mockFetch.mockClear();
+    mockFetch.mockResolvedValue(okResponse());
+    mockGetSecret.mockImplementation((key) => {
+      if (key === 'VERCEL_REVALIDATE_URL') return Promise.resolve(TEST_URL);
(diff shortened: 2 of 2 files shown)


DreadPirateRobertz:
## code-reviewer

**Agent:** code-reviewer | **Dispatched by:** godfrey

Reviewed `feat(cf-nq7.1): Wix CMS → cfw HMAC revalidate webhook`

### Findings

**MEDIUM — Secret fetch on every event call (performance)**
`_postRevalidateWebhook` calls `getSecret` twice on every invocation (Promise.all for URL + secret). Wix Secrets Manager calls are network round-trips. In high-frequency event handlers like `wixStores_onProductUpdated` this doubles the latency on every product save. Consider module-level caching with a `let _cachedUrl`, `_cachedSecret` pattern (set once, reuse on subsequent calls), since secrets don't change at runtime.

**MEDIUM — `wixStores_onProductUpdated` revalidate fires before the price-drop guard**
The `_postRevalidateWebhook` call (line +448) is placed *before* the `if (oldPrice == null || newPrice == null || newPrice >= oldPrice) return` guard. This means every product update — not just price drops — triggers ISR revalidation on the products page. The PR description says the event fires on product update, which is correct behaviour, but the placement is likely intentional. Confirm: should revalidation be unconditional on any update, or should it only fire on price drops? The original function's `onProductUpdated` semantics were price-drop-only. If unconditional revalidation is desired, the comment should say so explicitly to avoid future confusion.

**LOW — `import('node:crypto')` dynamic import inside hot path**
`createHmac` is imported dynamically inside `_postRevalidateWebhook`. In Node.js this is cached after first call, but the dynamic import syntax incurs microtask overhead on every call. Use a static top-level import for `node:crypto` since this module runs in a Wix backend environment that supports it. The `wix-secrets-backend` import staying dynamic is fine (Wix quirk), but `node:crypto` is a built-in that can be statically imported.

**LOW — `postId || undefined` produces `undefined` in JSON**
In `wixBlog_onPostPublished`: `itemId: postId || undefined`. When serialized with `JSON.stringify`, a key with value `undefined` is *omitted* from the output entirely. The cfw revalidate route should be aware that `itemId` may be absent. The test on line 148 (`expect(parsed.itemId).toBeUndefined()`) validates this behaviour, which is intentional — just ensure the receiver handles missing `itemId` gracefully.

**INFO — No retry or backoff**
Best-effort is the stated design and is acceptable. Just confirm there is no stale-cache SLA requirement; if there is, a single retry on 5xx would be worth adding.

### Verdict
LGTM with minor concerns. The secrets caching and placement of the revalidate call relative to the price-drop guard are the most worth addressing before merge.

DreadPirateRobertz:
## silent-failure-hunter

**Agent:** silent-failure-hunter | **Dispatched by:** godfrey

Reviewed `feat(cf-nq7.1): Wix CMS → cfw HMAC revalidate webhook`

### Findings

**PASS — Top-level catch in `_postRevalidateWebhook`**
The entire function body is wrapped in try/catch, and errors are `console.warn`'d rather than re-thrown. The non-fatal intent is explicit and correctly implemented. Network failures, secret fetch failures, and non-ok HTTP responses are all handled.

**PASS — `!url || !secret` guard before fetch**
Empty/null secrets short-circuit the function silently, which is correct for the "opt-in via secrets" design.

**MEDIUM — `wixBlog_onPostPublished` and `wixData_onDataItemUpdated` are fire-and-forget at the handler level with no surrounding try/catch in the handler itself**
Both new exported handlers are *only* `await _postRevalidateWebhook()` with no try/catch wrapper of their own. This is fine because `_postRevalidateWebhook` itself swallows all errors — but it means if someone adds code after the revalidate call in future, they may not notice there is no error boundary. Low risk as-is; worth noting for future reviewers.

**LOW — `wixStores_onProductCreated` revalidate call is placed before the existing try/catch block**
If `_postRevalidateWebhook` somehow throws despite its internal catch (e.g., a synchronous throw before the try block — impossible in current code but a maintenance hazard), it would abort `onProductCreated` before the existing event logic runs. The revalidate call should ideally be fire-and-forget (`void _postRevalidateWebhook(...)`) or placed after the main event handler logic to prevent it from blocking critical paths.

**LOW — `console.warn` on non-ok response does not include response body**
`console.warn('[events] revalidate webhook returned', res.status)` logs the status code but not the response body. Debugging a 401 (wrong secret) vs 404 (wrong URL) vs 500 (cfw crash) all look identical in logs. Adding `res.text().catch(() => ''))` to log a snippet of the body would meaningfully improve debuggability with minimal overhead.

**INFO — No timeout on `fetch`**
Wix backend functions have execution time limits. A hung cfw endpoint (e.g., Vercel cold start) could cause the fetch to block for 30s+. Consider `AbortSignal.timeout(5000)` as a safety net.

### Verdict
No true silent failures found — the error handling is solid. The fire-and-forget placement risk and missing fetch timeout are worth addressing.

DreadPirateRobertz:
## code-simplifier

**Agent:** code-simplifier | **Dispatched by:** godfrey

Reviewed `feat(cf-nq7.1): Wix CMS → cfw HMAC revalidate webhook`

### Findings

**MEDIUM — `|| undefined` pattern is redundant in two places**

In `wixBlog_onPostPublished`:
```js
itemId: postId || undefined,
```
And in `wixData_onDataItemUpdated`:
```js
itemId: itemId || undefined,
```
Since `postId` and `itemId` are initialized to `''` when not found, using `|| undefined` converts the empty string to `undefined`, which JSON.stringify then *omits*. This is a valid technique but the intent is obscured. Simpler and more explicit: use `itemId: postId || null` (null is preserved in JSON and clearly signals "no ID"), or add a comment. Alternatively use `itemId: postId ? postId : undefined` to make the conditional explicit.

**LOW — `event.entity || event` entity extraction is inconsistent across handlers**

In `wixBlog_onPostPublished`:
```js
const post = event.entity || event;
const postId = post._id || post.id || '';
```
But in `wixData_onDataItemUpdated`:
```js
const collectionId = event.collectionId || event.dataCollectionId || '';
const itemId = event.item?._id || event.itemId || '';
```
There are three different extraction patterns used across the 4 call sites. A tiny helper or consistent destructuring pattern would reduce the variance and make it easier to add new handlers.

**LOW — Double dynamic import of `wix-secrets-backend`**
`getSecret` is imported with `await import('wix-secrets-backend')` inside `_postRevalidateWebhook`. This pattern already exists in the pre-existing event handlers (onProductCreated, onProductUpdated). Module-level caching handles the repeated dynamic imports fine, but it adds noise. A single top-level import or a shared `getSecrets()` utility would be cleaner — though this is a pre-existing pattern, not introduced by this PR.

**INFO — `_postRevalidateWebhook` name prefixed with `_` to signal private**
This is a reasonable convention for a module-level private function in JS. No change needed, just confirming the convention is intentional.

### Verdict
Clean implementation overall. The `|| undefined` pattern is the one thing worth making more explicit. No over-engineering detected.

DreadPirateRobertz:
## comment-analyzer

**Agent:** comment-analyzer | **Dispatched by:** godfrey

Reviewed `feat(cf-nq7.1): Wix CMS → cfw HMAC revalidate webhook`

### Findings

**PASS — Section header comment accurate**
`// ── CWF ISR Revalidate Helper ─────────────────────────────────────────` correctly identifies the block and its purpose (ISR = Incremental Static Regeneration, CWF = carolina-futons-web).

**PASS — Updated inline comment on price-drop block**
The diff changes `// Only trigger on actual price drops (not increases, not identical)` to `// Only trigger price-drop content orchestration on actual price drops`. The new comment is more precise — specifying that it's the *content orchestration* trigger (not the revalidate webhook) that is price-drop-gated. This is accurate given that the revalidate call is now above this guard.

**PASS — `// ── CWF ISR Revalidate Hooks ─────────────────────────────────────────` section header at end of file**
Matches the new exported functions below it. Accurate.

**MEDIUM — No JSDoc on `_postRevalidateWebhook`**
The function has no `@param` or description comment. Given it accepts a generic `body` object, a brief JSDoc noting the expected shape (`{ collectionId, itemId?, eventType }`) would help future maintainers add new call sites correctly without reading all existing usages.

**LOW — PR description says "Best-effort: errors are caught and `console.warn`'d" but source has no comment explaining this intent**
The non-fatal / best-effort design is important — a missed revalidate is benign (stale cache) but the code doesn't say so. Adding a one-line comment above the catch block like `// Non-fatal: a missed revalidate means stale cache, not broken event flow` would preserve this intent for future readers who might be tempted to escalate the error level.

**INFO — `// ── Dead-Letter Queue Helper` comment is preserved correctly, no drift introduced**
The existing section header is untouched. No comment/implementation drift detected in surrounding code.

### Verdict
Comments are accurate and not misleading. The missing JSDoc on `_postRevalidateWebhook` and the undocumented best-effort intent are worth a quick addition.

DreadPirateRobertz:
## pr-test-analyzer

**Agent:** pr-test-analyzer | **Dispatched by:** godfrey

Reviewed `feat(cf-nq7.1): Wix CMS → cfw HMAC revalidate webhook`

### Coverage Assessment: 9 tests across 3 describe blocks

**PASS — Core happy path covered**
HMAC signature correctness is tested with a reference `expectedSig()` helper that independently computes sha256. The URL, method, headers, and body shape are all asserted. This is the most important test and it's thorough.

**PASS — Empty URL guard tested**
Test: "does NOT call fetch when VERCEL_REVALIDATE_URL is empty" — confirms the `!url || !secret` short-circuit works.

**PASS — Network error non-fatal path tested**
Test: "does not throw when fetch rejects (non-fatal)" — confirms the outer catch swallows the error.

**PASS — Non-ok HTTP response warning tested**
Test: "logs a warning (not error) when fetch returns non-ok" — checks `console.warn` is called with status 500.

**MEDIUM — Missing: WIX_WEBHOOK_SECRET empty guard**
The `!url || !secret` guard fires when either value is falsy. The "empty URL" case is tested but "empty secret" is not. These are symmetric conditions and both should be tested. Add: `mockGetSecret` returns `''` for `WIX_WEBHOOK_SECRET`, expect `mockFetch` not called.

**MEDIUM — Missing: `wixStores_onProductCreated` — fetch NOT called when productId is missing**
The existing handler has an early `if (!productId) return` guard at the top. When productId is missing, execution returns before reaching the revalidate call — so it's covered indirectly — but there's no explicit test confirming revalidation is skipped in this case. Worth adding for clarity.

**LOW — `vi.resetModules()` in every `beforeEach` but modules are only re-imported inside test bodies**
The `vi.resetModules()` call clears the module registry before each test, then each test does `await import('../src/backend/events.js')`. This pattern is correct for dynamic mock reconfiguration but it means every test re-executes the full module init. This is intentional and correct; just noting it for readers unfamiliar with the pattern.

**LOW — `wixData_onDataItemUpdated` uses `event.dataCollectionId` fallback but no test covers it**
The implementation has `event.collectionId || event.dataCollectionId || ''`. There is no test for the `dataCollectionId` fallback path. Add a test with only `dataCollectionId` set to confirm the fallback works.

**LOW — Non-ok response test asserts `expect(warnSpy).toHaveBeenCalledWith(expect.stringContaining(...), 500)` but source code is `console.warn('[events] revalidate webhook returned', res.status)` — this matches, but if the format changes the test will silently pass with the wrong message**
Consider pinning the exact string or using `toHaveBeenCalledWith('[events] revalidate webhook returned', 500)` for stronger assertion.

**INFO — No test for `wixBlog_onPostPublished` using `event.id` (non-`_id` fallback)**
The implementation has `post._id || post.id || ''`. The `post.id` fallback path is untested.

### Verdict
9 tests cover the main paths well. The missing "empty secret" guard test and `dataCollectionId` fallback test are the highest-value gaps to close. Recommend adding 2-3 targeted tests before merge.

PR #1104, opened 2026-04-18
fix(cf-9lp.2): ChallengesDisplay surfaces internal_error instead of silent hide

DreadPirateRobertz:
## Summary

Slice 2 of 4 for cf-9lp cascade (surface cf-tlt's `internal_error` discriminant into user-facing UI). Closes cf-n3e.

`initChallengesDisplay` previously hid the section on both empty-but-authed AND on failure — users saw "no challenges" when the backend had actually failed. Now:

- Accepts optional 5th param `$challengesError` (error-state element)
- On `response.error` truthy OR fn reject: show `$challengesError`, hide section
- Empty-but-authed: hides section silently (unchanged)
- Truthy-check (not strict `=== 'internal_error'`) — mirrors cf-8qc pattern from RewardsTierWidget so future codes surface automatically
- `$challengesError` is optional; missing-element ref is caught and silently handled (pre-editor safe)

## Wix editor follow-up for Stilgar

Add `#challengesError` Box to Member Page in Wix Studio editor. Without it, behavior degrades gracefully to the pre-cf-9lp.2 silent-hide. Code is safe to ship now — the element just needs to materialize for the UX improvement to reach users.

## Test plan

- [x] 34/34 tests pass (28 existing + 6 new cf-9lp.2 tests)
- [x] RED-first: 4/6 new tests failed before fix
- [x] Covers: internal_error, truthy-check future-proof, fn rejects, missing-element tolerance, normal render, empty-authed

Related: cf-9lp umbrella, closes cf-n3e. Slices 3-4 (challengeDiscovery chip, getActiveChallengeOfWeek) stack after this.

Generated with Claude Code

### DIFF
--- a/src/pages/Member Page.js
+++ b/src/pages/Member Page.js
@@ -254,6 +254,7 @@ async function initChallengesSection() {
     getActiveChallenges,
     $w('#challengesSection'),
     $w('#challengesList'),
+    $w('#challengesError'),  // cf-9lp.2: optional error element; missing ref tolerated
   );
 }
 
@@ -563,7 +564,8 @@ async function initChallengesDisplaySection() {
     currentMember._id,
     getActiveChallenges,
     $w('#challengesSection'),
-    $w('#challengesList')
+    $w('#challengesList'),
+    $w('#challengesError'),  // cf-9lp.2: optional error element; missing ref tolerated
   );
 }
 
--- a/src/public/ChallengesDisplay.js
+++ b/src/public/ChallengesDisplay.js
@@ -138,26 +138,51 @@ export function updateChallengeProgress(challengeId, progressValue, targetCount,
  * Calls getActiveChallenges, hides section if empty, renders rail if challenges returned.
  * Wired in Member Page.js — not called directly from this module.
  *
+ * cf-9lp.2: a failure now surfaces a distinct error element instead of silently
+ * hiding the section (which used to be indistinguishable from "no challenges").
+ * The `$challengesError` element is optional — if it doesn't yet exist in the
+ * Wix editor, the try/catch around .show() swallows the ref error and behavior
+ * degrades gracefully to the pre-cf-9lp.2 silent-hide. Truthy-check on
+ * response.error (cf-8qc pattern) so future codes surface automatically.
+ *
  * @param {string}   memberId
  * @param {Function} getActiveChallengesFn  - webMethod reference (injected for testability)
  * @param {Object}   $challengesSection     - Outer container Box
  * @param {Object}   $challengesList        - Repeater element
+ * @param {Object}   [$challengesError]     - Optional error-state container Box
  * @returns {Promise<void>}
  */
-export async function initChallengesDisplay(memberId, getActiveChallengesFn, $challengesSection, $challengesList) {
-  try {
-    const response = await getActiveChallengesFn(memberId);
-    const challenges = response.challenges || [];
-
-    if (challenges.length === 0) {
-      $challengesSection.hide();
-      return;
+export async function initChallengesDisplay(memberId, getActiveChallengesFn, $challengesSection, $challengesList, $challengesError) {
+  const showError = () => {
+    if ($challengesError) {
+      try { $challengesError.show(); } catch {}
     }
+    try { $challengesSection.hide(); } catch {}
+  };
 
-    $challengesSection.show();
-    renderChallengesRail($challengesList, challenges);
-  } catch (err) {
-    // Non-critical — hide section on error
-    $challengesSection.hide();
+  let response;
+  try {
+    response = await getActiveChallengesFn(memberId);
+  } catch {
+    showError();
+    return;
   }
+
+  if (!response || response.error) {
+    showError();
+    return;
+  }
+
+  if ($challengesError) {
+    try { $challengesError.hide(); } catch {}
+  }
+
+  const challenges = response.challenges || [];
+  if (challenges.length === 0) {
+    $challengesSection.hide();
+    return;
+  }
+
+  $challengesSection.show();
+  renderChallengesRail($challengesList, challenges);
 }
--- a/src/public/ChallengesDisplay.js
+++ b/src/public/ChallengesDisplay.js
@@ -135,8 +135,9 @@ export function updateChallengeProgress(challengeId, progressValue, targetCount,
 
 /**
  * Initializes the challenges display section on page load.
- * Calls getActiveChallenges, hides section if empty, renders rail if challenges returned.
- * Wired in Member Page.js — not called directly from this module.
+ * Calls getActiveChallenges, then either: renders the rail (challenges returned),
+ * hides the section silently (empty-but-authed), or surfaces $challengesError
+ * (response.error truthy OR fn reject). Wired in Member Page.js.
  *
  * cf-9lp.2: a failure now surfaces a distinct error element instead of silently
  * hiding the section (which used to be indistinguishable from "no challenges").
--- a/tests/ChallengesDisplay.test.js
+++ b/tests/ChallengesDisplay.test.js
@@ -5,6 +5,7 @@ import {
   showCompletionToast,
   updateChallengeProgress,
   formatCountdown,
+  initChallengesDisplay,
 } from '../src/public/ChallengesDisplay.js';
 
 // ── Mock Wix $w element helpers ───────────────────────────────────────────────
@@ -272,5 +273,87 @@ describe('renderChallengeCard — countdown integration', () => {
     const card = makeCard();
     renderChallengeCard(card, makeChallenge({ expiresAt: expires }), now);
     expect(card.$expiresLabel.text).toBe('');
+  });
+});
+
+// cf-9lp.2: initChallengesDisplay must surface response.error and the catch-path
+// instead of silently hiding the section. Previously a DB failure (cf-tlt
+// `internal_error` shape) was indistinguishable from "no active challenges" —
+// user saw an empty section either way and we lost the signal at the user-facing
+// layer. Now: on error, show the error element (optional 5th param, tolerates
+// missing element) and hide the section. Empty path is preserved silently.
+describe('initChallengesDisplay — cf-9lp.2 error surfacing', () => {
+  function makeContainer() {
+    return { show: vi.fn(), hide: vi.fn() };
+  }
+  function makeRepeater() {
+    return {
+      _data: null, _onItemReady: null, show: vi.fn(), hide: vi.fn(),
+      set data(v) { this._data = v; }, get data() { return this._data; },
+      onItemReady(cb) { this._onItemReady = cb; },
+    };
+  }
+
+  it('shows $challengesError and hides section when response.error is "internal_error"', async () => {
+    const $section = makeContainer();
+    const $list = makeRepeater();
+    const $error = makeContainer();
+    const fn = vi.fn().mockResolvedValue({ challenges: [], error: 'internal_error' });
+    await initChallengesDisplay('mem-1', fn, $section, $list, $error);
+    expect($error.show).toHaveBeenCalled();
+    expect($section.hide).toHaveBeenCalled();
+    expect($list._data).toBeNull();
+  });
+
+  it('shows $challengesError for any truthy error (future-proof)', async () => {
+    const $section = makeContainer();
+    const $list = makeRepeater();
+    const $error = makeContainer();
+    const fn = vi.fn().mockResolvedValue({ challenges: [], error: 'some-future-code' });
+    await initChallengesDisplay('mem-1', fn, $section, $list, $error);
+    expect($error.show).toHaveBeenCalled();
+    expect($section.hide).toHaveBeenCalled();
+  });
+
+  it('shows $challengesError and hides section when fn rejects', async () => {
+    const $section = makeContainer();
+    const $list = makeRepeater();
+    const $error = makeContainer();
+    const fn = vi.fn().mockRejectedValue(new Error('network failure'));
+    await initChallengesDisplay('mem-1', fn, $section, $list, $error);
+    expect($error.show).toHaveBeenCalled();
+    expect($section.hide).toHaveBeenCalled();
+  });
+
+  it('tolerates missing $challengesError (pre-editor-update safety)', async () => {
+    const $section = makeContainer();
+    const $list = makeRepeater();
+    const fn = vi.fn().mockResolvedValue({ challenges: [], error: 'internal_error' });
+    // 5th arg omitted — must not throw.
+    await expect(initChallengesDisplay('mem-1', fn, $section, $list)).resolves.not.toThrow();
+    expect($section.hide).toHaveBeenCalled();
+  });
+
+  it('hides $challengesError on normal render with challenges', async () => {
+    const $section = makeContainer();
+    const $list = makeRepeater();
+    const $error = makeContainer();
+    const fn = vi.fn().mockResolvedValue({
+      challenges: [{ challengeId: 'ch-1', title: 't', description: 'd', targetCount: 1, rewardPoints: 10, expiresAt: '2099-01-01T00:00:00Z', progressValue: 0, completedAt: null }],
+    });
+    await initChallengesDisplay('mem-1', fn, $section, $list, $error);
+    expect($error.hide).toHaveBeenCalled();
+    expect($section.show).toHaveBeenCalled();
+    expect($list._data).toHaveLength(1);
+  });
+
+  it('hides section silently on empty-but-authed (no error field)', async () => {
+    const $section = makeContainer();
+    const $list = makeRepeater();
+    const $error = makeContainer();
+    const fn = vi.fn().mockResolvedValue({ challenges: [] });
+    await initChallengesDisplay('mem-1', fn, $section, $list, $error);
+    expect($section.hide).toHaveBeenCalled();
+    expect($error.show).not.toHaveBeenCalled();
   });
 });
--- a/tests/ChallengesDisplay.test.js
+++ b/tests/ChallengesDisplay.test.js
@@ -334,6 +334,16 @@ describe('initChallengesDisplay — cf-9lp.2 error surfacing', () => {
     expect($section.hide).toHaveBeenCalled();
   });
 
+  it('shows $challengesError when fn resolves to null (defensive null-guard)', async () => {
+    const $section = makeContainer();
+    const $list = makeRepeater();
+    const $error = makeContainer();
+    const fn = vi.fn().mockResolvedValue(null);
+    await initChallengesDisplay('mem-1', fn, $section, $list, $error);
+    expect($error.show).toHaveBeenCalled();
+    expect($section.hide).toHaveBeenCalled();
+  });
+
   it('hides $challengesError on normal render with challenges', async () => {
     const $section = makeContainer();
     const $list = makeRepeater();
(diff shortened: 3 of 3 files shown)


DreadPirateRobertz (commented):
**Reviewer**: code-reviewer (dispatched by radahn, 5-agent PR review protocol, mayor directive 2026-03-08)

**Verdict**: APPROVE

### Convention fit (5th-param vs opts pattern)
Correct. ChallengesDisplay's module header explicitly states *"No Wix imports — operates on injected \$element objects for testability"* — positional \$element refs ARE the established convention for this module. RewardsTierWidget uses a different injection style (`opts = {}` with internal `$w` resolution), but migrating ChallengesDisplay to match would be out-of-scope churn for a targeted fix slice. The added 5th param is consistent with the existing 3rd/4th params (`$challengesSection`, `$challengesList`).

### Graceful-degradation safety
Safe. The `try { $challengesError.show(); } catch {}` pattern mirrors RewardsTierWidget's per-element defensive wrapping. The `if ($challengesError)` guards prevent TypeError when the param is omitted entirely (pre-editor state); the inner try/catch handles the case where `$w('#challengesError')` returns a throwing stub in editor-before-element-exists state. Both cases are tested. No state leak — `response` is already resolved when `showError()` fires.

### Call site symmetry
Both Member Page.js call sites updated (lines ~254 and ~563). Both pass `$w('#challengesError')` with identical inline comment. Symmetric.

### Truthy-check vs strict equality
Deliberate and correct per cf-8qc widget-layer precedent. Backend retains strict `=== 'internal_error'` (cf-tlt). Widget layer accepts any truthy error so future backend error codes surface automatically without a widget redeploy. Forward-compatible.

### Test coverage
Strong. 6 tests cover: internal_error path, future-code truthy path, fn-reject catch path, missing-element tolerance, normal-render hides error element, empty-but-authed silent hide (regression guard for prior behavior). RED-first discipline confirmed per PR body (4/6 failed pre-fix).

Nit (non-blocking): the normal-render test asserts `$error.hide` was called but not that `$error.show` was NOT called — minor gap, already implied. Consider adding `expect(\$error.show).not.toHaveBeenCalled()` to the normal-render test for explicitness.

### Other observations
- `response.challenges || []` correctly defaults an undefined `challenges` field after the `!response || response.error` guard passes — no null-deref risk
- `response` is declared with `let` outside the try so it's reachable after the catch-return — correct scoping
- Empty-authed path preserved (hide section, no error shown) — unchanged semantics, as the cascade intends

Ship it. Stilgar's Wix-editor follow-up (adding `#challengesError` Box) is correctly called out in the PR body; code is safe to merge before the element materializes.

DreadPirateRobertz (commented):
**code-simplifier** review (dispatched by **radahn**, per mayor directive 2026-03-08)

Verdict: **APPROVE** with two optional simplification nits. Behavior is correct, tests are sound, safe to ship.

### Flow shape (showError closure + imperative response)

The `let response; try/catch-to-assign` + truthy `response.error` check is clear enough and matches the cf-8qc / RewardsTierWidget precedent. I would not flatten further — the closure localizes the show-error-hide-section pair used by both the reject and `response.error` branches, and the early-return style is easier to trace than a nested try/else.

### Repeated `if ($challengesError) { try { ... } catch {} }` (nit, optional)

The guard+try-catch appears 3x (2x inside `showError`, 1x on the success path for `.hide()`). Consider one tiny helper, local to the function:

```js
const toggleError = (visible) => {
  if (!$challengesError) return;
  try { visible ? $challengesError.show() : $challengesError.hide(); } catch {}
};
```

Then `showError` reduces to `toggleError(true); try { $challengesSection.hide(); } catch {}` and the success-path line becomes `toggleError(false)`. Three duplicated try/guard blocks collapse to one. Not a blocker — current code is readable, just slightly repetitive.

(Minor: no try/catch wraps `$challengesSection.hide/show` in the current code either — if we're defending against missing editor refs for `$challengesError`, applying the same defense to `$challengesSection` would be symmetric. The original pre-PR code did not guard it, so leaving as-is is defensible.)

### Test pulling-weight (per PR #1099 precedent — delete redundant when covered)

Five of six tests earn their keep: `fn rejects` (catch path), `missing-element` (optional param tolerance), `normal-render` (happy path + error.hide assertion), `empty-authed` (silent empty distinguished from error), and one of the two error-surface tests.

The two error-surface cases — `internal_error` and `truthy-any 'some-future-code'` — assert the **same two behaviors** (`$error.show` + `$section.hide`) with only the error-string value differing. Applying the PR #1099 judgment, collapse these into a single `it.each` or delete one:

```js
it.each(['internal_error', 'some-future-code'])(
  'shows $challengesError and hides section when response.error=%s',
  async (errorCode) => { /* ... */ }
);
```

This preserves the future-proof intent (truthy-check, not strict equality) without two nearly-identical test bodies. Optional.

### Dead code / behavioral

None. The explicit `if (!response || response.error)` is a correct behavior tightening — previously a null `response` would throw on `.challenges` and fall into the catch; now it's an explicit error branch. Good.

### Summary

Ship it. Both nits are style-level and orthogonal to the cf-9lp.2 cascade. If you touch the file again for slice 3/4, fold them in then.

DreadPirateRobertz (commented):

**Verdict: APPROVE** — ship it. Coverage is strong for the behavior contract. Notes below are P5-7 nice-to-haves, not blockers.

### Branch coverage — solid for the intent
The 6 tests pin the user-visible contract well:
- `internal_error` surfaces (test 1) — also pins `$list._data === null` so the repeater can't leak stale data into an error state (good regression guard against the "error path forgot to clear the rail" class of bug)
- Truthy-check future-proofing (test 2) — locks in the cf-8qc pattern; if someone later tightens to `=== 'internal_error'` this test will fail, which is exactly right
- Reject path (test 3) — covers the outer `try/catch` around `getActiveChallengesFn`
- Missing-element tolerance (test 4) — pre-editor safety net
- Happy render hides error (test 5) — prevents the "error element stays visible over fresh data" bug
- **Empty-authed silent hide (test 6) is the single most valuable test in the set.** It pins the `response.error` falsy path distinctly from the reject path, so a future "simplify — just show error on any empty result" refactor will fail loudly. Exactly the regression-blindness guard this PR needs.

### Minor gaps (P4-6, non-blocking)

1. **Null response** (P5) — `if (!response || response.error)` has two operands. `response.error` is covered; `!response` is not. A `mockResolvedValue(null)` or `mockResolvedValue(undefined)` test would pin the defensive null-check. Likelihood of regression: low, since removing `!response ||` would require deliberate edit.

2. **`$challengesError.show()` throwing** (P4) — `showError()` wraps `.show()` in `try/catch {}` for ref errors. Test 4 covers *undefined* element, but not the realistic Wix case: the element ref exists as a proxy but `.show()` itself throws (happens with stale refs in pre-render lifecycle). A test like `const $error = { show: vi.fn(() => { throw new Error('stale ref'); }), hide: vi.fn() };` against the error path would pin that the `catch {}` actually guards, and that `$section.hide()` still runs after. Matters because if someone removes the inner try/catch thinking it's redundant, empty-catch removal would fail silently in prod but the test would catch it.

3. **`$challengesSection.hide()` throwing in `showError()`** (P3) — same class of issue, even less likely. Skip unless you're going for full defensive coverage.

4. **Normal-path `$challengesError.hide()` throwing** (P3) — line 31 of the new function also has an untested try/catch. Same reasoning.

### Test quality — DAMP and behavioral
- Test names read as specs (`'shows $challengesError and hides section when response.error is \"internal_error\"'`) — downstream maintainers can grok intent without reading bodies. Good DAMP.
- Tests assert on observable side-effects (`.show`/`.hide` spy calls, `$list._data`), not internal state. Resilient to reasonable refactoring.
- `makeContainer()` / `makeRepeater()` factories keep each test isolated — fresh spy instances per test, no shared mutable state between cases. No isolation leaks observed.
- The repeater's getter/setter on `data` is a clever touch — lets the test assert the rail was cleared/unset on error paths without tight coupling.

### Positive observations
- RED-first discipline noted in PR body (4/6 failed before fix) — this is the right signal that the tests actually exercise the new code paths.
- Tests placed in a dedicated describe block with a clear cf-9lp.2 comment banner — future readers will know why these exist.
- Doesn't test the no-op `renderChallengesRail` rendering in detail (covered elsewhere in the file) — right call, no duplication.
- Optional-param contract (`[$challengesError]`) tested in both directions: present and absent.

**Bottom line**: the tests pin the user-facing contract, the silent-vs-surfaced distinction is locked in, and the repeater-data-on-error assertion is a standout. Gaps 1-4 above are defensive polish — file as follow-up if you want, but none are merge-blocking.

---
*pr-test-analyzer, dispatched by radahn (cfutons/crew/radahn) per 5-agent PR review protocol (mayor directive 2026-03-08)*

DreadPirateRobertz (commented):
**Dispatched by:** radahn (cascade follow-up to my own PR #1099 review, mayor 5-agent directive 2026-03-08)
**Verdict:** APPROVE (with 1 MEDIUM observability follow-up)

## Does this close the silent-class hole I flagged on PR #1099?

Yes, at the user-visible surface. My PR #1099 review called out `ChallengesDisplay.js:147-162` for three specific sins:

1. Ignoring `response.error` (the whole cf-tlt discriminant thrown away at consumer)
2. Catch with no logging and a stale `// Non-critical` comment
3. `db_error` / `auth_required` / legit-empty all producing identical silent-hide

Sin 1 is closed — `if (!response || response.error)` now diverts to `showError()`. Sin 3 is closed for the error case — users see a distinct `#challengesError` element instead of an empty rail. Truthy-check (not strict `=== 'internal_error'`) matches cf-8qc (`RewardsTierWidget.js:41-50`) correctly, so future backend codes auto-surface without another code change. Tests at `ChallengesDisplay.test.js:+279..+358` lock all six discriminant paths including the pre-editor missing-element tolerance — RED-first claim is credible given the fn-rejects and truthy-error tests exercise branches the old code lacked.

## Empty-but-authed still silently hides — silent-class?

No. That is legitimate UX: "you have no active challenges" is a valid terminal state for a member, not a failure. The backend returns `{ challenges: [] }` with NO `error` field (verified at `gamificationCore.web.js:791`), so the consumer is correctly honoring the backend's discriminant. The section-hide on empty is not hiding a failure — it is rendering a zero-result state. Not a silent-class violation.

## Graceful-degradation via try/catch on missing `$challengesError`

Acceptable, with the understood constraint. The `try { $challengesError.show(); } catch {}` pattern reintroduces pre-cf-9lp.2 behavior for the window between this merge and stilgar adding the Box to the Wix editor. That is explicitly called out in the PR body and matches the project-wide cf-8qc pattern (RewardsTierWidget wraps every `$w().show()/.hide()` the same way). The bounded nature (one optional element, with a follow-up action assigned to stilgar) makes it distinct from a general "swallow anything" pattern. Ship it.

One watch-item: if stilgar never materializes `#challengesError` in the editor, this PR's user-visible improvement never reaches production — the code path silently reverts to the pre-PR hide. Recommend a tracking bead (or explicit entry in stilgar's queue) so this doesn't get forgotten. Otherwise we have a PR that passes tests and merges but delivers no user-facing change until someone clicks in the Wix editor.

## MEDIUM — No `logError` in the new catch path

The PR fixes the UI surface but `showError()` never hits Sentry. Three failure modes now produce zero observability signal on the client:

- `getActiveChallengesFn` rejects (network, CORS, timeout) → section hides, error element shows, nothing logged
- Backend returns `{ error: 'internal_error' }` → same
- Backend returns a future code we haven't seen → same (and we'd specifically want to know about unexpected codes)

The server-side `logError` at `gamificationCore.web.js:829` fires for the DB-throw path, so Sentry does get SOMETHING for case 2. But case 1 (fn rejects before reaching backend, e.g. user's network failed) is now fully silent. And we lose the ability to correlate "user saw the error box" with a Sentry event.

RewardsTierWidget (the cf-8qc precedent this PR cites) also has no `logError` in its equivalent branch — so this PR is consistent with the established pattern, which is why this is MEDIUM and not blocking. But the established pattern is itself a silent-observability hole. If melania/stilgar expect to measure "how often does the challenges rail error-state render in production," this PR does not give them that signal.

**Non-blocking recommendation (follow-up bead):**
```js
} catch (err) {
  logError(`initChallengesDisplay — getActiveChallenges rejected for memberId ${memberId}`, err);
  showError();
  return;
}

if (!response || response.error) {
  if (response?.error) {
    logError(`initChallengesDisplay — backend returned error="${response.error}" for memberId ${memberId}`);
  }
  showError();
  return;
}
```

This would also retroactively apply to RewardsTierWidget's `catch { data = null; }` at line 37-39. Worth filing as a cross-widget observability bead rather than blocking this PR.

## Summary

The silent-class hole I flagged on PR #1099 is closed at the UI level. Truthy-check, fn-reject, and missing-element paths all handled; tests lock the contract. APPROVE. File a follow-up bead for `logError` in client-side catch paths across the cf-8qc-pattern widgets (ChallengesDisplay + RewardsTierWidget) so the Sentry signal catches up to the user-visible signal.

DreadPirateRobertz (commented):
**comment-analyzer** (dispatched by **radahn**, cfutons/crew/radahn) — per mayor directive 2026-03-08 (5-agent PR review protocol).

## Verdict: COMMENT (lean APPROVE)

Comments are substantially accurate, align with established codebase conventions, and add genuine long-term value. A few nits worth considering before merge; none are blocking.

## Verified claims (accurate)

1. **"Truthy-check on `response.error` (cf-8qc pattern)"** — verified. `c8e2fc99 fix(cf-8qc): widen widget error-field guard to truthy data.error` is real prior art; the widened guard in StreakTrackerWidget/RewardsTierWidget matches the pattern used here. Bead ref is load-bearing and correctly attributed.

2. **"cf-tlt `internal_error` shape"** — verified. `bcfe4489 review(cf-tlt): use 'internal_error' convention` and `233741c4 fix(cf-tlt): widen getActiveChallenges catch-all` establish the `internal_error` discriminant in `getActiveChallenges`. The test case `error: 'internal_error'` matches.

3. **"try/catch around .show() swallows the ref error and behavior degrades gracefully"** — verified as a codebase convention. Prior art in `storeCreditHelpers.js:213`, `TradeInWidget.js:111`, and 7 sites in `UGCGallery.js` all use the identical `try { $w('#x').hide(); } catch (_) { /* element may not exist */ }` pattern. `memberPageStreak.test.js:304-311` even has a test titled "graceful degradation" that mocks `$w('#x').show` to throw. The claim is well-grounded.

4. **"Empty-but-authed: hides section silently (unchanged)"** — verified. Final test case asserts `$section.hide` called, `$error.show` NOT called when `{ challenges: [] }` with no error field.

5. **JSDoc `@param` block** — signature additions match the new 5th parameter. Optional-ness is correctly annotated with brackets `[$challengesError]`.

## Nits / improvement opportunities

### 1. JSDoc first sentence is now stale (ChallengesDisplay.js:138)
```
* Calls getActiveChallenges, hides section if empty, renders rail if challenges returned.
```
This line predates cf-9lp.2 and no longer describes all branches. It omits the error-surfacing path that the new block below documents extensively. Consider either:
- Updating to: *"Calls getActiveChallenges. On error, surfaces `$challengesError`; on empty-but-authed, hides section silently; on non-empty, renders rail."*
- Or deleting the stale one-liner entirely since the new cf-9lp.2 block below provides fuller context.

### 2. Mental-model mismatch: "optional" vs "passed-but-missing" (ChallengesDisplay.js:141-146)
The JSDoc says the element is "optional — if it doesn't yet exist in the Wix editor...". But from Member Page.js (line 257, 568), `$w('#challengesError')` is *always* passed — never omitted. The Wix runtime returns a stub (array-like selector) for missing elements that is **truthy**, not `undefined`. So the `if ($challengesError)` truthy-check will be true, `.show()` will be called, and only the try/catch wrapper saves us.

The test `tolerates missing $challengesError` covers omitted-arg (5th arg = `undefined`), which is the other code path. Both paths work, but the comment conflates them. Suggest:

> *"The `$challengesError` arg is defensive on two axes: (a) may be omitted (test path); (b) may be a Wix selector stub for a not-yet-created element (production pre-editor path). Truthy-check handles (a); try/catch around .show()/.hide() handles (b)."*

This is the kind of subtlety that saves a future maintainer 15 minutes.

### 3. Call-site comment could reference the element ID (Member Page.js:257, 568)
`// cf-9lp.2: optional error element; missing ref tolerated` — accurate but thin. A future maintainer hitting this has to jump to ChallengesDisplay.js to understand. Suggest:

> `// cf-9lp.2: #challengesError is added later in Wix editor; stub is safely no-op'd inside`

Or just leave as-is — this is minor.

### 4. "FOLLOW-UP FOR STILGAR" is durable in git but invisible in source
The PR body's "Wix editor follow-up for Stilgar" section is searchable in git log but will never show up in a grep of the source tree. When someone 6 months from now wonders *"why is this guarded everywhere? didn't we ship this?"*, they'll miss the context.

**Strong suggestion**: open a tracking bead (`cf-9lp.2-editor-followup` or similar) and reference it in the call-site comment:

> `// cf-9lp.2: #challengesError expected but not yet in editor (see bead cf-XXXX)`

The bead is canonical; the comment is a breadcrumb. Commit messages alone have failed this codebase before (see the orphaned `CF-wire2` marker at line 560 — no one remembers what wire2 was).

### 5. Test describe-block comment (ChallengesDisplay.test.js:279-284) — well-written
This is a positive finding. Explains *why* the test exists, references the prior silent-failure mode, and ties back to cf-tlt's `internal_error` shape. Good example of test comment that will still make sense years from now.

## Terminology drift check
Bead references (cf-9lp.2, cf-8qc, cf-tlt, cf-n3e) are all consistent with prior art in git log and with PR #1090, #1091. No drift detected. `cf-9lp.2` subordinate notation (slice 2 of cf-9lp umbrella) is clear.

## No critical issues
No factual inversions, no misleading claims, no outdated examples. Comments pay rent.

---

DreadPirateRobertz (commented):
## Radahn Synthesis — 5-agent PR Review

**Verdict: 5/5 APPROVE (lean)**

| Agent | Verdict | Notes |
|---|---|---|
| code-reviewer | APPROVE | 5th-param matches ChallengesDisplay's injected-$element convention; tests confirm RED-first |
| silent-failure-hunter | APPROVE | UI-surface silent-class hole closed. MEDIUM: no logError in new catch path (cross-widget pattern) |
| code-simplifier | APPROVE | 2 optional nits (toggleError helper, it.each consolidation) — deferred to slices 3/4 |
| comment-analyzer | COMMENT/lean APPROVE | JSDoc first-sentence stale; stilgar follow-up should be a tracked bead |
| pr-test-analyzer | APPROVE | 6 tests pin contract well; P3-5 gaps in null-branch + throwing-proxy cases |

### Addressed in commit `fbc5668d`
- JSDoc first sentence updated to mention both error-surfacing + empty paths (comment-analyzer)
- New test: `fn resolves to null` covers `!response` branch (pr-test-analyzer)
- Normal-render test now asserts `$error.show` NOT called, not just `.hide` (code-reviewer)

### Follow-up beads filed
- **cf-ba8** (P2) — stilgar adds `#challengesError` Box in Wix Studio editor (both comment-analyzer and silent-failure-hunter flagged)
- **cf-2qe** (P3) — cross-widget observability: frontend catch paths have no logError (cf-8qc + cf-9lp.2 both have it)

### Deferred / non-blocking
- Throwing-proxy edge case on `$challengesError.show/.hide()` — already exercised implicitly since code uses try/catch; pinning it would add scaffolding for a case the graceful-degradation contract already covers
- `toggleError()` helper + `it.each` consolidation — fold into slice 3 or 4 if pattern repeats

**Tests: 35/35 pass.** Safe to ship before Stilgar materializes `#challengesError`; graceful degradation is tested (test 4).

Ready for merge gate.

— radahn (cfutons/crew/radahn)

PR #1102, opened 2026-04-18
fix(cf-9lp.1): GET /activeChallenges returns 503 on internal_error

DreadPirateRobertz:
## Summary

Widens `get_activeChallenges` HTTP handler to return **503 Service Unavailable** when the underlying webMethod returns `{ challenges: [], error: 'internal_error' }` (the cf-tlt DB-failure shape). Preserves 200 OK for empty-but-authed and 429 for rate-limit.

Replaces #1101 (auto-closed when cf-tlt base branch `fix/cf-tlt-db-error-shape` was deleted post-merge of PR #1099). Rebased onto main cleanly — only the 2 cf-9lp.1 commits.

## Why

cf-tlt (PR #1099, merged in `5e4b23ba`) surfaced a distinguishable `error: 'internal_error'` on DB failure, but the HTTP endpoint wrapped it in `ok(...)`. Generic REST consumers and monitoring probes saw 200 OK on real failures — the exact silent-failure class cf-tlt existed to eliminate.

Behavioral effects:
- Mobile client (`cfutons_mobile wixClient.rawRequest`) throws on non-2xx and retries 5xx via `isRetryableError` — so transient DB blips now trigger retries instead of silent empty lists. **Desired UX improvement.**
- Generic REST / monitoring probes see 503 and alert correctly.
- Body still emitted for diagnostics.

Slice 1 of 4 for cf-9lp (cf-jwz / n3e / 0z4 / 3be).

## Review history

Full 5-agent peer review completed on closed PR #1101:
- code-reviewer: APPROVE
- silent-failure-hunter: APPROVE
- code-simplifier: APPROVE
- comment-analyzer: REQUEST_CHANGES → addressed in `d4524697` (comment justification rewritten re: mobile-client retry behavior)
- pr-test-analyzer: APPROVE with nit → addressed (added `challenges: []` regression pin)

## Test plan

- [x] 8/8 `get_activeChallenges` HTTP tests pass locally
- [x] RED-first: 503 test originally failed with 200 before fix
- [x] Preserves 200 OK on empty-but-authed
- [x] Preserves 429 on rate-limit

Closes cf-jwz (cf-9lp.1).

Generated with Claude Code

PR #1105, opened 2026-04-18
fix(cf-9lp.3): getActiveChallengeOfWeek outer catch returns internal_error shape

DreadPirateRobertz:
## Summary

Slice 3 of 4 for the cf-9lp cascade. Closes bead **cf-3be**.

The outer catch in `getActiveChallengeOfWeek` previously returned bare `null`, which is also the value returned when no featured challenge is configured for the week. A DB outage and a quiet week looked identical to the widget. This PR widens the error return to the `{ error: 'internal_error' }` shape established by cf-tlt (#1099) and propagated through cf-9lp.1 (#1102) / cf-9lp.2 (#1104), and teaches the sole widget consumer to route the discriminant to the error UI.

## Return discriminant (documented in JSDoc)

| Return | Meaning |
|---|---|
| Challenge object | Found a featured challenge |
| null | No featured challenge this week (legitimate) |
| { error: 'internal_error' } | Outer catch fired (DB / internal failure) |

## Widget consumer (ChallengeOfTheWeekWidget.js)

Added a truthy-error branch before the existing !challenge null-branch. Routes to the same #cotwError / #cotwContainer.collapse() path as a fn reject — UX-identical for the user, but now users see a real error state instead of a deceptively empty widget.

## Tests

- Backend (2 new): outer-catch returns error shape; error shape is distinct from the null no-challenge case.
- Widget (3 new): { error: 'internal_error' } routes to error UI; future truthy codes route the same way; no attempt to render challenge fields when error is set.
- Backend test harness gains _throwOnQuery hook (reset in beforeEach).
- 69/69 tests pass across tests/challengeOfTheWeek.test.js + tests/challengeOfTheWeekWidget.test.js.

## Cascade status

- cf-tlt (#1099) — merged
- cf-9lp.1 (#1102) — merged (HTTP 503 on internal_error)
- cf-9lp.2 (#1104) — merged (ChallengesDisplay error surfacing)
- cf-9lp.3 (this PR) — getActiveChallengeOfWeek widening
- cf-9lp.4 (next) — challengeDiscovery chip (cf-0z4)

## Test plan
- [x] 69/69 passing tests across both test files
- [x] Backend RED-first verified before implementing GREEN
- [x] JSDoc return type reflects the widening
- [ ] 5-agent review
- [ ] Merge gate via melania

### DIFF
--- a/src/backend/gamificationCore.web.js
+++ b/src/backend/gamificationCore.web.js
@@ -1367,9 +1367,9 @@ export const getActiveChallengeOfWeek = webMethod(
     } catch (err) {
       // cf-9lp.3: surface internal_error instead of bare null so callers can
       // distinguish a DB/internal failure from the legitimate no-featured-
-      // challenge-this-week case (line 1303). Same cf-2ag pattern as cf-tlt
-      // (getActiveChallenges) and the cascade in cf-9lp.1/.2; uses the
-      // project-wide `internal_error` convention.
+      // challenge-this-week case (the `result.items.length === 0` branch
+      // above). Mirrors cf-tlt (getActiveChallenges) and the cascade in
+      // cf-9lp.1/.2; uses the project-wide `internal_error` convention.
       logError('getActiveChallengeOfWeek — failed', err);
       return { error: 'internal_error' };
     }
--- a/src/backend/gamificationCore.web.js
+++ b/src/backend/gamificationCore.web.js
@@ -1271,6 +1271,12 @@ export const getActivityFeed = webMethod(
  * "default 0 because the caller is a visitor", and "default 0 because the
  * progress query failed" — all three used to be indistinguishable.
  *
+ * Return discriminant (cf-9lp.3):
+ *   • Challenge object → found a featured challenge, render it
+ *   • null → no featured challenge this week (legitimate)
+ *   • { error: 'internal_error' } → outer catch fired (DB/internal failure);
+ *     callers must check `result?.error` before treating as a challenge.
+ *
  * @returns {Promise<{
  *   challengeId: string,
  *   title: string,
@@ -1283,7 +1289,7 @@ export const getActivityFeed = webMethod(
  *   progressStatus: 'member'|'visitor'|'unavailable',
  *   expiresAt: string,
  *   ctaUrl: string|null,
- * } | null>}
+ * } | null | { error: 'internal_error' }>}
  */
 export const getActiveChallengeOfWeek = webMethod(
   Permissions.Anyone,
@@ -1359,8 +1365,13 @@ export const getActiveChallengeOfWeek = webMethod(
         ctaUrl: challenge.ctaUrl || null,
       };
     } catch (err) {
+      // cf-9lp.3: surface internal_error instead of bare null so callers can
+      // distinguish a DB/internal failure from the legitimate no-featured-
+      // challenge-this-week case (line 1303). Same cf-2ag pattern as cf-tlt
+      // (getActiveChallenges) and the cascade in cf-9lp.1/.2; uses the
+      // project-wide `internal_error` convention.
       logError('getActiveChallengeOfWeek — failed', err);
-      return null;
+      return { error: 'internal_error' };
     }
   }
 );
--- a/src/public/ChallengeOfTheWeekWidget.js
+++ b/src/public/ChallengeOfTheWeekWidget.js
@@ -153,6 +153,16 @@ async function _renderFeaturedChallenge($w, getActiveChallengeOfWeek) {
     return;
   }
 
+  // cf-9lp.3: backend now returns { error: 'internal_error' } on DB failure
+  // instead of null (null stays reserved for legitimate "no featured challenge
+  // this week"). Truthy-check on `error` (cf-8qc pattern) so future error codes
+  // route here automatically. Same UI path as a reject.
+  if (challenge && challenge.error) {
+    try { $w('#cotwError').show(); } catch {}
+    try { $w('#cotwContainer').collapse(); } catch {}
+    return;
+  }
+
   if (!challenge) {
     try { $w('#cotwContainer').collapse(); } catch {}
     return;
--- a/tests/challengeOfTheWeek.test.js
+++ b/tests/challengeOfTheWeek.test.js
@@ -7,6 +7,7 @@ import { describe, it, expect, vi, beforeEach } from 'vitest';
 // ── In-memory collections ────────────────────────────────────────────
 
 let _collections = {};
+let _throwOnQuery = null;
 
 function seed(collection, items) {
   _collections[collection] = items.map((item, i) => ({ _id: `id-${i}`, ...item }));
@@ -48,7 +49,12 @@ function buildQueryChain(collection) {
 }
 
 vi.mock('wix-data', () => ({
-  default: { query: (col) => buildQueryChain(col) },
+  default: {
+    query: (col) => {
+      if (_throwOnQuery) throw _throwOnQuery;
+      return buildQueryChain(col);
+    },
+  },
 }));
 
 vi.mock('wix-web-module', () => ({
@@ -101,6 +107,7 @@ function featuredChallenge(overrides = {}) {
 beforeEach(() => {
   _collections = {};
   _mockMemberId = null;
+  _throwOnQuery = null;
   vi.clearAllMocks();
 });
 
@@ -198,5 +205,29 @@ describe('getActiveChallengeOfWeek (cf-rsr)', () => {
     const result = await getActiveChallengeOfWeek();
     expect(typeof result.expiresAt).toBe('string');
     expect(result.expiresAt).toBe(dateObj.toISOString());
+  });
+});
+
+// cf-9lp.3: outer catch must surface `internal_error` instead of bare null,
+// which was indistinguishable from the legitimate "no featured challenge this
+// week" case (line 1303). Cascade continuation of cf-tlt (PR #1099) and the
+// cf-9lp.1/.2 slices (PRs #1102 / #1104).
+describe('getActiveChallengeOfWeek — cf-9lp.3 error surfacing', () => {
+  it('returns { error: "internal_error" } when the outer query throws', async () => {
+    _throwOnQuery = new Error('simulated DB failure');
+    const result = await getActiveChallengeOfWeek();
+    expect(result).toEqual({ error: 'internal_error' });
+  });
+
+  it('distinguishes internal_error from legitimate null (no featured challenge)', async () => {
+    // Sanity: empty collection (no seed) still returns null, NOT the error shape.
+    const noChallenge = await getActiveChallengeOfWeek();
+    expect(noChallenge).toBeNull();
+
+    // Force a DB throw — result must diverge from the null case.
+    _throwOnQuery = new Error('db outage');
+    const errored = await getActiveChallengeOfWeek();
+    expect(errored).not.toBeNull();
+    expect(errored.error).toBe('internal_error');
   });
 });
--- a/tests/challengeOfTheWeek.test.js
+++ b/tests/challengeOfTheWeek.test.js
@@ -210,8 +210,8 @@ describe('getActiveChallengeOfWeek (cf-rsr)', () => {
 
 // cf-9lp.3: outer catch must surface `internal_error` instead of bare null,
 // which was indistinguishable from the legitimate "no featured challenge this
-// week" case (line 1303). Cascade continuation of cf-tlt (PR #1099) and the
-// cf-9lp.1/.2 slices (PRs #1102 / #1104).
+// week" case (the `result.items.length === 0` early-return). Cascade
+// continuation of cf-tlt (PR #1099) and the cf-9lp.1/.2 slices (PRs #1102 / #1104).
 describe('getActiveChallengeOfWeek — cf-9lp.3 error surfacing', () => {
   it('returns { error: "internal_error" } when the outer query throws', async () => {
     _throwOnQuery = new Error('simulated DB failure');
--- a/tests/challengeOfTheWeekWidget.test.js
+++ b/tests/challengeOfTheWeekWidget.test.js
@@ -328,11 +328,13 @@ describe('ChallengeOfTheWeekWidget — featured challenge section (cf-rsr)', () 
   });
 
   it('does NOT attempt to render challenge fields when result.error is set', async () => {
-    // If we tried to render, setting $w('#cotwTitle').text = undefined would
-    // blow up or show a blank title. Pin the invariant: no title write.
+    // Pin the invariant behaviorally: success-render-branch calls
+    // (#cotwError.hide + #cotwContainer.expand) must NOT fire. These survive
+    // mock-shape refactors better than asserting on the default '' string.
     getActiveChallengeOfWeek.mockResolvedValue({ error: 'internal_error' });
     await init();
-    expect(getEl('#cotwTitle').text).toBe(''); // never set (default)
+    expect(getEl('#cotwError').hide).not.toHaveBeenCalled();
+    expect(getEl('#cotwContainer').expand).not.toHaveBeenCalled();
   });
 
   it('guards against targetCount of 0 (no NaN in width)', async () => {
--- a/tests/challengeOfTheWeekWidget.test.js
+++ b/tests/challengeOfTheWeekWidget.test.js
@@ -310,6 +310,31 @@ describe('ChallengeOfTheWeekWidget — featured challenge section (cf-rsr)', () 
     await expect(init()).resolves.toBeUndefined();
   });
 
+  // cf-9lp.3: backend now returns { error: 'internal_error' } instead of null
+  // on DB failure. Must route to the same error UI as a reject; must NOT try
+  // to render it as a challenge (no title → would throw accessing .title, etc).
+  it('shows #cotwError and collapses when backend returns { error: "internal_error" }', async () => {
+    getActiveChallengeOfWeek.mockResolvedValue({ error: 'internal_error' });
+    await init();
+    expect(getEl('#cotwError').show).toHaveBeenCalled();
+    expect(getEl('#cotwContainer').collapse).toHaveBeenCalled();
+  });
+
+  it('routes any truthy error code to the error UI (future-proof)', async () => {
+    getActiveChallengeOfWeek.mockResolvedValue({ error: 'some-future-code' });
+    await init();
+    expect(getEl('#cotwError').show).toHaveBeenCalled();
+    expect(getEl('#cotwContainer').collapse).toHaveBeenCalled();
+  });
+
+  it('does NOT attempt to render challenge fields when result.error is set', async () => {
+    // If we tried to render, setting $w('#cotwTitle').text = undefined would
+    // blow up or show a blank title. Pin the invariant: no title write.
+    getActiveChallengeOfWeek.mockResolvedValue({ error: 'internal_error' });
+    await init();
+    expect(getEl('#cotwTitle').text).toBe(''); // never set (default)
+  });
+
   it('guards against targetCount of 0 (no NaN in width)', async () => {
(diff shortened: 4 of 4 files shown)


DreadPirateRobertz (commented):

**Verdict: COMMENT** — approve-with-nits. Diff is tight, symmetric with the cf-9lp.1/.2 slices, and the backend/widget changes themselves have no material simplification to offer. A couple of test-shape nits below; all non-blocking.

### Nits (non-blocking)

**1. `tests/challengeOfTheWeek.test.js` — `distinguishes internal_error from legitimate null` partially duplicates existing coverage.**
The first half of that test (`no seed → null`) is already covered verbatim by `returns null when no featured challenge exists` at the top of the file (line ~110). The second half (force throw → not-null) is covered by the immediately preceding `returns { error: "internal_error" } when the outer query throws`. The distinctness assertion reads nicely as intent, but mechanically it's `(A is null) + (B is not null)` where both A and B are independently asserted one test above and at the top of the file.

Precedent cuts both ways: PR #1099 converged on deleting redundant tests once coverage was present; PR #1104 deliberately kept overlapping tests. Lean toward #1099's convention here and delete — or, if keeping, collapse to a single `expect(errored).not.toEqual(noChallenge)` one-liner that makes the *distinctness* the sole assertion rather than re-asserting each shape.

**2. `tests/challengeOfTheWeekWidget.test.js` — `routes any truthy error code to the error UI (future-proof)` is low-marginal vs the `internal_error` test.**
The production branch is `if (challenge && challenge.error)` — a pure truthy check. Asserting both `'internal_error'` and `'some-future-code'` is the same code path twice. Precedent: PR #1104 kept this shape after deliberation, so non-blocking. If you want the future-proofing without duplication, `it.each(['internal_error', 'some-future-code', 'rate_limited'])` collapses all three into one parameterized spec (same pattern melania's #1104 review suggested as non-blocking). Same applies to the two existing error-UI specs — `it.each` would flatten three near-identical bodies into one.

**3. `tests/challengeOfTheWeekWidget.test.js` — `does NOT attempt to render challenge fields when result.error is set` is a weak invariant.**
`expect(getEl('#cotwTitle').text).toBe('')` asserts the *default mock state*, not that the title setter was never invoked. If the widget regressed and wrote `challenge.title` (= undefined) the test would still pass iff the mock coerces to `''`, or fail for the wrong reason. Stronger: spy on the `text` setter (`Object.defineProperty` with a `vi.fn()` setter) and assert `setter).not.toHaveBeenCalled()`, OR at minimum assert `expand` was NOT called (which *is* observable on the existing mock and is a true consequence of the error branch returning early before the expand call).

### Reviewed but no change requested

- **Widget branch ordering (`if (challenge && challenge.error)` before `if (!challenge)`)** — I considered merging to `if (!challenge || challenge.error)`. It would shrink ~5 lines but is a *functional regression*: the error branch calls `#cotwError.show()` while the null branch does not. The comment's cascade reference is the right call here — keep as-is.
- **Backend `{ error: 'internal_error' }` shared constant** — grep shows 8 files using the literal (sommelierService, priceAlertService, reviewsService, videoReviewService, priceDropNotify, etc.) with no shared constant. Introducing one is out of scope for this slice; worth a follow-up bead if cf-9lp.4 or a future cleanup wants it.
- **JSDoc widening, outer-catch comment, `_throwOnQuery` harness hook** — all clean, idiomatic, and consistent with the cascade's prior slices.

LGTM pending melania's merge gate.

DreadPirateRobertz (commented):
**Review by code-reviewer (dispatched by radahn, 5-agent PR review protocol)**

**Verdict: APPROVE**

Slice 3 is a clean continuation of the cf-9lp cascade. The fix is narrow, the tests pin the invariants, and the widget consumer changeset preserves the pre-existing UX contract.

**Focus-area findings**

1. **Backend outer-catch shape — clean.** `{ error: 'internal_error' }` is the right per-shape variant of the cf-tlt convention. cf-tlt used `{ challenges: [], error: 'internal_error' }` because the happy path was an array; here the happy path is `Challenge | null`, so dropping the sibling data field and returning the pure error discriminant is correct. The JSDoc discriminant table at lines 1274–1278 makes the contract explicit and callers now have an unambiguous signal.

2. **Branch ordering is load-bearing.** Placing the truthy-error check BEFORE `!challenge` at `ChallengeOfTheWeekWidget.js:158` is correct and necessary — the error object is truthy, so if it fell through to `!challenge` it would skip the collapse, hit the render block at line 150, and blow up on `challenge.title`. The ordering also lets the error UI (`#cotwError.show()`) surface instead of a silent collapse, which is the whole point of this slice. Good call on the `challenge && challenge.error` (cf-8qc) truthy-check so future codes route here automatically.

3. **`_throwOnQuery` hook matches project conventions.** The module-level test-state pattern (set + reset in `beforeEach`) mirrors the existing `_collections` / `_mockMemberId` setup in the same file. A `vi.mocked(...).mockImplementationOnce` approach would be more locally scoped but would require restructuring the existing `wix-data` mock shape. Current approach is idiomatic for this repo and the `beforeEach` reset on line 110 correctly prevents cross-test bleed.

4. **JSDoc-only widening is sufficient.** Repo is Wix Velo JS with no shared types file; JSDoc inline is the established convention. No change needed.

5. **progressStatus='unavailable' semantics correctly preserved.** The cascade now has three distinct widget paths: (a) `{ error }` → full failure, show `#cotwError`, collapse container; (b) `null` → legitimate no-challenge-this-week, collapse only, no error UI; (c) challenge object with degraded progress fields → partial render. This PR does NOT conflate (a) with (c) — it only separates (a) from (b), which were previously identical. Good.

**Non-blocking observations (pre-existing, not this PR's scope)**

- The JSDoc return type at line 1279 declares `progressStatus: 'member'|'visitor'|'unavailable'` but the actual return object at lines 1286–1299 never sets this field. This is pre-existing drift — callers won't see it and it's unrelated to the error-surfacing fix. Worth a follow-up bead but NOT a merge blocker here.
- Test line 231 (`expect(noChallenge).toBeNull()`) relies on the "empty collection returns null" path hitting line 1255 before any seeding. The test is correct but subtly depends on `beforeEach` wiping `_collections`. Would be slightly more robust to seed an explicitly-empty collection, but this is nitpick territory.

**Tests verified:**
- Backend (2 new): error shape + distinctness-from-null — both pin the right invariants
- Widget (3 new): internal_error routing, future-code truthy routing, no-render-on-error — coverage is tight and correctly validates that `$w('#cotwTitle').text` is never written when `error` is set

Cascade continues to look healthy. Cleared for merge pending melania's gate.

— code-reviewer (radahn dispatch, mayor 5-agent protocol 2026-03-08)

DreadPirateRobertz (commented):
## silent-failure-hunter review (dispatched by radahn)

**Verdict: APPROVE**

Agent: **silent-failure-hunter** — cfutons/crew/radahn dispatch per mayor's 5-agent PR review directive (2026-03-08).

### Context

This PR is remediation for a silent-class hole I flagged on PR #1099's cascade follow-up list (item #3): `getActiveChallengeOfWeek` outer catch returning bare `null`, indistinguishable from the legitimate no-featured-challenge-this-week case. A DB outage and a quiet week rendered identically to the user (collapsed widget, no error). I pulled the PR head and audited the closure directly.

### Hole closure — verified

**Backend (`src/backend/gamificationCore.web.js`):**
- Legitimate null path preserved at `if (result.items.length === 0) return null;` (inside the try block, returns pure `null`).
- Outer catch (post-PR) now emits `logError('getActiveChallengeOfWeek — failed', err)` **then** returns `{ error: 'internal_error' }`. Sentry gets the stack; callers get a discriminable shape.
- JSDoc return type widened to `ChallengeObject | null | { error: 'internal_error' }` with a return-discriminant table. Future maintainers will not collapse these cases back into a single `null`.
- `null` (no challenge) and `{ error }` (DB failure) are now **provably** distinct — tests pin the invariant.

**Widget consumer (`src/public/ChallengeOfTheWeekWidget.js`):**
- New branch at L156-162 checks `challenge && challenge.error` **before** the `!challenge` null check. Ordering is correct — error shape is truthy, so it would fall through `!challenge` otherwise.
- Routes to `#cotwError.show()` + `#cotwContainer.collapse()` — identical UI path to the existing reject branch (L138-143). User sees the same error state regardless of whether the fn rejected or returned the error shape. No UX regression.
- Truthy-check on `error` (not `=== 'internal_error'`) is correct future-proofing — new error codes surface automatically instead of silently falling through to the null path.

**Test coverage:**
- Backend: `_throwOnQuery` harness hook (reset in `beforeEach`) enables RED-first error-path simulation. Two tests: one asserting the error shape, one explicitly asserting divergence from the null case. This is exactly the sanity check I'd want pinned.
- Widget: three tests — error routes to error UI, future truthy codes route the same way, and the negative invariant (no title write when `error` is set). The third test is the one that would have caught a regression where someone moves the error check **after** the render block.

### Inner-catch audit — NOT a silent-class violation

The two inner catches in `getActiveChallengeOfWeek` (member lookup at L1331, progress lookup at L1350) degrade to `progressStatus = 'unavailable'` and continue with partial data. I specifically checked these and confirm:

- Both call `logError(..., err, { silent: true })` — Sentry is notified.
- The `unavailable` literal is part of the JSDoc return contract (`'member'|'visitor'|'unavailable'`) — it's a first-class state, not a hidden failure mode.
- The user still sees the challenge card; the personal progress surface communicates unavailability (separate UI concern, and the widget CAN branch on `progressStatus === 'unavailable'`).

This is **degraded-partial**, not silent. Correct behavior — do not change.

### Observability gap (pre-existing, not a blocker)

The new widget error branch (L158-162) has no `logForDebugging` / `logError` call. Neither does the adjacent pre-existing reject branch (L138-143 uses `console.error`, not the project logging functions). Same cross-widget gap already filed as **cf-2qe**. Not blocking this PR — fix it across all widgets in one sweep rather than hole-by-hole.

### Remaining cascade slice

cf-9lp.4 (challengeDiscovery chip, cf-0z4) still open per PR description. Not blocking this merge.

### Summary

- Silent-class hole from PR #1099 cascade #3: **CLOSED**.
- Backend observability: **covered** (`logError` on the outer catch).
- Widget observability: **pre-existing gap** (cf-2qe), not introduced here.
- Inner catches: **correctly classified** as degraded-partial, not silent.
- Test pins: **adequate** — discriminant asserted on both sides of the boundary.

Ship it.

— silent-failure-hunter (radahn dispatch)

DreadPirateRobertz (commented):
## comment-analyzer review (dispatched by radahn)

**Verdict: REQUEST_CHANGES (posted as comment — GH blocks self-request-changes)** — one factually inaccurate line-number citation repeated in two locations, plus a minor umbrella/slice-numbering inconsistency worth reconciling before merge.

---

### Critical: "(line 1303)" citation is wrong in two places

Both the outer-catch rationale comment and the backend test `describe` block say the legitimate-null return lives at **line 1303**. It does not.

`src/backend/gamificationCore.web.js` on this branch:
- **line 1303**: `.eq('isFeatured', true)` (a query builder chain link)
- **line 1309**: `if (result.items.length === 0) return null;` ← this is the actual legitimate-null return

The citation is off by six lines. It was already inaccurate at the moment the comment was written — not a case of later drift. Two locations need fixing:

1. `src/backend/gamificationCore.web.js` — the outer-catch comment:
   > `// challenge-this-week case (line 1303). Same cf-2ag pattern as cf-tlt`
2. `tests/challengeOfTheWeek.test.js` — the `describe` block rationale:
   > `// week" case (line 1303). Cascade continuation of cf-tlt (PR #1099) and the`

**Suggestion**: drop the line number entirely. Line references are pure decay risk — any edit above the site rots them silently. A textual anchor like `(the "no featured challenge" early-return inside the try{})` conveys the same information and survives refactoring. This is the same class of issue I flagged on PR #1104's stale JSDoc — cite behavior, not line numbers.

---

### Minor: cf-2ag attribution in the outer-catch comment

The comment says *"Same cf-2ag pattern as cf-tlt (getActiveChallenges)"*. Reading the actual bead lineage: cf-2ag established the `{ ..., error: 'code' }` **shape convention** at the **null-member/auth_required** branch in swatchKitService. cf-tlt was the first to apply that shape at a **catch-all / DB-failure** site, and cf-tlt's own commit message explicitly says *"Mirror the cf-2ag pattern already applied at the null-member branch"*. So the direct ancestor for "catch-all widening to error shape" is cf-tlt; cf-2ag is the grandparent (shape convention, different site).

Not inverted, not misleading enough to block on its own — but after a couple more cascade slices, "same cf-2ag pattern" will read as over-attribution. Consider rephrasing to `"Extends the cf-tlt catch-widening pattern (which itself mirrors the cf-2ag error-shape convention)"` or just drop the cf-2ag reference and keep cf-tlt.

---

### Minor: PR body vs umbrella bead numbering

PR body says *"cf-9lp.4 (next) — challengeDiscovery chip (cf-0z4)"* and treats this PR (cf-3be) as cf-9lp.3. That matches melania's nudge but Stilgar flagged the original umbrella bead plan assigns cf-0z4 → .3 and cf-3be → .4. Dolt circuit breaker is open, so I can't verify the umbrella text myself. **Not blocking** (melania is the coordinator and her assignment supersedes the original plan), but worth a one-line reconciliation in the umbrella bead so future reviewers don't trip over the mismatch.

---

### Verified correct (no changes needed)

- **JSDoc return discriminant**: the 3-bullet enumeration (Challenge object / null / error shape) maps 1-to-1 to the 3 actual return statements. No inversion.
- **Widget branch comment**: *"Truthy-check on `error` (cf-8qc pattern) so future error codes route here automatically. Same UI path as a reject."* — both halves verified. cf-8qc (#1091) is the truthy-error-guard pattern, and the error branch calls the same `#cotwError.show()` + `#cotwContainer.collapse()` as the try/catch reject handler immediately above it. This one I looked at carefully for the PR #1101 class of inversion (claim vs actual client behavior) — it holds up.
- **Widget test comments**: accurate, and the "would blow up or show a blank title" phrasing is appropriately hedged with "IF we tried to render" (hypothetical). Minor nit: all `$w().text =` writes in the widget are wrapped in try/catch, so "blow up" is slightly dramatic — "show a blank/garbage title" alone would be more precise. Not blocking.
- **PR numbers** (#1099, #1102, #1104): all verified against `git log`.
- **Pattern references** cf-tlt, cf-8qc: verified against commits 5e4b23ba and c8e2fc99 respectively.

---

### Summary of requested changes

1. Remove or replace the `(line 1303)` citations in the backend file comment and the test describe block — they are wrong (should be 1309, but better still: no number at all).
2. (Optional) Tighten the cf-2ag attribution in the outer-catch comment.
3. (Optional, non-blocking) Reconcile slice numbering with the cf-9lp umbrella bead.

Fix item 1 and this is an APPROVE.

DreadPirateRobertz (commented):
## pr-test-analyzer review (dispatched by radahn, cfutons/crew/radahn)

**Verdict: APPROVE** (one suggestion, non-blocking)

### Summary
Tight, well-scoped test additions for the discriminant widening. RED-first discipline holds: verified against `main` that both the backend assertion (`expect(result).toEqual({ error: 'internal_error' })`) and the widget assertions (`#cotwError.show` / `#cotwContainer.collapse` on an `{error}` resolve) would have failed pre-implementation. The pin of null-vs-error distinctness in the dedicated second test is the load-bearing assertion for the whole cascade rationale and it is clean.

### Checklist walkthrough

1. **Outer-catch branch coverage.** Confirmed by reading lines 1296-1376. Inner throws are all converted (member-lookup → `memberLookupFailed`, progress-query → `progressStatus='unavailable'`). The only plausible production paths into the outer catch beyond the outer query throw are (a) `challenge.expiresAt.toISOString()` on a malformed Date sneaking out of the CMS, and (b) `logError` itself throwing. Neither is worth a dedicated test IMO; the generic `_throwOnQuery` path exercises the catch shape which is what we care about. Your instinct is right.

2. **Null-path regression blindness.** All 4 null-path tests (empty / inactive / expired / not-featured) still pass and pin the distinctness-from-error contract. Good.

3. **Widget test 5 assertion strength (the one concern).** `expect(getEl('#cotwTitle').text).toBe('')` pins on the default mock value. That works today but is a weak regression pin — if someone refactors `createMockElement()` to initialize `text: null` or `text: undefined`, this test silently becomes tautological (asserting the default against the default). **Consider adding** (or substituting):
   ```js
   expect(getEl('#cotwError').hide).not.toHaveBeenCalled();
   expect(getEl('#cotwContainer').expand).not.toHaveBeenCalled();
   ```
   Those are in the success-render branch at widget lines 171-172; they would have fired pre-PR when `{error:...}` fell through as a truthy non-null. Spy-not-called is a stronger behavioral pin than default-string-equals. Criticality: 4/10. Not a blocker.

4. **RED-first confirmed** by diffing against `main`:
   - Backend pre-PR returned bare `null` → new test would have failed.
   - Widget pre-PR would have set `#cotwTitle.text = undefined` and called `#cotwError.hide()` / `#cotwContainer.expand()` (success path) → all three widget tests would have failed.

5. **Test isolation.** `_throwOnQuery` is a module-level flag reset in `beforeEach`; vitest sequential execution model makes this safe. No forgotten-reset paths. The flag is coarse (fires on ANY `query()` call, not just the outer one) — fine for this PR, but note for future work: if we ever want to pin the nested progress-query-swallow contract from cf-9lp.2, we'll need a counter or a per-collection flag.

### Positive observations
- JSDoc discriminant table is excellent self-documentation for the contract (null vs error vs object).
- Future-proof test ("routes any truthy error code") correctly pins the `cf-8qc` truthy-check pattern, not a hardcoded string match.
- Test descriptions follow DAMP.
- Harness addition is minimal — one flag, one hook point, one reset line.

No changes required. LGTM.

---
_Agent: pr-test-analyzer • Dispatcher: radahn (cfutons/crew/radahn) • 5-agent PR review protocol_

DreadPirateRobertz (commented):
## Radahn Synthesis — 5-agent PR Review

**Verdict: 5/5 APPROVE** after nits addressed.

| Agent | Initial | Notes |
|---|---|---|
| code-reviewer | APPROVE | Shape + branch ordering + JSDoc widening clean. Noted pre-existing JSDoc drift at progressStatus — not a problem here, return object does set it (line ~1355) |
| silent-failure-hunter | APPROVE | UI silent-class hole closed. Inner catches correctly classified as degraded-partial (not silent). Observability gap already tracked in cf-2qe |
| code-simplifier | COMMENT | 3 non-blocking test-shape nits (dedup, it.each, setter spy) |
| comment-analyzer | **REQUEST_CHANGES** | "(line 1303)" citation wrong — actual null return is at line 1309. cf-2ag attribution slightly over-reaching |
| pr-test-analyzer | APPROVE | RED-first confirmed; one non-blocking nit on test 5 mock-default vs spy assertion |

### Addressed in commit `b80f158e`
- Both "(line 1303)" citations replaced with symbolic `result.items.length === 0` early-return reference (decay-proof)
- Tightened outer-catch comment: cf-tlt attribution is direct; cf-2ag removed as over-reaching
- Test 5 now pins no-render invariant via `spy.not.toHaveBeenCalled()` on `#cotwError.hide` + `#cotwContainer.expand` (pr-test-analyzer's stronger-pin suggestion)

### Deferred / non-blocking
- code-simplifier's test dedup (`it.each`, collapse distinctness test) — same judgment as #1104: keep for readability, fold into a future consolidation pass if the pattern grows
- code-reviewer's pre-existing `progressStatus` JSDoc concern — not a drift; the return object at ~line 1355 DOES include `progressStatus`. No action needed.

**Tests: 69/69 pass** across `tests/challengeOfTheWeek.test.js` + `tests/challengeOfTheWeekWidget.test.js`.

Ready for merge gate.

— radahn (cfutons/crew/radahn)

DreadPirateRobertz (commented):
FINAL ARBITER: APPROVE.

All CI green: test (20)/(22) SUCCESS, lint, CodeQL, codecov/patch SUCCESS, hookup-assistant. cf-9lp.3 outer catch returns internal_error shape (continues .1 cascade pattern from radahn). Bar met. Admin-squash incoming.

PR #1108, opened 2026-04-18
fix(cf-9lp.4): challengeDiscovery chip discriminates internal_error

DreadPirateRobertz:
## Summary

Final slice of the **cf-9lp cascade** — surface `internal_error` discriminant to the challenge discovery chip.

- `src/public/challengeDiscovery.js` now branches on truthy `result.error` BEFORE treating `result.challenges` as empty. Before, the cf-tlt shape `{ challenges: [], error: 'internal_error' }` slipped through identically to a clean empty response — a silent-failure masquerade on every product/catalog page.
- UX preserved: chip stays hidden either way (ambient UI, peripheral to the page — intrusive error state would be hostile here). Observability restored: `console.error` logs the error code so ops can see the underlying failure class.
- Truthy-check (cf-8qc widget-layer precedent) so future error codes route through the same path automatically.

## Cascade context

Closes the final silent-hide entry point for the `{ error: 'internal_error' }` shape:

| Slice | Site | PR | Status |
|-------|------|-----|--------|
| cf-9lp.1 | http-functions `GET /activeChallenges` → 503 | #1102 | merged |
| cf-9lp.2 | `ChallengesDisplay.js` → error container | #1104 | merged |
| cf-9lp.3 | `getActiveChallengeOfWeek` outer catch | #1105 | merged |
| **cf-9lp.4** | **`challengeDiscovery.js` chip → hide + log** | **this PR** | **open** |

## Test plan

- [x] Local: `npx vitest run tests/challengeDiscovery.test.js` — **20/20 passing**
- [x] 5 new RED-first tests under `describe('initChallengeDiscoveryChip — cf-9lp.4 error discrimination')`:
  - hides chip + logs on `{ challenges: [], error: 'internal_error' }`
  - hides chip + logs even when `challenges` contains a match (no partial render)
  - routes any truthy error code through the same hide+log path (future-proof)
  - does NOT log on normal empty-but-authed response
  - does NOT log on happy-path render
- [x] Rebased on main cleanly
- [ ] CI green
- [ ] 5-agent review (code-reviewer, silent-failure-hunter, code-simplifier, comment-analyzer, pr-test-analyzer)
- [ ] melania admin-merge gate

Closes **cf-0z4**. Completes the cf-9lp umbrella.

### DIFF
--- a/src/public/challengeDiscovery.js
+++ b/src/public/challengeDiscovery.js
@@ -61,6 +61,18 @@ export async function initChallengeDiscoveryChip(elements, memberId, getActiveCh
 
   try {
     const result = await getActiveChallengesFn(memberId);
+
+    // cf-9lp.4: discriminate backend errors from a clean empty result. The
+    // cf-tlt shape `{ challenges: [], error: 'internal_error' }` (or any
+    // truthy error code) previously slipped through as an empty list, so a DB
+    // failure hid silently — indistinguishable from "no matching challenge".
+    // UX stays ambient (chip hides either way), but we log so ops can see it.
+    if (result?.error) {
+      console.error('[challengeDiscovery] backend returned error:', result.error);
+      _hideChip($chip);
+      return;
+    }
+
     const challenges = result?.challenges ?? [];
 
     const match = challenges.find(c =>
--- a/tests/challengeDiscovery.test.js
+++ b/tests/challengeDiscovery.test.js
@@ -194,3 +194,58 @@ describe('initChallengeDiscoveryChip', () => {
     expect(elements.$chipTitle.text).toBe('Second chance');
   });
 });
+
+// cf-9lp.4: chip previously treated `{ challenges: [], error: 'internal_error' }`
+// (cf-tlt shape) identically to a clean empty result — users' chip silently
+// hid on DB failure, indistinguishable from "no matching challenge". Cascade
+// continuation of cf-tlt (PR #1099), cf-9lp.1 (PR #1102), cf-9lp.2 (PR #1104),
+// cf-9lp.3 (PR #1105). UX note: chip stays hidden on error (ambient UI, no
+// intrusive error state), but observability improves — discriminate + log.
+describe('initChallengeDiscoveryChip — cf-9lp.4 error discrimination', () => {
+  let elements;
+  let mockGetActiveChallenges;
+  let errorSpy;
+
+  beforeEach(() => {
+    elements = makeElements();
+    mockGetActiveChallenges = vi.fn();
+    errorSpy = vi.spyOn(console, 'error').mockImplementation(() => {});
+    errorSpy.mockClear();
+  });
+
+  it('hides chip AND logs when result.error is "internal_error" (even with empty challenges)', async () => {
+    mockGetActiveChallenges.mockResolvedValue({ challenges: [], error: 'internal_error' });
+    await initChallengeDiscoveryChip(elements, 'mem-1', mockGetActiveChallenges, 'add_to_cart');
+    expect(elements.$chip.hide).toHaveBeenCalled();
+    expect(errorSpy).toHaveBeenCalled();
+  });
+
+  it('hides chip AND logs when result.error is truthy even with matching challenges (no partial render)', async () => {
+    mockGetActiveChallenges.mockResolvedValue({ challenges: [ADD_TO_CART_CHALLENGE], error: 'internal_error' });
+    await initChallengeDiscoveryChip(elements, 'mem-1', mockGetActiveChallenges, 'add_to_cart');
+    expect(elements.$chip.hide).toHaveBeenCalled();
+    expect(elements.$chip.show).not.toHaveBeenCalled();
+    expect(errorSpy).toHaveBeenCalled();
+  });
+
+  it('routes any truthy error code to the same hide+log path (future-proof)', async () => {
+    mockGetActiveChallenges.mockResolvedValue({ challenges: [], error: 'some-future-code' });
+    await initChallengeDiscoveryChip(elements, 'mem-1', mockGetActiveChallenges, 'add_to_cart');
+    expect(elements.$chip.hide).toHaveBeenCalled();
+    expect(errorSpy).toHaveBeenCalled();
+  });
+
+  it('does NOT log when result is a normal empty-but-authed response (no error field)', async () => {
+    mockGetActiveChallenges.mockResolvedValue({ challenges: [] });
+    await initChallengeDiscoveryChip(elements, 'mem-1', mockGetActiveChallenges, 'add_to_cart');
+    expect(elements.$chip.hide).toHaveBeenCalled();
+    expect(errorSpy).not.toHaveBeenCalled();
+  });
+
+  it('does NOT log on normal happy-path render', async () => {
+    mockGetActiveChallenges.mockResolvedValue({ challenges: [ADD_TO_CART_CHALLENGE] });
+    await initChallengeDiscoveryChip(elements, 'mem-1', mockGetActiveChallenges, 'add_to_cart');
+    expect(elements.$chip.show).toHaveBeenCalled();
+    expect(errorSpy).not.toHaveBeenCalled();
+  });
+});


DreadPirateRobertz (commented):
**code-reviewer agent (dispatched by radahn)**

## Verdict: LGTM (comment — cannot self-approve)

Final slice of the cf-9lp cascade lands cleanly. Targeted review against cfutons patterns and cascade precedent:

### Correctness
- **Error-ordering is right**: `if (result?.error)` sits inside the existing `try` block, BEFORE `const challenges = result?.challenges ?? []`. A cf-tlt-shaped `{ challenges: [], error: 'internal_error' }` can no longer slip through as an innocent empty array. The existing `catch` still handles thrown errors from `getActiveChallengesFn`; the new branch handles structured error returns. Both paths converge on `_hideChip + return` — no double-hide, no partial render.
- **Truthy check (`result?.error`) matches cascade precedent**: cf-9lp.3 (widget/consumer layer) uses the same pattern, while cf-9lp.1 uses strict equality at the HTTP boundary. This is the right layer for truthy — forward-compatible with any future `'rate_limited'`, `'timeout'`, etc. discriminator codes without another code change.
- **UX decision is defensible and well-documented**: ambient chip on product/catalog pages — intrusive error state would be hostile UX on every product view. Silent-hide is preserved; observability is added via `console.error`. Contrast with cf-9lp.2 (ChallengesDisplay) correctly cited in PR body — that component is page-primary, so intrusive surfacing is warranted there.

### Consistency with module
- `console.error('[challengeDiscovery] ...')` prefix matches the existing catch-block and `_showChip` error logs — ops log-grep stays clean.
- `_hideChip($chip); return;` mirrors the null-memberId and unknown-context branches. Control flow is uniform.

### Tests (5 RED-first tests, all tight)
- Canonical `internal_error` with empty challenges — the cf-tlt shape itself.
- **Partial-render guard**: error + non-empty challenges asserts `show` was NOT called. This is the important one — it proves we don't trust a poisoned payload.
- Future-proof: any truthy code routes to hide+log.
- Non-regression 1: clean empty `{ challenges: [] }` — hides WITHOUT logging. Critical for not polluting Sentry on normal "no active challenge" cases.
- Non-regression 2: happy path — shows chip, no log.
- `errorSpy` scoped in `beforeEach` with `mockImplementation(() => {})` keeps test output clean. No global spy leakage.

### Minor nit (non-blocking)
- `errorSpy.mockClear()` in `beforeEach` immediately after `vi.spyOn(...)` is redundant — the spy is fresh. Harmless. Not worth a revision.

### Cascade status
With this merge, all four cf-9lp sites (HTTP boundary → ChallengesDisplay → widget → challengeDiscovery chip) discriminate `internal_error` correctly. The silent-hide class of bug that masked DB failures as "empty-but-authed" across this surface area is closed.

Recommend merge.

DreadPirateRobertz (commented):
**silent-failure-hunter agent (dispatched by radahn)**

Audited the cf-9lp.4 fix. The new `result?.error` branch correctly closes the specific cf-tlt shape hole (`{ challenges: [], error: 'internal_error' }` no longer slips through as a clean empty). Good: observability restored via `console.error`, tests assert both hide AND log, and a negative test confirms happy-path does NOT log. UX asymmetry vs cf-9lp.2 is clearly documented.

**Residual silent-failure holes — hunt results:**

1. **CRITICAL — Error-shape coverage is narrow** (`challengeDiscovery.js:68`).
   The truthy check `result?.error` only catches one shape. These sibling error shapes still slip through as silent empty-hide:
   - `{ errors: [...] }` (plural, common REST/GraphQL convention)
   - `{ errorCode: 'DB_DOWN' }` / `{ errorMessage: '...' }`
   - `{ ok: false }` / `{ success: false }` / `{ status: 'error' }`
   - `{ challenges: null }` with no error field (ambiguous — could be DB-returned-null vs no-data)
   - `result === undefined` (the `?.` silently short-circuits to empty array; logs nothing)

   The PR description says \"any truthy error code\" is handled, and the test `routes any truthy error code to the same hide+log path (future-proof)` reinforces that intent — but \"future-proof\" is only true for the `.error` key. A future backend that returns `{ errors: [...] }` reintroduces the exact bug this PR fixes.

   **Recommendation**: broaden to
   ```js
   const hasError = result?.error || result?.errors || result?.errorCode || result?.ok === false || result?.success === false;
   if (hasError) { console.error('[challengeDiscovery] backend returned error:', result); _hideChip($chip); return; }
   ```
   Or lock the contract: log a warn when `result` shape is unexpected (e.g. missing `challenges` array entirely).

2. **MEDIUM — `result === null` is treated as silent empty** (`challengeDiscovery.js:64`).
   `result?.challenges ?? []` coalesces `null`/`undefined` result into `[]` with no log. The existing test `hides chip when getActiveChallenges returns null` locks this behavior in as \"correct\" — but a `null` return from the web method is almost always an error signal (timeout, 500, serialization crash). Consider logging when `result == null` since a well-behaved backend should return `{ challenges: [] }`, never bare `null`.

3. **MEDIUM — Pre-existing empty catch in `_hideChip`** (`challengeDiscovery.js:84`).
   ```js
   function _hideChip($chip) {
     try { $chip.hide(); } catch { /* element may not exist on this page */ }
   }
   ```
   Empty catch. Swallows ANY error from `$chip.hide()` — not just \"element doesn't exist\". Would hide a genuine Wix runtime regression. Not introduced by this PR but worth flagging: at minimum log at `console.debug` level, or narrow the catch to the specific missing-element shape.

4. **LOW — Broad outer catch** (`challengeDiscovery.js:77`).
   `catch (err)` catches everything thrown by `getActiveChallengesFn` AND by `_showChip` AND by `challenges.find`. A bug in the `.find` predicate (e.g. a challenge with a non-string `conditionType` crashing `.includes`) gets logged as \"initChallengeDiscoveryChip failed\" indistinguishable from a network error. Consider wrapping only the await, then letting render errors bubble to `_showChip`'s own catch. Confirmed: `_showChip failed` and `backend returned error` and `initChallengeDiscoveryChip failed` ARE discriminable by suffix — good.

5. **LOW — Test assertion weakness** (`challengeDiscovery.test.js:212,220,228`).
   `expect(errorSpy).toHaveBeenCalled()` doesn't assert the log content. A refactor that replaced `console.error('...backend returned error...', result.error)` with `console.error('')` (empty) would still pass. Stronger: `expect(errorSpy).toHaveBeenCalledWith(expect.stringContaining('backend returned error'), 'internal_error')`.

**Answers to the dispatch questions:**
- Q1 (thrown error still logs): Yes — line 78 still `console.error`s. Confirmed not a regression.
- Q2 (other slip-through shapes): Yes — see Finding 1. Significant gap.
- Q3 (discriminable logs): Yes — `_showChip failed` / `backend returned error` / `initChallengeDiscoveryChip failed` all have distinct suffixes.
- Q4 (silent DB failure at chip layer): Reduced but not eliminated. `{ errors: [...] }`, `null` result, and `{ ok: false }` still fail silently.
- Q5 (new silent failures introduced): None. The PR is strictly additive.

**Verdict**: Net improvement; ships safely. Finding 1 (broaden the shape check) is worth a follow-up bead before the next cascade stage — the \"future-proof\" claim in the test name is aspirational, not actual.

DreadPirateRobertz (commented):
**code-simplifier agent (dispatched by radahn)**

## Verdict: ship as-is (one minor nit, no blocking changes)

Reviewed the 12-line source diff and 58-line test diff against the existing `challengeDiscovery.js`. The implementation is already well-shaped; I found no simplifications that would preserve the load-bearing invariants (truthy-check, console.error, ambient-UI UX).

### Source file (`src/public/challengeDiscovery.js`)

1. **Comment density — justified, keep it.** The 5-line explanation captures three pieces of non-obvious context that cannot be recovered from the code alone:
   - The specific cf-tlt shape `{ challenges: [], error: 'internal_error' }` (otherwise `if (result?.error)` looks like a generic guard against nothing).
   - The silent-hide bug it prevents (log-only behavior would look over-defensive without this).
   - The UX decision: chip hides either way, but we log so ops can see DB failures.
   
   This is fundamentally different from cf-9lp.3's PR #1105 flag, which was about rephrased code-describes-itself comments. Here the comment encodes a product UX choice (why we preserved silent-hide instead of rendering an error state) that a future maintainer would otherwise second-guess and "fix" into an intrusive error banner. Load-bearing. Leave it.
   
   Possible trim: the phrase "(or any truthy error code)" is mildly redundant with "backend errors" in the opening — but it sets up the future-proof test below, so I wouldn't touch it.

2. **Folding the error branch into the no-match `_hideChip` path — do not do it.** A single `if (!match || result?.error)` would lose the `console.error` observability that is explicitly load-bearing per the PR brief. The two branches have legitimately different side-effects (log vs no-log) and must stay separate. Current shape is correct.

3. **`_hideChip` / `_showChip` helpers — no unnecessary wrapping.** `_hideChip` is a 1-line try/catch wrapper around `$chip.hide()` — the try/catch is load-bearing (page may not mount the element), and it's called from 5+ sites, so inlining would duplicate the defensive pattern. `_showChip` bundles three mutations under one try/catch with a consistent fallback to `_hideChip` — worth the helper. Leave both.

4. **Micro-nit (optional, non-blocking):** the inserted block could drop the blank line after the try-brace for tighter visual coupling with `const result = await …`, but the current spacing also reads fine. Taste call.

### Test file (`tests/challengeDiscovery.test.js`)

The 5 new tests cover distinct behaviors and I would not collapse any of them:

1. `error + empty challenges` — the canonical cf-tlt shape. Primary regression test. **Keep.**
2. `error + matching challenges (no partial render)` — guards against the fold-into-`!match` refactor I rejected above. Without this test, a future dev *would* collapse the branch and silently break partial-render protection. **Keep — high future value.**
3. `any truthy error code (future-proof)` — documents the truthy-check contract (cf-8qc pattern). Low runtime cost, pins the semantic. **Keep.**
4. `normal empty, no log` — negative assertion on the error path. Without it, test #1 could accidentally pass even if someone logged on every empty response. **Keep.**
5. `happy path, no log` — parallel negative for the show-chip path. Arguably the weakest — `errorSpy` assertion on a happy render is somewhat redundant with test #4's coverage of the log-only invariant. **Keep but noted as the most collapse-eligible if the PR later needs trimming.**

One small cleanup: `errorSpy.mockClear()` on line 208 is redundant — the `mockImplementation(() => {})` call on line 207 just created the spy, so there's nothing to clear. Removing the line is safe. Non-blocking.

### Summary
- Source diff: no simplifications recommended. Comment is load-bearing, branch separation is load-bearing, helpers are correctly factored.
- Test diff: no tests to collapse. One redundant `mockClear()` line is removable but not blocking.
- Invariants preserved: truthy-check (`result?.error`), console.error observability, ambient-UI silent-hide UX.

LGTM from a simplification standpoint. Ship it.

DreadPirateRobertz (commented):
**comment-analyzer agent (dispatched by radahn)**

## Summary
Audited the two new comment blocks (src/public/challengeDiscovery.js ~L63-71 and tests/challengeDiscovery.test.js ~L197-202). Both are accurate, well-scoped, and avoid the rot patterns flagged on cf-9lp.3 (PR #1105). No critical issues. Clean pass.

## Audit results against the five checkpoints

### 1. Line-number citations that will rot? ZERO
Both comment blocks use symbolic references only: `{ challenges: [], error: 'internal_error' }` shape and `result?.error` branch. No line numbers referenced — the cf-9lp.3 lesson clearly carried forward.

### 2. PR number cross-reference — ALL FOUR CORRECT
Verified via `gh pr view`:
- PR #1099 -> `fix/cf-tlt-db-error-shape` — matches "cf-tlt (PR #1099)" ✓
- PR #1102 -> `fix/cf-9lp.1-http-503` — matches "cf-9lp.1 (PR #1102)" ✓
- PR #1104 -> `fix/cf-9lp.2-challenges-display-error` — matches "cf-9lp.2 (PR #1104)" ✓
- PR #1105 -> `fix/cf-9lp.3-get-active-challenge-of-week` — matches "cf-9lp.3 (PR #1105)" ✓

No mismatches. Cascade chain is factually correct.

### 3. Cascade attribution scope — PROPERLY TIGHTENED
Attribution lists only direct cascade ancestors: cf-tlt (origin shape) -> cf-9lp.1 -> cf-9lp.2 -> cf-9lp.3 -> cf-9lp.4. No mention of cf-2ag (correctly excluded as grandparent at a different site). The cf-9lp.3 over-reach lesson was applied correctly here.

### 4. UX note accuracy — MATCHES CODE
Comment claims:
- "chip stays hidden on error (ambient UI, no intrusive error state)" — code calls `_hideChip($chip); return;` ✓
- "but observability improves — discriminate + log" — code calls `console.error('[challengeDiscovery] backend returned error:', result.error)` ✓
- "users' chip silently hid on DB failure, indistinguishable from 'no matching challenge'" — accurately describes the pre-fix behavior where `result?.challenges ?? []` swallowed the error shape into an empty list ✓

Test assertions align: `expect($chip.hide).toHaveBeenCalled()` + `expect(errorSpy).toHaveBeenCalled()` for error cases; `expect(errorSpy).not.toHaveBeenCalled()` for clean empty + happy paths. The "future-proof" test (`some-future-code`) correctly exercises the `result?.error` truthy check rather than a string-literal match.

### 5. "cf-tlt shape" reference clarity — CLEAR AND ACCURATE
The prose in the src comment spells out the shape inline: `{ challenges: [], error: 'internal_error' }`. A future maintainer does not need to grep for cf-tlt to understand the structure — the comment is self-contained while still providing the bead-id breadcrumb for deeper archaeology. This is the right balance.

## Minor observation (non-blocking)
The source comment's "(or any truthy error code)" parenthetical nicely pre-documents the `result?.error` truthy check, which the third test case ("future-proof") validates. Comment + test co-evolved well — no drift risk.

## Positive findings
- Symbolic references over line numbers (cf-9lp.3 correction fully internalized)
- Self-contained error-shape description in the src comment (no external doc dependency)
- Cascade attribution matches verified merged PR metadata
- UX rationale ("ambient UI, no intrusive error state") explains the *why* behind hide-on-error, which is non-obvious and genuinely useful for future maintainers
- Test describe-block header mirrors the src comment's terminology — grep for "cf-9lp.4" lands on both source and tests

## Critical Issues
None.

## Improvement Opportunities
None material. Comments are concise, accurate, and earn their place.

## Recommended Removals
None.

LGTM from comment-analyzer perspective — ship it.

DreadPirateRobertz (commented):
**pr-test-analyzer agent (dispatched by radahn)**

## Summary

Tight, focused suite covering the new error-discrimination branch. 20/20 passing, 5 behavioral assertions anchor the new code path. Overall coverage quality is good for a cf-9lp.4-sized slice, with a few gaps worth calling out and one test-hygiene concern.

## Critical Gaps (8-10)

None. The main failure modes (error + empty, error + match, future error code, happy empty, happy match) are covered.

## Important Improvements (5-7)

**1. Branch-ordering invariant not explicitly proven (criticality 7).**
The new `result?.error` check MUST fire *before* the challenges-array coalesce, or else a future refactor that moves the error-check after `.find(match)` would re-introduce the silent-failure. Test 2 (`error: 'internal_error'` + matching challenge with `$chip.show` not-called) is the closest thing, and it does cover the happy-path ordering risk. Recommend adding an explicit assertion that `$chipTitle.text` and `$chipProg.text` were NOT mutated — today's test only asserts `show` wasn't called, but a regression could still set title/progress before bailing and leave stale values on the DOM element. Proposed:
```js
expect(elements.$chipTitle.text).toBe('');
expect(elements.$chipProg.text).toBe('');
```
This hardens test 2 from ordering-regression without adding a new test.

**2. Error-value type coverage is narrow (criticality 6).**
Tests only use string error codes. Production contract per cf-tlt is `string`, but the `if (result?.error)` is truthy-check — so `error: true`, `error: 1`, `error: { code: 'x' }`, `error: []` (truthy!), `error: 0`, `error: ''`, `error: null` all behave differently. A single parametrised test for `error: 0` and `error: ''` (non-regression: these should NOT hide+log) would nail down the falsy-boundary. Today a future contributor could change `if (result?.error)` to `if (result?.error != null)` and silently break falsy-empty-string semantics.

**3. Missing: `result.error` alongside `null`/`undefined` challenges field (criticality 5).**
`{ error: 'internal_error' }` with no `challenges` key at all isn't tested. The optional-chaining + coalesce handles it today, but the error-branch should preempt — worth one line of coverage to lock down that ordering.

## Test Quality Issues

**4. `errorSpy.mockClear()` after `vi.spyOn` is a band-aid, not a fix (criticality 6).**
The fact that spy history leaked across tests means either (a) `console.error` is being spied *outside* `beforeEach` elsewhere in the file, or (b) vitest isn't auto-restoring mocks. I don't see (a) in the file — the existing `describe` block doesn't spy on console.error. More likely: `vi.spyOn` attaches to `console.error` and — without `mockRestore()` in `afterEach` — the spy persists into subsequent test files in the same vitest worker, and its `.mock.calls` array retains prior entries. The `mockClear()` on fresh-spy-creation is benign but suggests the real fix is an `afterEach(() => errorSpy.mockRestore())` OR setting `restoreMocks: true` in `vitest.config.js`. Without restore, other test files importing this module may see phantom `console.error` calls. **Recommend adding `afterEach(() => errorSpy.mockRestore())`** — the `mockClear()` can stay as defense-in-depth but is not the root fix.

**5. Tests 4 and 5 (no-log non-regression) are adequate but fragile (criticality 4).**
`expect(errorSpy).not.toHaveBeenCalled()` would fail loudly if anything else in the module logs an error during happy-path — e.g. the `_showChip` catch block if `$chipTitle.text =` throws. Today's stubs don't throw, so it's fine, but this assertion is coupled to "no other logs in this code path" rather than "the new branch didn't fire." Consider `expect(errorSpy).not.toHaveBeenCalledWith(expect.stringContaining('backend returned error'), expect.anything())` — targets the new log specifically and decouples from unrelated logging.

## Positive Observations

- Five tests, five distinct behavioral outcomes — DAMP naming, no over-specification.
- Test 2 (error + match, no partial render) is the standout: it proves the bug couldn't have been papered over by "error-but-still-render-the-match" — real behavioral regression-catcher.
- Test 3 (future-proof truthy code) correctly encodes the intent of the production truthy-check rather than hardcoding `'internal_error'` as the only discriminant.
- Tests 4 and 5 (non-regression happy/empty) are cheap and catch the most likely refactor mistake — someone tightening the `if` predicate and accidentally logging on every empty response.
- Cascade context in the describe-block comment links cf-tlt → cf-9lp.1 → .2 → .3 → .4 — excellent for future archaeologists.

## Verdict

Ship-ready. Recommend applying (1) and (4) as follow-up polish; (2) and (3) are optional but cheap. No blocking gaps.

DreadPirateRobertz:
## radahn synthesis — 5-agent review (cf-9lp.4)

| Agent | Verdict | Notes |
|-------|---------|-------|
| code-reviewer | **LGTM** | Error-ordering correct; truthy-check matches cf-9lp.3 precedent; UX asymmetry vs cf-9lp.2 well-justified. |
| silent-failure-hunter | **Ships safely (net improvement)** | No new silent failures. Residual holes noted (see below) are pre-existing or out-of-scope. |
| code-simplifier | **LGTM, ship as-is** | Comment is load-bearing (cf-tlt shape + UX decision); no collapse opportunities that preserve invariants. |
| comment-analyzer | **Clean pass** | Zero line-number citations, all PR refs verified, cf-2ag correctly excluded. cf-9lp.3 lessons applied. |
| pr-test-analyzer | **Ship-ready** | No blocking gaps. Test 2 (error+match, no partial render) is the standout regression-catcher. |

### Residual findings (non-blocking — deferred)

**silent-failure-hunter** flagged a CRITICAL-labeled concern: the `result?.error` check misses alternative shapes (`{ errors: [...] }`, `{ errorCode }`, `{ ok: false }`). This is **correct but out-of-scope for cf-9lp**. The cascade's contract is the specific cf-tlt shape `{ error: '<code>' }`. If the backend contract ever widens, all four cf-9lp consumers would need a coordinated update — worth its own follow-up bead, not a blocker here.

**pr-test-analyzer** + **code-simplifier** both noted the `errorSpy.mockClear()` line is redundant after a fresh `vi.spyOn`. I added it to fix a real test-pollution issue (20→20 passing). Deeper fix (`restoreMocks: true` in vitest config) is a cross-cutting cleanup — leaving as-is to avoid scope creep.

**pr-test-analyzer** suggested tightening test 2 to also assert `$chipTitle.text` / `$chipProg.text` weren't mutated (catches a future refactor that sets text before bailing on error). Reasonable tightening — can fold into follow-up if desired.

### Summary

All 5 agents unanimously **LGTM / ship-ready**. No change-requests. Cascade complete: cf-9lp.1 → .4 all shipping the same shape-stable error discriminant from backend through the four frontend consumers. Ready for melania admin-merge gate.

🤖 Posted by radahn per 5-agent PR review protocol (mayor directive 2026-03-08)

PR #1114, opened 2026-04-25
feat(cf-3qt.4.6): Velo /_functions/contactSubmissions HTTP endpoint

DreadPirateRobertz:
## Why
Phase 4.6 of cf-3qt: the carolina-futons-web Next.js app needs to submit the /contact form through the existing Velo path so notifications continue to flow to BUSINESS.email via the shared rate-limit + triggered-email plumbing — not via a parallel nodemailer SMTP transport on the Next side.

## What
New HTTP function in src/backend/http-functions.js:
- POST /_functions/contactSubmissions — JSON body { name, email, phone?, subject, message } → delegates to emailService.sendEmail (handles validation, sanitization, 3/hour rate limit, triggered email to SITE_OWNER_CONTACT_ID)
- OPTIONS /_functions/contactSubmissions — CORS preflight via shared backend/utils/cors

Origin allowlist already covers: prod Vercel domain, project-scoped Vercel domain, per-branch preview URLs, and localhost:3000.

## Coordination — Stilgar
Per cf-3qt.4.6 acceptance criteria: \"Stilgar receives test email through existing Velo path; coordinate with stilgar before flipping.\"

This PR adds the **endpoint only** (additive — does not change any existing behavior). The companion PR in carolina-futons-web will flip /contact actions.ts from nodemailer to fetch this endpoint. **Hold flipping until Stilgar confirms a real test email arrives via this path on staging.**

## Tests
+7 cases in tests/httpFunctionsCoverage.test.js: invalid JSON / missing message / invalid email / ACAO allowlist hit / ACAO miss + observability log / OPTIONS 204 for allowed / OPTIONS 403 for blocked. Full file: 92/92 green.

## Risk
Pure addition. Existing webMethod surface untouched; existing nodemailer path on cfw side untouched until companion PR.

## Follow-up
- carolina-futons-web PR: replace nodemailer in src/app/contact/actions.ts with fetch \`${WIX_VELO_SITE_URL}/_functions/contactSubmissions\`
- Drop nodemailer dep + SMTP env vars from cfw .env.example after companion PR ships

### DIFF
--- a/src/backend/http-functions.js
+++ b/src/backend/http-functions.js
@@ -2550,20 +2550,29 @@ export async function get_weeklyBlogDigestCron(request) {
 
 // ── /_functions/contactSubmissions ───────────────────────────────────
 //
-// cf-3qt.4.6: HTTP wrapper around emailService.sendEmail so external
-// front-ends (carolina-futons-web Next.js) can submit the contact form
-// through the same Velo path that powers the legacy Wix Studio site —
-// preserving the existing rate-limit + Stilgar-notification pipeline
-// and removing the standalone nodemailer SMTP dep on the Next side.
+// HTTP wrapper around emailService.sendEmail so external front-ends
+// (carolina-futons-web Next.js) submit the contact form through the
+// same Velo path that powers the legacy Wix Studio site — reusing the
+// existing rate-limit + site-owner triggered-email pipeline.
 //
-// Body shape mirrors emailService.sendEmail:
-//   { name, email, phone?, subject, message }
-// CORS: emits permissive headers via shared corsHeaders() helper so
-// browser-side fallback fetches work; primary caller is server-side.
+// CORS allowlist (see backend/utils/cors): prod Vercel domain,
+// project-scoped Vercel domain, branch preview URLs, localhost dev.
 
 /**
  * @function post_contactSubmissions
  * @route POST /_functions/contactSubmissions
+ * @param {Object} request.body.json
+ * @param {string} request.body.json.name    — required, ≤200 chars
+ * @param {string} request.body.json.email   — required, ≤254 chars, valid format
+ * @param {string} [request.body.json.phone] — optional, ≤20 chars
+ * @param {string} [request.body.json.subject] — optional, ≤300 chars
+ * @param {string} request.body.json.message — required, ≤2000 chars
+ * @returns {Promise<{status: number, body: string, headers: object}>}
+ *   200 { success: true } on send;
+ *   400 { success: false, error } on validation;
+ *   429 { success: false, error } on per-email rate limit (3/hour);
+ *   500 { success: false, error } on transport failure or sendEmail
+ *   returning a non-object (webMethod proxy fault).
  */
 export async function post_contactSubmissions(request) {
   const JSON_HEADERS = corsHeaders(request, { 'Content-Type': 'application/json' });
@@ -2572,7 +2581,8 @@ export async function post_contactSubmissions(request) {
     try {
       const bodyText = await request.body.text();
       body = JSON.parse(bodyText);
-    } catch (_) {
+    } catch (parseErr) {
+      console.warn('[contactSubmissions] body parse failed:', parseErr?.message ?? parseErr);
       return badRequest({ body: JSON.stringify({ success: false, error: 'Invalid JSON body' }), headers: JSON_HEADERS });
     }
 
@@ -2584,9 +2594,33 @@ export async function post_contactSubmissions(request) {
       message: body.message,
     });
 
-    if (!result || result.success !== true) {
-      return badRequest({
-        body: JSON.stringify({ success: false, error: result?.message || 'Submission rejected' }),
+    // Defensive: webMethod proxy can theoretically resolve to undefined on
+    // backend infrastructure failures. Treat as 500 — it's not a client bug.
+    if (!result) {
+      console.error('[contactSubmissions] sendEmail resolved without a result envelope');
+      return serverError({
+        body: JSON.stringify({ success: false, error: 'Internal server error' }),
+        headers: JSON_HEADERS,
+      });
+    }
+
+    if (result.success !== true) {
+      // Distinguish rate-limit (429) and transport failure (500) from
+      // validation (400). sendEmail's outer catch returns success:false
+      // with a "Failed to send message…" message — that is an infra
+      // outage, not a client validation error. Other handlers in this
+      // file already return 429 explicitly for rate-limit so the cfw
+      // client can back off (e.g., L1867, L2183, L2255).
+      const message = result.message ?? '';
+      const lowered = message.toLowerCase();
+      const status = lowered.includes('too many requests')
+        ? 429
+        : lowered.startsWith('failed to send')
+        ? 500
+        : 400;
+      return response({
+        status,
+        body: JSON.stringify({ success: false, error: message || 'Submission rejected' }),
         headers: JSON_HEADERS,
       });
     }
--- a/src/backend/http-functions.js
+++ b/src/backend/http-functions.js
@@ -7,6 +7,7 @@ import { resolveTierFromPoints } from 'backend/utils/loyaltyData';
 import { generateFeed } from 'backend/googleMerchantFeed.web';
 import { getImageUrl } from 'backend/utils/mediaHelpers';
 import { recordPriceSnapshots, checkWishlistAlerts } from 'backend/notificationService.web';
+import { sendEmail } from 'backend/emailService.web';
 import { triggerBrowseRecovery } from 'backend/browseAbandonment.web';
 import { triggerAbandonedCartRecovery, processEmailQueue, triggerReengagement, triggerPostPurchaseSequence, getCampaignAnalytics } from 'backend/emailAutomation.web';
 import { scanAndTriggerWinback, runReviewRequestEmails } from 'backend/marketingSequences.web';
@@ -2545,4 +2546,58 @@ export async function get_weeklyBlogDigestCron(request) {
       headers: JSON_HEADERS,
     });
   }
+}
+
+// ── /_functions/contactSubmissions ───────────────────────────────────
+//
+// cf-3qt.4.6: HTTP wrapper around emailService.sendEmail so external
+// front-ends (carolina-futons-web Next.js) can submit the contact form
+// through the same Velo path that powers the legacy Wix Studio site —
+// preserving the existing rate-limit + Stilgar-notification pipeline
+// and removing the standalone nodemailer SMTP dep on the Next side.
+//
+// Body shape mirrors emailService.sendEmail:
+//   { name, email, phone?, subject, message }
+// CORS: emits permissive headers via shared corsHeaders() helper so
+// browser-side fallback fetches work; primary caller is server-side.
+
+/**
+ * @function post_contactSubmissions
+ * @route POST /_functions/contactSubmissions
+ */
+export async function post_contactSubmissions(request) {
+  const JSON_HEADERS = corsHeaders(request, { 'Content-Type': 'application/json' });
+  try {
+    let body;
+    try {
+      const bodyText = await request.body.text();
+      body = JSON.parse(bodyText);
+    } catch (_) {
+      return badRequest({ body: JSON.stringify({ success: false, error: 'Invalid JSON body' }), headers: JSON_HEADERS });
+    }
+
+    const result = await sendEmail({
+      name: body.name,
+      email: body.email,
+      phone: body.phone,
+      subject: body.subject,
+      message: body.message,
+    });
+
+    if (!result || result.success !== true) {
+      return badRequest({
+        body: JSON.stringify({ success: false, error: result?.message || 'Submission rejected' }),
+        headers: JSON_HEADERS,
+      });
+    }
+
+    return ok({ body: JSON.stringify({ success: true }), headers: JSON_HEADERS });
+  } catch (err) {
+    console.error('HTTP function error (contactSubmissions):', err);
+    return serverError({ body: JSON.stringify({ success: false, error: 'Internal server error' }), headers: JSON_HEADERS });
+  }
+}
+
+export function options_contactSubmissions(request) {
+  return response(corsPreflight(request));
 }
--- a/tests/httpFunctionsCoverage.test.js
+++ b/tests/httpFunctionsCoverage.test.js
@@ -13,6 +13,11 @@ import {
   __setInsertError,
 } from './__mocks__/wix-data.js';
 import { __setSecrets } from './__mocks__/wix-secrets-backend.js';
+import {
+  __reset as crmReset,
+  __getEmailLog as crmEmailLog,
+  __failNextEmail as crmFailNextEmail,
+} from './__mocks__/wix-crm-backend.js';
 import { __setMember, __reset as resetMembers } from './__mocks__/wix-members-backend.js';
 import { __setAccount, __reset as resetLoyalty } from './__mocks__/wix-loyalty.v2.js';
 import {
@@ -928,9 +933,9 @@ describe('get_cleanupRateLimitCron', () => {
 describe('post_contactSubmissions', () => {
   const goodOrigin = 'https://carolina-futons-web.vercel.app';
 
-  function buildReq({ body, origin = goodOrigin }) {
+  function buildReq({ body, origin = goodOrigin } = {}) {
     return {
-      headers: origin ? { origin } : {},
+      headers: { origin },
       body: { text: async () => body },
     };
   }
@@ -938,6 +943,7 @@ describe('post_contactSubmissions', () => {
   beforeEach(() => {
     __setSecrets({ SITE_OWNER_CONTACT_ID: 'owner-1' });
     __seed('EmailRateLimit', []);
+    crmReset();
   });
 
   it('returns 400 for invalid JSON body', async () => {
@@ -948,10 +954,10 @@ describe('post_contactSubmissions', () => {
     expect(parsed.error).toMatch(/Invalid JSON/i);
   });
 
-  it('returns 400 when sendEmail rejects (missing required field)', async () => {
-    // sendEmail validates {name, email, subject, message} as required;
-    // omitting message reproduces a real client misuse without needing to
-    // mock the email transport.
+  it('returns 400 when sendEmail rejects required-field validation', async () => {
+    // sendEmail requires {name, email, message}; subject/phone are optional.
+    // Omitting `message` reproduces a real client misuse without engaging
+    // the triggeredEmails mock.
     const result = await post_contactSubmissions(
       buildReq({
         body: JSON.stringify({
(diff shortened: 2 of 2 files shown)


DreadPirateRobertz (commented):
**code-simplifier (dispatched by godfrey)**

## Grade: A-

PR is small, additive, and reads cleanly. The handler shape mirrors adjacent POST handlers (`post_addBundleToCart`, `post_submitQuestion`) and the test file follows existing conventions. A couple of small consistency / dead-code nits, no blockers.

### Findings

**1. (minor) Body parse style differs from the closest sibling — pick one**

`post_contactSubmissions` parses with `text() + JSON.parse`:
```js
const bodyText = await request.body.text();
body = JSON.parse(bodyText);
```

The closest public POST sibling, `post_submitQuestion` (line 1731), uses the equivalent one-liner:
```js
body = await request.body.json();
```

Behavior is identical (Wix `request.body.json()` rejects on parse failure, caught by the same try/catch). Switching matches the closer adjacent handler. `post_addBundleToCart` (line 1661) uses the longer form, so this is a coin-flip — but the one-liner is what `submitQuestion`, `answerQuestion`, `gamificationEvent`, `challengeProgress`, and `busEvent` all use, so it's the dominant convention in this file.

Not a shared-helper proposal — just a one-line shrink to match neighbors. Take it or leave it.

**2. (minor) Redundant nullish guard on `result`**

```js
if (!result || result.success !== true) {
  return badRequest({
    body: JSON.stringify({ success: false, error: result?.message || 'Submission rejected' }),
    ...
  });
}
```

`sendEmail` always returns an object (`{success: true}` or `{success: false, message}`) — never null/undefined per its JSDoc and code path. The `!result ||` branch is dead code, and `result?.message` inside the branch then assumes the same. Either:
- Drop the guard: `if (result.success !== true)` and `result.message`, OR
- Keep the defensive guard but stay consistent with `result?.message` (current state mixes `!result` defensiveness with non-defensive access on the next line).

I'd drop it — `sendEmail`'s contract is well-defined and tested.

**3. (minor) Dead branch in `buildReq` test helper**

```js
function buildReq({ body, origin = goodOrigin }) {
  return {
    headers: origin ? { origin } : {},
    body: { text: async () => body },
  };
}
```

No call site passes `origin: undefined` or `origin: null` — the default `goodOrigin` fires when omitted, and the two override cases pass real strings (`'https://attacker.example.com'`). The `origin ? {origin} : {}` ternary is unreachable. Simplify to `headers: { origin }`. (If you want to preserve the "no origin header" path for a future test, leave it — but nothing exercises it today.)

**4. (no action) Body field re-destructure is fine**

```js
const result = await sendEmail({
  name: body.name, email: body.email, phone: body.phone,
  subject: body.subject, message: body.message,
});
```

`sendEmail` itself destructures the same five fields, so `await sendEmail(body)` would behave identically and ignore extras. Both styles are defensible: the explicit form documents the wire contract at the boundary and prevents accidental forwarding of unexpected fields. Keep as-is.

**5. (no action — out of scope) Triple-pattern parse boilerplate**

`post_addBundleToCart`, `post_submitQuestion`, and now `post_contactSubmissions` all open with the same `let body; try { ... } catch (_) { return badRequest(...) }` block. That's 3+ existing call sites with the same pattern — a real `parseJsonBody(request)` helper would land cleanly. **Not in scope for this PR**, but worth a follow-up bead if someone touches this file again. (Per godfrey's bar: 2+ existing sites to justify extraction; we're at 3+.)

### Verdict

Ship it. The two source-side nits (#1, #2) and the test-side nit (#3) are quality-of-life only — none affect behavior or tests. If you fold them in, the diff stays trivially small.

DreadPirateRobertz:
**silent-failure-hunter (dispatched by godfrey)**

## Grade: B+

Endpoint follows established sibling patterns (`post_addBundleToCart`, `post_submitQuestion`) and improves on them in one place (always returns `success: false` in body shape — siblings only return `error`). No CRITICAL silent failures, but a few HIGH and MEDIUM issues worth addressing.

---

### HIGH — JSON.parse error fully discarded (`catch (_)`)

**Location:** `src/backend/http-functions.js:2578`

```js
} catch (_) {
  return badRequest({ body: JSON.stringify({ success: false, error: 'Invalid JSON body' }), headers: JSON_HEADERS });
}
```

The user gets a generic "Invalid JSON body" but the operator gets nothing. If the Next.js front-end starts shipping malformed payloads — wrong content-type, double-stringified bodies, BOM-prefixed UTF-8, truncated requests — there is zero log trail. The existing `console.error` only fires in the outer catch.

**Hidden errors this could mask:**
- `SyntaxError` with position info that would pinpoint a serializer bug on the caller side
- `TypeError` if `request.body.text()` itself blows up (e.g. body already consumed)
- Any non-`SyntaxError` thrown by `request.body.text()` — currently treated as "Invalid JSON" and reported as a 400 to the client when it's actually a 500-class server fault

**Recommendation:** Log at warn level and narrow the catch to only treat `SyntaxError` as 400. Anything else should bubble to the outer catch and return 500.

```js
} catch (parseErr) {
  if (parseErr instanceof SyntaxError) {
    console.warn('[contactSubmissions] Invalid JSON body:', parseErr.message);
    return badRequest({ body: JSON.stringify({ success: false, error: 'Invalid JSON body' }), headers: JSON_HEADERS });
  }
  throw parseErr; // fall through to outer catch + serverError
}
```

This is consistent with how `post_addBundleToCart` and `post_submitQuestion` are written today — meaning **the same defect exists in those siblings** and could be hardened in a follow-up.

---

### MEDIUM — `result?.message || 'Submission rejected'` masks the "no result returned" case

**Location:** `src/backend/http-functions.js:2592`

I audited every `return` path in `sendEmail` (`src/backend/emailService.web.js:116-185`). Every `success: false` branch attaches a `message`:
- L127 schema validation → `schemaErrors[0]`
- L136 invalid email → `'Invalid email address.'`
- L142 rate limit → `RATE_LIMIT_MESSAGE`
- L182 catch → `'Failed to send message. Please try calling us at (828) 252-9449.'`

So `result?.message || 'Submission rejected'` only fires the fallback when:
1. `result` is `null`/`undefined` (would require `webMethod` to return nothing — possible if the `webMethod` proxy itself fails)
2. A future contributor adds a `success: false` branch without a `message`

The optional chaining + `||` chain is hiding case (1), which is a 500-class condition (the email service did not respond), not a 400 user error. Returning `badRequest` here tells the user "your submission was rejected" when the actual problem is that our backend module call returned nothing.

**Recommendation:** Treat null-result as a server error explicitly:

```js
if (!result) {
  console.error('[contactSubmissions] sendEmail returned no result');
  return serverError({
    body: JSON.stringify({ success: false, error: 'Internal server error' }),
    headers: JSON_HEADERS,
  });
}
if (result.success !== true) {
  return badRequest({
    body: JSON.stringify({ success: false, error: result.message || 'Submission failed' }),
    headers: JSON_HEADERS,
  });
}
```

Bonus: the user-facing string from `sendEmail`'s outer catch ("Failed to send message. Please try calling us at (828) 252-9449.") is being returned with HTTP 400 — that's misleading. A failed triggered-email send is a 500, not a client error. Consider distinguishing `rate_limited` / validation (400) from transport failure (502/503) in the response status.

---

### MEDIUM — `console.error` instead of `logError` / Sentry

**Location:** `src/backend/http-functions.js:2599`

```js
} catch (err) {
  console.error('HTTP function error (contactSubmissions):', err);
```

I checked: the entire `http-functions.js` module uses `console.error` exclusively — there are zero `logError` or `errorIds` imports. So this PR is **consistent with the existing module convention**, not a new defect.

That said, contact form submissions are a revenue-adjacent flow (lead capture for a furniture retailer). A 500 here that doesn't page anyone means leads silently disappear into Wix logs that nobody monitors. Recommend either (a) following the project-wide `logError` + `constants/errorIds.ts` pattern as a follow-up bead covering all of `http-functions.js`, or (b) adding a TODO referencing such a bead.

For this PR alone: **acceptable, matches sibling pattern**. Don't block merge on it, but file the follow-up.

---

### LOW — `console.error` on JSON.parse-adjacent failures will show garbage `err`

If `request.body.text()` itself throws (network truncation, body already read), the outer catch logs `err` with no context about which header / origin / content-length triggered it. Consider including request metadata:

```js
console.error('HTTP function error (contactSubmissions):', err, {
  origin: request.headers?.origin,
  contentLength: request.headers?.['content-length'],
});
```

---

### Tests — minor gap

The test suite covers:
- Invalid JSON → 400
- Missing required field → 400 (via real schema)
- Invalid email → 400
- CORS allowlist on error path
- OPTIONS preflight allowlist

**Missing:**
- Success path (sendEmail returns `{success: true}`) — would catch a regression where the handler accidentally always returns `badRequest`
- Outer catch path — no test currently forces `request.body.text()` itself to throw, so the `serverError` branch at line 2599 has zero coverage. A test that passes `{ body: { text: () => Promise.reject(new Error('boom')) } }` would fix this.
- `result === null` / `result === undefined` from `sendEmail` — the `'Submission rejected'` fallback string is currently dead code per my audit above; either delete it or test it.

---

### Things done well

- Always returns `success: false` in the body alongside `error` — siblings only return `error`. This makes the response shape predictable for the Next.js client.
- CORS allowlist is enforced (verified by the disallowed-origin test).
- Outer catch is **specific in intent** (logs + 500), not a swallow-and-continue.
- Inner JSON.parse catch returns the right HTTP status (400) — only the lack of logging is the issue.
- Body shape mirrors `sendEmail` input, no field coercion sneaking in.

---

### Summary

Ship-able with the suggested follow-ups. The one issue I'd ask to fix in this PR is the **`result === null` masking**: that's a true silent failure where a backend outage looks like client-side validation. The JSON.parse log is a nice-to-have. The `logError` migration is repo-wide and out of scope.

DreadPirateRobertz (commented):
**code-reviewer (dispatched by godfrey)**

Reviewed: `src/backend/http-functions.js` (post_contactSubmissions, options_contactSubmissions @ ~L2549-2603) and `tests/httpFunctionsCoverage.test.js` (+7 tests). Small, focused additive endpoint that wraps `emailService.sendEmail` over HTTP for the carolina-futons-web Next.js front-end. CORS uses the shared `corsHeaders` / `corsPreflight` helpers, body-parse + try/catch follow the sibling pattern in `post_addBundleToCart` (L1655) and `post_submitQuestion` (L1725). Tests exercise invalid JSON, missing required field, invalid email, CORS echo on error, disallowed origin, and OPTIONS allow/deny — solid coverage for the surface.

### Important (confidence 80-85)

**1. Rate-limit responses are returned as HTTP 400 instead of 429** — `src/backend/http-functions.js:2587-2592`

`emailService.sendEmail` returns `{ success: false, message: 'Too many requests. Please try again later.' }` when `_checkEmailRateLimit` rejects (see `emailService.web.js:98,142`). The new handler maps every non-success result through a single `badRequest(...)` branch, so a rate-limited caller gets a 400 with the rate-limit text in `error`. That contradicts the convention used by every other rate-limited POST in this file — e.g. `post_recordGamificationEvent` at L1867, `post_engagementEvent` at L2183, `post_socialAction` at L2255 all explicitly `response({ status: 429, ... })`. The Next.js client (and any future caller) can't distinguish "your input was malformed" from "back off and retry" without status-code inspection.

Suggested fix — sniff the message (or, better, expose a `reason` field from `sendEmail`) and branch:
```js
if (!result || result.success !== true) {
  const msg = result?.message || 'Submission rejected';
  if (/too many requests/i.test(msg)) {
    return response({ status: 429, body: JSON.stringify({ success: false, error: msg }), headers: JSON_HEADERS });
  }
  return badRequest({ body: JSON.stringify({ success: false, error: msg }), headers: JSON_HEADERS });
}
```
Cleaner option: change `_checkEmailRateLimit` callers in `emailService` to surface `reason: 'rate_limited'` alongside `message`, then the handler can branch on `result.reason` without string-matching. Either way, please add a test asserting 429 on the rate-limit path (the test file already seeds `EmailRateLimit` in `beforeEach`, so triggering it via 4 sequential calls is straightforward).

### Notes / non-blocking

- **Response shape**: `{ success: true }` on the happy path discards anything `sendEmail` might add later. That's fine today (sendEmail also returns just `{ success: true }`), but if a follow-up adds a submission ID (useful for client-side dedup / analytics), this handler will silently drop it. Consider `JSON.stringify({ success: true, ...result })` or returning `result` verbatim now, to leave the door open without a future API change.
- **Error string passthrough**: `result?.message || 'Submission rejected'` propagates whatever `sendEmail` returns directly to the client, including the fallback "Failed to send message. Please try calling us at (828) 252-9449." That's intentional (the phone number is already public on the contact page), so this is informational, not a finding — but worth noting in the JSDoc that the error text is user-facing and may include PII-adjacent content like the store phone.
- **Tests coverage gap**: no positive-path test asserting `status: 200` + `{ success: true }` on a fully-valid submission. The existing tests rely on `sendEmail` rejecting, which means a regression that makes `sendEmail` always return `{ success: true }` (e.g. accidentally short-circuiting validation) wouldn't be caught here. Suggest one happy-path test that mocks `triggeredEmails.emailContact` (already mocked elsewhere in this file based on `__seed`/`__setSecrets` usage) and asserts the 200 + CORS headers.

CORS contract, allowlist behavior, OPTIONS 204/403 split, and origin-rejected logging via `_checkOrigin('preflight')` all look correct and consistent with `backend/utils/cors.js`. No data-leak concerns — the handler doesn't log request bodies and `sendEmail`'s audit log is appropriate.

### Grade: B+

Ship-blocker: none. Recommend addressing the 429 mapping before merge (small change, single test).

DreadPirateRobertz:
**pr-test-analyzer (dispatched by godfrey)**

## Grade: B+

Solid coverage on the failure surface (invalid JSON, schema-rejected, invalid email, CORS allow/deny on both POST + OPTIONS preflight). The negative-path matrix is genuinely good and would catch regressions in CORS allowlist drift, JSON parsing fallthrough, and any change that swaps the `result.success !== true` predicate for a looser truthy check. Held back from A by the missing happy path.

## Main gaps

### 1. No 200 / happy-path test (criticality 8/10)

This is the load-bearing assertion the endpoint exists for, and the test infra cost is essentially zero — every prerequisite is already satisfied in `beforeEach`:

- `__setSecrets({ SITE_OWNER_CONTACT_ID: 'owner-1' })` — already done
- `__seed('EmailRateLimit', [])` — already done
- `triggeredEmails.emailContact` — mocked at `tests/__mocks__/wix-crm-backend.js:24` with `_emailLog` capture and returns `{ success: true }` by default
- `wixData.insert('ContactSubmissions', ...)` — already mocked (other tests in the same file exercise CMS inserts)

The "test infra cost" you mentioned is already paid by your existing 4 tests. A happy-path assertion added to the same `describe` block would catch:

- Anyone changing the `ok({ body: JSON.stringify({ success: true }) })` shape (callers in carolina-futons-web parse this exact field)
- Anyone forgetting to `await sendEmail` (would silently return success before the email actually fired)
- Anyone changing the body shape passed to `sendEmail` (e.g. dropping `phone` mapping)
- Regressions to the ACAO header on the 200 path — your CORS tests only exercise the error branch, so a future refactor that splits `JSON_HEADERS` between success and error paths could ship with broken CORS on success and tests would still pass

Suggested addition (~10 lines):
```js
it('returns 200 with success:true and forwards body fields to sendEmail', async () => {
  const result = await post_contactSubmissions(
    buildReq({
      body: JSON.stringify({
        name: 'Stilgar', email: 'stilgar@example.com', phone: '828-555-0100',
        subject: 'Frame question', message: 'Do you ship to Hendersonville?',
      }),
    }),
  );
  expect(result.status).toBe(200);
  expect(JSON.parse(result.body)).toEqual({ success: true });
  expect(result.headers['Access-Control-Allow-Origin']).toBe(goodOrigin);
  // optional: import { __getEmailLog } and assert template + variables
});
```

### 2. No rate-limited path test (criticality 6/10)

You seed `EmailRateLimit` empty but never seed it full. The endpoint relies on `sendEmail` returning `{success:false, message:'Too many requests...'}` and the wrapper surfacing that as 400. A regression where someone changes the wrapper to `if (!result?.success && result?.message !== RATE_LIMIT_MESSAGE)` (e.g. trying to map rate-limit to 429) would silently break the contract. Worth one test that seeds 3 records for the email and asserts 400 + the rate-limit error string surfaces in `parsed.error`.

Note: a 429 status code would actually be more correct than 400 for rate-limiting — flagging for product/PR author consideration, not a test gap per se. If you ever change the status, your current test would catch the regression in the wrong direction (it locks in 400 for rate-limit).

### 3. No sendEmail-throws / 500 path test (criticality 5/10)

The `try/catch` around `sendEmail` returns `serverError(...)` with `Internal server error`. Currently untested. `sendEmail` itself wraps in try/catch and returns `{success:false}` rather than throwing, so the only realistic throw is from `await request.body.text()` inside the inner catch (already covered) or a future refactor that removes `sendEmail`'s outer try/catch. Lower priority but trivial — flip `_shouldFail = true` on the mock and verify 500 + `Internal server error` body shape.

### 4. OPTIONS preflight Allow-Methods content (criticality 4/10)

You assert `Access-Control-Allow-Methods` matches `/POST/`, which is good. Consider also asserting `OPTIONS` is present (the spec requires it; some CORS-strict browsers reject preflights that don't list OPTIONS in the allowed methods). One-line addition.

### 5. Wildcard preview origin not exercised (criticality 3/10)

`cors.js` has a `WILDCARD_ORIGIN_PATTERN` for per-branch Vercel preview URLs. None of the new tests hit a `carolina-futons-web-git-<branch>-...` origin. The existing test at `__TEST__` exports suggests the cors utility itself is tested elsewhere — but a single integration test that pings POST + OPTIONS with a preview-URL origin would catch a regex regression in the wrapper context. Skip if you've verified the cors utility has its own regex tests.

## Quality observations

- DAMP test names are descriptive ("returns 400 for invalid JSON body", "echoes Access-Control-Allow-Origin for allowlisted origin on error") — easy to read failure output
- Good reuse of the `buildReq` helper — refactor-resilient
- Comment explaining why "missing message" reproduces real client misuse without mocking the transport is exactly the right level of context
- Tests are behavioral, not implementation-coupled — would survive most internal refactors

## Bottom line

Ship-blocking: none. Recommend adding the happy-path test (item 1) before merge — it costs ~10 lines, no new mock infra, and locks in the contract that callers in carolina-futons-web actually depend on. Items 2-5 are polish.

DreadPirateRobertz (commented):
**comment-analyzer (dispatched by godfrey)**

**Grade: B-**

The new endpoint ships with above-average documentation effort (preamble box + JSDoc + an inline test rationale), but three concrete inaccuracies and a couple of rot vectors keep this out of the A range. None block merge; all are easy to fix in a follow-up.

**Critical Issues (factually wrong / misleading)**

- `src/backend/http-functions.js` ~L2554 — Preamble: *"CORS: emits permissive headers via shared corsHeaders() helper..."* `corsHeaders()` in `src/backend/utils/cors.js` is a strict allowlist (3 exact origins + a per-branch Vercel-preview regex). It's the opposite of permissive — and the new tests in this PR explicitly assert rejection of `https://attacker.example.com`. Suggested wording: *"CORS: emits headers via the shared corsHeaders() helper, which echoes ACAO only for allowlisted origins (production + Vercel preview URLs + localhost:3000)."*
- `src/backend/http-functions.js` ~L2552 — Preamble: *"preserving the existing rate-limit + Stilgar-notification pipeline."* "Stilgar" is internal Gas Town crew terminology; the recipient in `emailService.web.js` is consistently called the "site owner" / "store owner" (`SITE_OWNER_CONTACT_ID`). External Velo devs and future maintainers won't know who Stilgar is. Replace with "site owner" to match the rest of the codebase.
- `tests/httpFunctionsCoverage.test.js` ~L953 — Test rationale: *"sendEmail validates {name, email, subject, message} as required."* `subject` is **not** marked `required: true` in the schema (`src/backend/emailService.web.js` L121-125). Only `name`, `email`, and `message` are required. A maintainer who later trusts this list could write a sibling test that omits `subject` expecting a 400 and be surprised. Fix to `{name, email, message}`. The "omitting message reproduces a real client misuse" framing is fine — only the field list is wrong.

**Improvement Opportunities**

- `src/backend/http-functions.js` ~L2562 (JSDoc on `post_contactSubmissions`) — JSDoc only carries `@function` + a non-standard `@route`. The body-shape contract (`{name, email, phone?, subject?, message}`) currently lives only in the prose preamble, so IDE tooling and call-site hover text show nothing useful. Suggest moving (or duplicating) the body-shape and the response-shape (`{success: boolean, error?: string}`) into proper `@param` / `@returns` tags. `@route` is fine as documentation but isn't a real JSDoc tag — Velo doesn't render it.
- `src/backend/http-functions.js` ~L2548 — Preamble: *"removing the standalone nodemailer SMTP dep on the Next side."* This describes a transitional state at the moment of merge. Six months from now "removing" is wrong tense (the dep is either long gone or was never there). Reword to descriptive present: *"so the Next.js front-end calls this endpoint instead of bundling its own SMTP transport."*
- `src/backend/http-functions.js` ~L2557 — Preamble: *"primary caller is server-side"* under-sells the actual surface. The CORS allowlist explicitly includes `http://localhost:3000` and per-branch Vercel preview URLs, all of which are browser callers. Either drop the "fallback" framing or note that browser-side calls are first-class for dev/preview environments.

**Watch (not blocking, but flag for the bead)**

- The `cf-3qt.4.6:` prefix on the preamble is fine for now but will rot once the Phase 2 epic closes. The rationale stands on its own without the bead id — consider whether the tag earns its place. Same pattern showed up in two recent PRs (cf-y2l3 Stage 1 idempotency comment had to be corrected post-merge) so it's a known maintenance hazard.

**Positive Findings**

- Section-divider style (`// ── /_functions/contactSubmissions ──`) matches surrounding conventions in both `http-functions.js` and `httpFunctionsCoverage.test.js`. Good visual scannability.
- The preamble correctly documents the body shape with the optional-marker convention (`phone?`) — that contract just needs to migrate into the JSDoc to be discoverable from call sites.
- The CORS-rejection tests are paired with a clear behavioural assertion (`Access-Control-Allow-Origin` undefined for disallowed origin), which makes the security boundary self-documenting at the test layer — no comment needed there.

DreadPirateRobertz (commented):
Melania verdict: APPROVE — Velo /_functions/contactSubmissions endpoint, unblocks cfw #169. Squash.

PR #1121, opened 2026-04-26
feat(cf-nt2f): getWishlistByMemberId — admin-permissioned wishlist share endpoint

DreadPirateRobertz:
## Summary
- Adds `getWishlistByMemberId(memberId)` to `wishlistService.web.js` with `Permissions.Anyone`
- `suppressAuth` wixData query enables unauthenticated access for the share-link UX (cf-u89z)
- Per-memberId rate limit (30 req/min via `WishlistShareRateLimit` CMS collection) prevents brute-force enumeration
- Returns same `{success, items, total}` shape as `getWishlist` for easy cfw integration
- 8 new tests (49/49 total)

## Test plan
- [ ] 49/49 wishlistService tests pass
- [ ] Returns items for valid memberId without auth
- [ ] Blank memberId → `success: false`
- [ ] Rate-limited → `success: false`
- [ ] No cross-member leakage

### DIFF
--- a/src/backend/wishlistService.web.js
+++ b/src/backend/wishlistService.web.js
@@ -24,6 +24,7 @@ import { Permissions, webMethod } from 'wix-web-module';
 import wixData from 'wix-data';
 import { currentMember } from 'wix-members-backend';
 import { sanitize } from 'backend/utils/sanitize';
+import { checkRateLimit } from 'backend/utils/rateLimit';
 
 const WISHLIST_MAX_ITEMS = 100;
 
@@ -158,6 +159,50 @@ export const getWishlist = webMethod(
       };
     } catch (err) {
       console.error('[wishlistService] getWishlist error:', err);
+      return { success: false, items: [], total: 0 };
+    }
+  }
+);
+
+// ── getWishlistByMemberId ─────────────────────────────────────────
+
+/**
+ * Read another member's wishlist by memberId.
+ * Used by the /wishlist/[token] share view in cfw — the HMAC token is
+ * verified on the cfw side before calling this; we only rate-limit here
+ * to prevent brute-force memberId enumeration.
+ *
+ * @param {string} memberId - Target member's _id (decoded from cfw share token)
+ * @returns {Promise<{success: boolean, items: Array, total: number}>}
+ */
+export const getWishlistByMemberId = webMethod(
+  Permissions.Anyone,
+  async (memberId) => {
+    try {
+      const cleanId = sanitize(memberId, 36);
+      if (!cleanId) return { success: false, items: [], total: 0 };
+
+      // Rate-limit per memberId to prevent enumeration of arbitrary IDs.
+      const { allowed } = await checkRateLimit(
+        'WishlistShareRateLimit',
+        cleanId,
+        { max: 30, windowMs: 60_000 },
+      );
+      if (!allowed) return { success: false, items: [], total: 0 };
+
+      const result = await wixData.query('Wishlist')
+        .eq('memberId', cleanId)
+        .descending('addedAt')
+        .limit(WISHLIST_MAX_ITEMS)
+        .find({ suppressAuth: true });
+
+      return {
+        success: true,
+        items: result.items.map(_mapItem),
+        total: result.totalCount,
+      };
+    } catch (err) {
+      console.error('[wishlistService] getWishlistByMemberId error:', err);
       return { success: false, items: [], total: 0 };
     }
   }
--- a/src/backend/wishlistService.web.js
+++ b/src/backend/wishlistService.web.js
@@ -13,7 +13,7 @@
  *   memberId (Text, indexed)
  *   productId (Text, indexed)
  *   name (Text)
- *   price (Number) — current display price (updated on load)
+ *   price (Number) — display price at time of last explicit update (not auto-refreshed)
  *   priceAtAdd (Number) — price at time of wishlist add (never updated)
  *   imageUrl (Text)
  *   productSlug (Text)
@@ -23,7 +23,7 @@
 import { Permissions, webMethod } from 'wix-web-module';
 import wixData from 'wix-data';
 import { currentMember } from 'wix-members-backend';
-import { sanitize } from 'backend/utils/sanitize';
+import { sanitize, validateId } from 'backend/utils/sanitize';
 import { checkRateLimit } from 'backend/utils/rateLimit';
 
 const WISHLIST_MAX_ITEMS = 100;
@@ -179,7 +179,7 @@ export const getWishlistByMemberId = webMethod(
   Permissions.Anyone,
   async (memberId) => {
     try {
-      const cleanId = sanitize(memberId, 36);
+      const cleanId = validateId(memberId);
       if (!cleanId) return { success: false, items: [], total: 0 };
 
       // Rate-limit per memberId to prevent enumeration of arbitrary IDs.
@@ -188,7 +188,7 @@ export const getWishlistByMemberId = webMethod(
         cleanId,
         { max: 30, windowMs: 60_000 },
       );
-      if (!allowed) return { success: false, items: [], total: 0 };
+      if (!allowed) return { success: false, error: 'rate_limited', items: [], total: 0 };
 
       const result = await wixData.query('Wishlist')
         .eq('memberId', cleanId)
@@ -203,7 +203,7 @@ export const getWishlistByMemberId = webMethod(
       };
     } catch (err) {
       console.error('[wishlistService] getWishlistByMemberId error:', err);
-      return { success: false, items: [], total: 0 };
+      return { success: false, error: 'server_error', items: [], total: 0 };
     }
   }
 );
--- a/tests/wishlistService.test.js
+++ b/tests/wishlistService.test.js
@@ -1,6 +1,6 @@
 /**
  * @file wishlistService.test.js
- * @description Tests for CF-ne83 Wishlist Service — CRUD + priceAtAdd tracking.
+ * @description Tests for CF-ne83 + CF-nt2f Wishlist Service — CRUD + priceAtAdd + getWishlistByMemberId.
  *
  * Tests:
  *  - addToWishlist: adds item, stores priceAtAdd, rejects invalid input, dedup, cap
@@ -11,7 +11,7 @@
  *  - priceAtAdd tracking: field persisted and immutable on add
  */
 import { describe, it, expect, beforeEach } from 'vitest';
-import { __seed, __onInsert, __onUpdate, __onRemove, __getInserted, __getUpdated, __getRemoved } from './__mocks__/wix-data.js';
+import { __seed, __onInsert, __onUpdate, __onRemove, __getInserted, __getUpdated, __getRemoved, __setQueryError, __reset as __wixDataReset } from './__mocks__/wix-data.js';
 import { __setMember } from './__mocks__/wix-members-backend.js';
 import { hashRateLimitKey } from '../src/backend/utils/rateLimit.js';
 import {
@@ -445,7 +445,7 @@ describe('getWishlistByMemberId', () => {
     expect(result.success).toBe(false);
   });
 
-  it('returns success:false when rate limit is exceeded', async () => {
+  it('returns success:false with error:rate_limited when rate limit is exceeded', async () => {
     __seed('WishlistShareRateLimit', [{
       _id: 'rl-1',
       key: hashRateLimitKey('member-99'),
@@ -454,7 +454,22 @@ describe('getWishlistByMemberId', () => {
     }]);
     const result = await getWishlistByMemberId('member-99');
     expect(result.success).toBe(false);
+    expect(result.error).toBe('rate_limited');
     expect(result.items).toHaveLength(0);
+  });
+
+  it('resets rate limit after window expires and allows access', async () => {
+    __seed('Wishlist', [makeWishlistItem({ memberId: 'member-99' })]);
+    // Exhausted window that started 2 minutes ago — past the 60s window
+    __seed('WishlistShareRateLimit', [{
+      _id: 'rl-1',
+      key: hashRateLimitKey('member-99'),
+      count: 30,
+      windowStart: Date.now() - 120_000,
+    }]);
+    const result = await getWishlistByMemberId('member-99');
+    expect(result.success).toBe(true);
+    expect(result.items).toHaveLength(1);
   });
 
   it('allows access when rate limit is not yet exceeded', async () => {
@@ -493,6 +508,16 @@ describe('getWishlistByMemberId', () => {
     const result = await getWishlistByMemberId('member-99');
     expect(result.items).toHaveLength(1);
     expect(result.items[0].id).toBe('wl-a');
+  });
+
+  it('returns success:false with error:server_error when wixData throws', async () => {
+    __setQueryError('Wishlist', new Error('wixData internal error'));
+    const result = await getWishlistByMemberId('member-99');
+    expect(result.success).toBe(false);
+    expect(result.error).toBe('server_error');
+    expect(result.items).toHaveLength(0);
+    // Reset query error so later tests are not affected
+    __wixDataReset();
   });
 });
 
--- a/tests/wishlistService.test.js
+++ b/tests/wishlistService.test.js
@@ -13,10 +13,12 @@
 import { describe, it, expect, beforeEach } from 'vitest';
 import { __seed, __onInsert, __onUpdate, __onRemove, __getInserted, __getUpdated, __getRemoved } from './__mocks__/wix-data.js';
 import { __setMember } from './__mocks__/wix-members-backend.js';
+import { hashRateLimitKey } from '../src/backend/utils/rateLimit.js';
 import {
   addToWishlist,
   removeFromWishlist,
   getWishlist,
+  getWishlistByMemberId,
   isOnWishlist,
   updateWishlistStock,
   _WISHLIST_MAX_ITEMS,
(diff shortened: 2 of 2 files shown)


DreadPirateRobertz (commented):

Overall: clean addition. The new `getWishlistByMemberId` mirrors the existing `getWishlist` shape consistently, naming is clear, and the rate-limit + sanitize guards are appropriate. A few simplification opportunities, all optional/non-blocking:

### 1. Shared query helper for `getWishlist` / `getWishlistByMemberId` (medium value)
The two functions have nearly identical bodies — only the query options (`suppressAuth`) and the source of `memberId` differ. Extracting a small private helper would remove duplication and make future changes (e.g. adding a new field to the response) a one-line edit:

```js
async function _queryWishlistByMemberId(memberId, { suppressAuth = false } = {}) {
  const result = await wixData.query('Wishlist')
    .eq('memberId', memberId)
    .descending('addedAt')
    .limit(WISHLIST_MAX_ITEMS)
    .find(suppressAuth ? { suppressAuth: true } : undefined);
  return {
    success: true,
    items: result.items.map(_mapItem),
    total: result.totalCount,
  };
}
```

Then both webMethods reduce to ~5 lines of orchestration (auth/sanitize/rate-limit + call helper). Not blocking — current duplication is small and readable.

### 2. Test helper for rate-limit seeding (small value)
The two rate-limit tests repeat the seed shape:

```js
__seed('WishlistShareRateLimit', [{
  _id: 'rl-1',
  key: hashRateLimitKey('member-99'),
  count: <n>,
  windowStart: Date.now() - 1000,
}]);
```

Consider a `seedRateLimit(memberId, count)` helper alongside `makeMember` / `makeWishlistItem` to keep intent visible (the meaningful input is just `count`). Cheap readability win.

### 3. `success: false` for blank/null memberId — consistency note (very small)
`getWishlistByMemberId` returns `success: false` when sanitize yields empty, which is more strict than `getWishlist`'s unauthenticated path (also `success: false`). That's correct here — but worth a one-line comment saying \"empty memberId is treated as caller error, not empty result\" so future readers don't accidentally relax it to `success: true, items: []`.

### 4. `cleanId` length cap of 36 (sanity check, not simplification)
36 fits a UUID. If Wix `_id` ever exceeds that (it has historically been longer in some collections), valid IDs would silently truncate to a non-matching string and return empty. Worth confirming against actual `member._id` length in the wild — but out of scope for code-simplifier.

### 5. Minor: `result.totalCount` semantics with `.limit()`
Both `getWishlist` and the new function return `total: result.totalCount` after a `.limit(WISHLIST_MAX_ITEMS)` call. In Wix data, `totalCount` is the *unbounded* count, so `total` can exceed `items.length` if a member ever has >100 items. Pre-existing behavior — flagging only because the new function inherits it. If shared-view UIs want \"items shown\", `items.length` may be the right field.

### Naming / style nits
- `_WISHLIST_MAX_ITEMS` export with leading underscore is a common test-only escape hatch — fine, but a `// test-only` comment on the export line would make intent obvious.
- `_mapItem` private convention is consistent throughout, good.

Nothing blocking. The PR can ship as-is; items 1 and 2 are the highest-leverage follow-ups if you want to tighten.

DreadPirateRobertz (commented):

Overall the `getWishlistByMemberId` suite is solid — 8 tests cover the happy path, empty result, blank/null input, rate-limit block, under-limit access, item shape, and member isolation. No spurious implementation-coupling. Below are gaps worth addressing before merge.

---

## Critical Gaps

**1. Rate-limit window expiry is not tested (criticality: 9/10)**

`checkRateLimit` resets the counter when `windowAge > windowMs`. The existing rate-limit test seeds a record with `count: 30` and `windowStart: Date.now() - 1000` (1 second ago), so the window has NOT expired and the block is correct. But there is no test covering the case where the window *has* expired — i.e., a record with `count: 30` and `windowStart: Date.now() - 61_000` (61 seconds ago, past the 60-second window). In that scenario the counter should reset and the call should be **allowed**, not blocked. Without a test for this, a regression that accidentally drops the window-expiry branch (e.g. a `>=` vs `>` change in `checkRateLimit`) would silently block every legitimate share-link visitor after the first 30 requests, permanently. Suggested test:

```js
it('allows access after the rate-limit window has expired', async () => {
  __seed('Wishlist', [makeWishlistItem({ memberId: 'member-99' })]);
  __seed('WishlistShareRateLimit', [{
    _id: 'rl-1',
    key: hashRateLimitKey('member-99'),
    count: 30,
    windowStart: Date.now() - 61_000, // window expired
  }]);
  const result = await getWishlistByMemberId('member-99');
  expect(result.success).toBe(true);
  expect(result.items).toHaveLength(1);
});
```

**2. `suppressAuth: true` is not verified (criticality: 8/10)**

The production query passes `{ suppressAuth: true }` to `wixData.find()`. If that option is accidentally dropped, the query would fall back to the caller's permissions — which for `Permissions.Anyone` callers means a potential permission error or empty result depending on Wix collection rules. No test currently asserts that `suppressAuth` is forwarded. The wix-data mock should capture the options object so a test can assert:

```js
it('queries wixData with suppressAuth:true', async () => {
  let capturedOpts = null;
  __onFind((_collection, opts) => { capturedOpts = opts; });
  __seed('Wishlist', [makeWishlistItem({ memberId: 'member-99' })]);
  await getWishlistByMemberId('member-99');
  expect(capturedOpts).toMatchObject({ suppressAuth: true });
});
```

If the mock does not expose `__onFind`, this is the right moment to add it — `suppressAuth` is a security-relevant option that should be regression-protected.

**3. wixData throw is not tested (criticality: 8/10)**

The `catch` block returns `{ success: false, items: [], total: 0 }` on DB error, but no test exercises this path. A future refactor that accidentally re-throws or changes the shape would go undetected. Suggested test:

```js
it('returns success:false when wixData.query throws', async () => {
  // arrange: valid memberId, rate limit clear, but DB explodes
  __seedThrow('Wishlist', new Error('DB unavailable'));
  const result = await getWishlistByMemberId('member-99');
  expect(result.items).toEqual([]);
  expect(result.total).toBe(0);
});
```

---

## Important Improvements

**4. memberId truncation at 36 characters is not tested (criticality: 6/10)**

`sanitize(memberId, 36)` will silently truncate a 37+-character ID to 36 chars, causing a false empty result rather than a validation error. A share link containing a malformed or legacy long ID would look like "no wishlist" to the visitor. Test:

```js
it('returns success:false (or empty) for a memberId longer than 36 chars', async () => {
  const result = await getWishlistByMemberId('a'.repeat(37));
  // sanitize truncates to 36 chars — no matching member, so empty is acceptable,
  // but success should be true (not an error) and items empty
  expect(result.items).toHaveLength(0);
});
```

This also acts as documentation that the 36-char limit is intentional and matches a UUID length assumption.

**5. Rate-limit counter increment is not verified (criticality: 5/10)**

The "allows access when rate limit is not yet exceeded" test (count: 5) checks that the call succeeds, but does not assert that the counter was incremented to 6 afterward. If the update path were accidentally omitted, the rate limit would never accumulate and the `max: 30` guard would be effectively disabled. An `__onUpdate` capture asserting `count: 6` would close this.

---

## Test Quality Observations

- The existing tests are behavior-focused and avoid over-specifying internal query chain calls — good.
- Seeding `WishlistShareRateLimit` in the nested `beforeEach` is correct isolation; no cross-suite pollution risk.
- Using `hashRateLimitKey` directly in tests is appropriate — it ties tests to the same hashing contract the production code uses, and if the hash function changes the tests will correctly break.
- The `null` memberId test (item 4) correctly reaches the `sanitize` branch and returns `{ success: false }` — well caught.

---

**Summary**: Two security-relevant gaps (window expiry reset and `suppressAuth` verification) are the blockers. The error path test is also important for silent-failure prevention. The other two are nice-to-have. Recommend addressing items 1–3 before merge.

DreadPirateRobertz (commented):

Reviewed `/Users/hal/gt/cfutons/src/backend/wishlistService.web.js` and `/Users/hal/gt/cfutons/tests/wishlistService.test.js`.

---

## Critical Issues

**`wishlistService.web.js` lines 1-22 — `@setup` schema comment claims price is updated on load**

The `price` field is annotated:
```
price (Number) — current display price (updated on load)
```
Nothing in this file updates `price` after insert. `addToWishlist` writes it once at insert time and never touches it again. `getWishlist` and `getWishlistByMemberId` are read-only. There is no update-on-load path. This claim is factually wrong and will mislead any maintainer who tries to rely on `price` being fresh — they may skip building an explicit refresh step because the comment implies it already happens.

Suggestion: Change to `price (Number) — price at time of last explicit update (not auto-refreshed on read)` or remove the parenthetical entirely if the field is in fact a historical snapshot like `priceAtAdd`.

---

**`wishlistService.test.js` line 3 — wrong bead reference in `@description`**

```js
@description Tests for CF-ne83 Wishlist Service
```
The PR branch is `feat/cf-nt2f-wishlist-get-by-member-id`. The bead tag `CF-ne83` does not match. This is a copy-paste error from a previous test file. Any future developer tracing a regression to this test file and searching for the originating bead will be sent to the wrong ticket.

Suggestion: Change `CF-ne83` to `CF-nt2f`.

---

## Improvement Opportunities

**`wishlistService.web.js` line 63 — `// Check cap` comment covers only half of what the query result is used for**

The query at line 64 fetches *all* of the member's wishlist items. The result is immediately used for two distinct purposes: (1) finding an exact duplicate at line 69, and (2) checking against the cap at line 74. The comment only mentions the cap, so a reader scanning the code sees "Check cap" and may be confused why a dup-check follows without any additional query.

Suggestion: Replace `// Check cap` with something like `// Fetch all items once — used for both dup-check (line 69) and cap enforcement (line 74)`.

**`wishlistService.web.js` lines 169-177 — `@param memberId` description assumes a specific caller**

```
@param {string} memberId - Target member's _id (decoded from cfw share token)
```
The parenthetical is useful context for the current integration, but it encodes a caller-side implementation detail as part of the parameter contract. If the function is ever called by a different path (e.g., an admin tool or a future API), the description is misleading.

Suggestion: State the type constraint separately from the caller assumption: `@param {string} memberId - Wix member _id to look up. In the share-link flow this is decoded from the HMAC share token by the caller before passing here.`

---

## Recommended Removals

None. No purely redundant WHAT-only comments found — all inline comments add at least some navigational or rationale value.

---

## Positive Findings

**`wishlistService.web.js` lines 169-177 — security rationale on `getWishlistByMemberId`** is well-written. It explains *why* `Permissions.Anyone` is safe (HMAC verified upstream), *why* rate-limiting is scoped to `memberId` (enumeration defense), and *who* calls this function — three distinct pieces of WHY that are not derivable from the code alone. This is the strongest comment in the file.

**`wishlistService.web.js` line 185** — `// Rate-limit per memberId to prevent enumeration of arbitrary IDs.` is a clean one-line WHY comment that pairs with the JSDoc above without duplicating it.

**`wishlistService.web.js` lines 1-22 — module-level `@setup` block** is genuinely useful: documenting the required CMS schema fields and their semantics (especially the `priceAtAdd` immutability invariant) gives future maintainers information that cannot be read from code alone. Aside from the inaccurate `(updated on load)` annotation flagged above, this block is a strong example of the right kind of documentation.

DreadPirateRobertz (commented):

Reviewed `src/backend/wishlistService.web.js` (new `getWishlistByMemberId` export) and the 8 new tests in `tests/wishlistService.test.js`.

## Summary
The implementation is clean, follows existing module conventions (`_mapItem`, error-swallowing pattern, shape parity with `getWishlist`), and the test coverage is solid (auth-free read, blank/null memberId rejection, rate-limit boundary, item shape, cross-member isolation). Below are the findings.

## Critical (none)

## Important

### 1. `sanitize(memberId, 36)` is length-only — does not constrain to UUID charset (confidence 82)
`sanitize()` strips HTML tags / decodes entities and slices to 36 chars; it does not enforce the GUID/UUID character set that Wix member `_id`s actually use. A junk string like `abc<x>def` becomes `abcdef` after sanitization and is then sent to `wixData.query('Wishlist').eq('memberId', cleanId)`. That query will safely return empty, so this is not a security bug today — but it does let an attacker burn rate-limit budget on garbage IDs that look syntactically valid to us. The codebase already exports `validateId(id, 50)` and `validateSlug(slug)` from `backend/utils/sanitize` for exactly this reason.

**Suggested fix:** either use `validateId(memberId, 36)` (returns `null` for invalid IDs) or add a UUID regex check after sanitization:
```js
const cleanId = sanitize(memberId, 36);
if (!cleanId || !/^[a-zA-Z0-9-]{8,36}$/.test(cleanId)) {
  return { success: false, items: [], total: 0 };
}
```

### 2. No per-member share opt-in / revocation mechanism (confidence 80)
The model relies entirely on the cfw-side HMAC token being unguessable. There is no `shareEnabled` flag on the member or wishlist record, and no `shareTokenVersion` to bump for revocation. Practical consequences:
- If a member's `/wishlist/[token]` URL leaks (shared in chat, indexed by a misconfigured cache, screenshotted on social), there is no way to revoke that single token without rotating the HMAC secret server-wide and invalidating every share link.
- A member who was once shared cannot opt out without losing access for everyone they intentionally shared with.

This is a product/architecture concern more than a code bug, but worth raising before this ships. At minimum, document the revocation story in the JSDoc (currently only says "HMAC token is verified on the cfw side"). If a `shareTokenVersion` field is added to the Members schema and incorporated into the HMAC payload, per-member revocation becomes a one-line increment.

### 3. Rate-limit fails open on DB errors (confidence 80, documented behavior)
`checkRateLimit()` returns `{ allowed: true }` if the rate-limit collection query/insert throws (line 98-101 of `rateLimit.js`). For a standard rate-limit on a member-authenticated endpoint that's the right call, but for an **enumeration-prevention** control on an anonymous endpoint, fail-open means a DB partial-outage gives an attacker an unbounded enumeration window. Consider passing an option to `checkRateLimit` (or branching here) that fails closed on DB errors for `Permissions.Anyone` reads. At a minimum, add an alert on rate-limit DB error spikes so a sustained outage is visible.

## Nits (not blocking)

- `result.totalCount` is returned as `total`, but `items` is capped at `WISHLIST_MAX_ITEMS` (100). If a member somehow has >100 items (shouldn't happen given `addToWishlist`'s cap, but the share view trusts this), the consumer will see "100 items of N total" inconsistency. Existing `getWishlist` has the same shape, so this is for consistency, not a new bug.
- Test on line 449-454 seeds `count: 30` (= max) so the next call should be rejected — this correctly verifies the boundary. Consider adding a test for `count: 29` (one below max) returning `allowed: true` to nail down the off-by-one. The "allows access when rate limit is not yet exceeded" test uses `count: 5`, which is fine but doesn't exercise the boundary.
- Consider logging `[wishlistService] getWishlistByMemberId rate-limited for memberId=<hashed>` when `!allowed` — silent rate-limit rejection makes incident triage harder. The other webMethods log on error; the rate-limit branch currently logs nothing.

## What's correct (verified)

- `Permissions.Anyone` is appropriate — HMAC boundary is on the cfw side, as documented.
- `suppressAuth: true` is required for the anonymous read against a member-scoped CMS collection.
- Per-memberId rate limiting (30/min) is a reasonable enumeration deterrent.
- `sanitize(memberId, 36)` matches the typical Wix member `_id` length (UUID = 36 chars including hyphens).
- Cross-member isolation is correctly tested (`does not leak items belonging to other members`).
- Item mapping uses the same `_mapItem` helper as `getWishlist`, ensuring shape parity.
- Hashed rate-limit key (via `hashRateLimitKey`) means the collection doesn't store plaintext memberIds.

DreadPirateRobertz (commented):

Reviewed `/Users/hal/gt/cfutons/src/backend/wishlistService.web.js` and `/Users/hal/gt/cfutons/tests/wishlistService.test.js` for silent failures, error swallowing, and rate-limit bypass edge cases.

---

## CRITICAL: All catch blocks use `console.error` instead of `logError`

**Location:** `wishlistService.web.js` lines 92-95, 128-131, 160-163, 204-207, 235-238, 274-277

**Issue:** Every catch block in this file calls `console.error(...)` directly instead of the project-standard `logError` from `backend/utils/errorHandler`. The `logError` utility is the Sentry/monitoring pipeline entry point — `console.error` goes only to runtime logs and is invisible to error aggregation. The `rateLimit.js` utility in the same PR correctly imports and uses `logError`. This file does not.

**Hidden errors this will miss in Sentry/dashboards:**
- `wixData.insert` permission denied (bad Wix collection permissions setup)
- `wixData.query` network timeout on the wishlist share endpoint (Permissions.Anyone, highest-traffic path)
- `currentMember.getMember()` SDK internal error distinct from unauthenticated state
- `checkRateLimit` throwing unexpectedly (it has its own fail-open try/catch but could still surface)

**User impact:** Errors that should trigger Sentry alerts and P1 pages go silently into runtime logs only. Ops team has no visibility. Debugging a production outage in `getWishlistByMemberId` will require log scraping instead of Sentry grouping.

**Recommendation:** Replace all six `console.error(...)` calls with `logError(context, err)` and add the import:

```js
// ...
} catch (err) {
  logError('wishlistService.addToWishlist', err);
  return { success: false, error: 'Failed to add to wishlist.' };
}
```

---

## HIGH: `getWishlistByMemberId` rate-limit failure returns no `error` field, caller cannot distinguish rate-limited from empty wishlist

**Location:** `wishlistService.web.js` lines 191-192

**Issue:**
```js
if (!allowed) return { success: false, items: [], total: 0 };
```
This returns the exact same shape as a valid empty wishlist (`success: true, items: [], total: 0`) except with `success: false` — but it also returns the same shape as the `!cleanId` guard on line 183. The calling cfw share-view page has no way to differentiate "rate limited, show a 429 message" from "memberId invalid" from "member has no wishlist items" (which is `success: true` but `items: []`). Consumers will default to showing a blank page with no feedback.

There is also no `error` field on the rate-limit path, while every other `success: false` return in this file includes `error: '...'`.

**Recommendation:**
```js
if (!allowed) return { success: false, items: [], total: 0, error: 'Too many requests. Please try again later.' };
// and for the missing cleanId path:
if (!cleanId) return { success: false, items: [], total: 0, error: 'Invalid member ID.' };
```

**Test gap:** The test at line 448 (`returns success:false when rate limit is exceeded`) only asserts `result.success === false` and `result.items` length. It does not assert that `result.error` is set to something actionable, so the missing `error` field goes undetected.

---

## HIGH: Rate-limit uses `memberId` as the bucket key — an attacker can supply arbitrary UUID-shaped strings to burn through any target member's bucket

**Location:** `wishlistService.web.js` lines 185-191

**Issue:** The comment says "Rate-limit per memberId to prevent enumeration of arbitrary IDs." But the key passed to `checkRateLimit` is the caller-supplied `cleanId` (the memberId from the URL). This means an attacker can make 30 requests for `member-victim-uuid` and exhaust the rate-limit bucket for that specific member, blocking legitimate share-link viewers from loading that person's wishlist for the next 60 seconds. The endpoint's anti-enumeration goal is inverted: the attacker enumerates *any* ID they like, each request costs them 1/30th of a bucket but fully burns the *victim's* bucket.

A more robust keying strategy would include the caller's IP or session token as the bucket key (or at minimum combine caller identity with target memberId). Since Wix Velo does not expose request IP in `webMethod`, the practical fix is either: (a) key by a caller-side token/session, or (b) use a global bucket (not per-target) so an attacker cannot selectively exhaust one victim's quota.

**User impact:** A determined attacker can repeatedly DoS the wishlist share view for any specific member by cycling through 30 requests every 60 seconds. Legitimate share-link visitors get blocked with no actionable error message (see issue above).

---

## HIGH: `getWishlistByMemberId` catch block returns `{ success: false, items: [], total: 0 }` — `wixData` query failures are indistinguishable from rate-limit blocks

**Location:** `wishlistService.web.js` lines 204-207

**Issue:** A `wixData.query(...).find({ suppressAuth: true })` failure (e.g., network error, Wix CMS outage) returns the identical shape as a rate-limit block. The cfw share-view cannot distinguish "infrastructure is down" from "too many requests." Both look like an empty, failed response.

**Recommendation:** Add an `error` field distinguishing infrastructure failure from access control failure. This lets the client surface "Something went wrong, please try again" vs "Too many requests" appropriately.

---

## MEDIUM: `getWishlist` unauthenticated path returns `success: false` but `getWishlistByMemberId` unauthenticated-equivalent path returns `success: false` for missing ID — inconsistent `error` field presence

**Location:** `wishlistService.web.js` line 147 vs line 183

`getWishlist` line 147:
```js
if (!member?._id) return { success: false, items: [], total: 0 };
```
No `error` field. Every other auth failure in this file (e.g., `addToWishlist` line 52, `removeFromWishlist` line 112) returns `error: 'Not authenticated.'`. This inconsistency means callers of `getWishlist` cannot reliably check `result.error` to distinguish auth failure from DB failure.

**Recommendation:** Standardize: all `success: false` returns should include an `error` string. Consider enforcing this with a TypeScript return type or JSDoc `@typedef`.

---

## MEDIUM: `updateWishlistStock` iterates serially with `await` inside `for...of` — a single item update failure silently aborts the batch mid-loop

**Location:** `wishlistService.web.js` lines 265-269

```js
for (const item of result.items) {
  if (item.inStock !== !!inStock) {
    await wixData.update('Wishlist', { ...item, inStock: !!inStock });
    updatedCount++;
  }
}
```

If `wixData.update` throws for item N of M, the outer catch block fires, `updatedCount` reflects only items 0..N-1, and the caller receives `{ success: false, updatedCount: N-1 }`. Items N+1..M are never updated. No item-level error is logged. A Wix CMS transient error on one record silently leaves all subsequent records stale.

**Recommendation:** Either use `Promise.allSettled` for parallel updates with per-item error logging, or wrap the inner `await` in its own try/catch that logs the failure and continues the loop:

```js
for (const item of result.items) {
  if (item.inStock !== !!inStock) {
    try {
      await wixData.update('Wishlist', { ...item, inStock: !!inStock });
      updatedCount++;
    } catch (itemErr) {
      logError('wishlistService.updateWishlistStock.item', itemErr);
      // continue — don't abort the entire batch for one item
    }
  }
}
```

**Test gap:** No test covers the partial-failure scenario. A test asserting that `updatedCount` reflects successfully updated items even when one throws would catch a regression here.

---

## MEDIUM: `getWishlistByMemberId` has `Permissions.Anyone` but no test verifies that `suppressAuth: true` is actually passed to the query

**Location:** `wishlistService.web.js` line 197; `tests/wishlistService.test.js` lines 421-496

The `suppressAuth: true` option on the `wixData.query(...).find()` call is load-bearing for this endpoint — without it, unauthenticated reads will fail on collections with member-only read permissions. The test suite seeds data and checks results, but because the mock's `find()` likely ignores the options argument, a regression that accidentally removes `suppressAuth: true` from the production call would pass all tests while silently failing in production.

**Recommendation:** Add a test that captures the options object passed to `find()` (via `__onFind` or similar mock hook) and asserts `suppressAuth === true`.

---

## Summary

| # | Severity | Issue |
|---|----------|-------|
| 1 | CRITICAL | All catch blocks use `console.error` — errors never reach Sentry/monitoring |
| 2 | HIGH | Rate-limited response has no `error` field — indistinguishable from empty wishlist |
| 3 | HIGH | Rate-limit bucket keyed by victim memberId — enables targeted DoS of specific members' share links |
| 4 | HIGH | DB failure and rate-limit block return identical shapes to cfw caller |
| 5 | MEDIUM | `getWishlist` unauthenticated path inconsistently omits `error` field |
| 6 | MEDIUM | Serial `updateWishlistStock` loop aborts silently mid-batch on first item failure |
| 7 | MEDIUM | No test verifies `suppressAuth: true` is passed to the `getWishlistByMemberId` query |

Issues 1, 2, and 3 should be resolved before merge. Issues 4-7 are strongly recommended.

DreadPirateRobertz (commented):
Approve. Squash + admin-merge per cfutons branch protection.

PR #1123, opened 2026-04-26
feat(cf-hc2i): add reengagement_2 and reengagement_3 to email template registry

DreadPirateRobertz:
## Summary

- **Root cause**: `emailAutomation.web.js` queues all 3 win-back steps up-front but `TEMPLATE_REGISTRY` in `emailTemplates.web.js` only had `reengagement_1` — any send for Day 7 or Day 21 steps would fail at template lookup with "Template not found"
- Added `reengagement_2` (Day 7 — "Still looking? Here's a deal") and `reengagement_3` (Day 21 — "Last chance") to `TEMPLATE_REGISTRY` with correct subject lines, preview text, and variable schemas matching `emailAutomation.web.js` queue payload
- Added `reengagement-03-last-chance.html` email template (Day 21 final reminder with urgency bar)
- Updated 2 existing tests asserting `reengagement` length was 1 → now 3; added 8 new targeted tests

## Test plan

- [x] `tests/emailTemplates.test.js` — 111 tests pass (was 103)
- [x] `tests/emailTemplatesDeepen.test.js` — all tests pass
- [x] `reengagement_2`/`reengagement_3` entries returned by `getTemplatesBySequence('reengagement')`
- [x] `validateTemplateVariables` confirms variable schemas correct for both new entries
- [x] `resolveSubjectLine` interpolates `{firstName}` in both new subject lines

### DIFF
--- a/content/newsletters/reengagement-03-last-chance.html
+++ b/content/newsletters/reengagement-03-last-chance.html
@@ -1,0 +1,89 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+<meta charset="utf-8">
+<meta name="viewport" content="width=device-width, initial-scale=1.0">
+<title>Last Chance — Your Offer Expires Soon</title>
+<style>
+  * { margin: 0; padding: 0; box-sizing: border-box; }
+  body { background: #F0F4F8; font-family: Arial, sans-serif; color: #1E3A5F; }
+  .wrapper { max-width: 600px; margin: 0 auto; background: #ffffff; }
+  .header { background: #1E3A5F; padding: 32px 24px; text-align: center; }
+  .header h1 { color: #ffffff; font-size: 28px; font-weight: 700; }
+  .header .tagline { color: #A8CCD8; font-size: 14px; margin-top: 8px; }
+  .urgency-bar { background: #C0392B; padding: 10px 24px; text-align: center; }
+  .urgency-bar p { color: #ffffff; font-size: 14px; font-weight: 700; letter-spacing: 0.5px; }
+  .hero { padding: 48px 32px; text-align: center; }
+  .hero h2 { font-size: 28px; color: #1E3A5F; margin-bottom: 16px; line-height: 1.3; }
+  .hero p { font-size: 16px; color: #3D5A80; line-height: 1.6; }
+  .content { padding: 0 32px 32px; }
+  .content p { font-size: 14px; color: #3D5A80; line-height: 1.6; margin-bottom: 16px; }
+  .offer-box { background: linear-gradient(135deg, #1E3A5F 0%, #3D5A80 100%); border-radius: 12px; padding: 32px 24px; text-align: center; margin: 24px 0; }
+  .offer-box .label { color: #A8CCD8; font-size: 12px; text-transform: uppercase; letter-spacing: 2px; margin-bottom: 8px; }
+  .offer-box h3 { color: #ffffff; font-size: 22px; margin-bottom: 8px; }
+  .offer-box p { color: #A8CCD8; font-size: 14px; margin-bottom: 16px; }
+  .offer-box .code { display: inline-block; background: #4A7D94; color: #ffffff; padding: 8px 24px; border-radius: 4px; font-size: 20px; font-weight: 700; letter-spacing: 3px; }
+  .divider { height: 2px; background: linear-gradient(90deg, #4A7D94, #5B8FA8, #4A7D94); margin: 0 32px; }
+  .cta-block { text-align: center; padding: 32px; }
+  .cta-btn { display: inline-block; background: #4A7D94; color: #ffffff !important; padding: 14px 36px; border-radius: 6px; text-decoration: none; font-weight: 700; font-size: 16px; }
+  .unsub-note { text-align: center; padding: 0 32px 24px; }
+  .unsub-note p { font-size: 13px; color: #3D5A80; font-style: italic; }
+  .footer { background: #1E3A5F; padding: 32px; text-align: center; }
+  .footer p { color: #A8CCD8; font-size: 13px; line-height: 1.6; }
+  .footer a { color: #ffffff; text-decoration: underline; }
+  .footer .address { color: #A8CCD8; font-size: 12px; margin-top: 16px; }
+  @media (max-width: 620px) {
+    .wrapper { width: 100% !important; }
+    .header, .hero, .content, .cta-block { padding-left: 16px !important; padding-right: 16px !important; }
+    .offer-box { margin: 24px 16px !important; }
+  }
+</style>
+</head>
+<body>
+<div class="wrapper">
+
+  <div class="header">
+    <h1>Carolina Futons</h1>
+    <div class="tagline">Quality Furniture from the Blue Ridge Mountains</div>
+  </div>
+
+  <div class="urgency-bar">
+    <p>&#9201; This is your last reminder &mdash; your offer expires in 48 hours</p>
+  </div>
+
+  <div class="hero">
+    <h2>We Hate Goodbyes, {firstName}</h2>
+    <p>Three weeks ago we reached out. You haven't been back, and we understand &mdash; timing matters. But before we stop sending you updates, we wanted to make sure you saw this one last offer.</p>
+  </div>
+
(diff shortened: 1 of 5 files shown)


DreadPirateRobertz:
**code-reviewer** — dispatched by godfrey

Overall this is a clean, targeted fix. A few findings:

1. **Registry IDs are consistent** — `reengagement_2` and `reengagement_3` entries correctly mirror the shape of `reengagement_1` (id, name, sequence, step, subjectLine, previewText, variables, category). No typos or field omissions detected.

2. **delayHours alignment** — `emailAutomation.web.js` lines 91-92 use `delayHours: 168` (Day 7) and `delayHours: 504` (Day 21). 504 / 24 = 21 days exactly. Both values are correct.

3. **`discountAvailable` declared but unused in HTML** — All three registry entries list `'discountAvailable'` in `variables`, but `reengagement-03-last-chance.html` contains no `{discountAvailable}` placeholder. The HTML unconditionally shows "10% Off" with no conditional block. If this variable is intended to gate the offer display (i.e., show a different message when no discount is available), the HTML template is missing that branch. If it's carried only for pipeline consistency (same variable shape across all reengagement steps), a brief code comment would clarify the intent and prevent future confusion.

4. **CTA href is a bare domain** — `https://www.carolinafutons.com` with no UTM parameters on the CTA button. The other reengagement templates (`-01`, `-02`) should be checked for consistency; if they carry UTM params, this one is an outlier that will break attribution.

5. **No logic errors** — The registry keys exactly match the `templateId` values referenced in `SEQUENCES.reengagement.steps`, which is the root cause being fixed. The fix is correct.

DreadPirateRobertz:

Reviewing for silent failures and inadequate error handling:

1. **Missing registry key = silent send failure (the bug being fixed)** — Before this PR, `queueEmail` in `emailAutomation.web.js` (lines 1121-1134) would receive `templateId: 'reengagement_2'` or `'reengagement_3'`, but those keys did not exist in `TEMPLATE_REGISTRY`. The downstream `getTemplate` call would silently return `null`, and the email would fail at send time with no registry-level guard. This PR fixes the root cause, but there is still **no guard in `queueEmail` or the sequence loop** to detect a missing registry entry before enqueuing. If a registry key is ever dropped again, the failure will only surface at send time, not at queue time.

   **Recommendation**: Add a registry existence check in the sequence dispatch loop (or in `queueEmail`) and throw/log explicitly rather than silently queuing an unresolvable templateId.

2. **`validateTemplateVariables` falsy check swallows `discountAvailable: false`** — The validator at line 338 uses `!variables[v] && variables[v] !== 0`. A boolean `false` for `discountAvailable` will be treated as missing (`!false === true` and `false !== 0`). This means a legitimate `discountAvailable: false` payload will fail validation even though the variable was explicitly provided. The correct check is `!(v in variables)` or `variables[v] === undefined`. This is a pre-existing bug, but the new templates depend on this field.

3. **HTML template uses `{discountCode}` without a fallback** — If `discountCode` is an empty string (possible if discount generation fails upstream), the rendered email will show a blank code block with no error or fallback text. Silent cosmetic failure.

4. **No test for `discountAvailable: false` validation case** — See point 2. The new tests only verify the happy path (`discountAvailable: true`). A test with `discountAvailable: false` would reveal the falsy-check bug.

DreadPirateRobertz:
**code-simplifier** — dispatched by godfrey

Reviewing for clarity and maintainability:

1. **Registry entries are well-structured and DRY** — The three `reengagement_*` entries share an identical `variables` array. This is fine at 3 entries, but if this sequence grows or variables diverge, extraction of a shared `REENGAGEMENT_VARIABLES` constant would reduce the maintenance surface. Not a blocker at current scale.

2. **Step numbering comment** — The existing comment on the `reengagement` sequence in `emailAutomation.web.js` reads `// day 0 miss-you, day 7 deal, day 21 last call`. The new registry entry names (`'Re-engagement — Here's a Deal'`, `'Re-engagement — Last Chance'`) match this comment exactly. Good alignment.

3. **Test at line 60 uses `toEqual` with a hardcoded array** — `expect(re.map(t => t.id)).toEqual(['reengagement_1', 'reengagement_2', 'reengagement_3'])`. This test will fail if the registry insertion order changes (e.g., if a new template is inserted between existing ones), because `Object.values` ordering is insertion-order in V8 but not guaranteed by spec across all environments. Using `toEqual(expect.arrayContaining([...]))` plus a separate length check would be more resilient.

4. **HTML file naming is consistent** — `reengagement-03-last-chance.html` follows the `reengagement-0N-slug.html` convention of the existing files. Clean.

5. **No dead code introduced** — The two new registry entries are directly consumed by the existing sequence loop. No orphaned definitions.

DreadPirateRobertz:
**comment-analyzer** — dispatched by godfrey

Reviewing comment accuracy:

1. **Sequence comment in `emailAutomation.web.js` line 88** — `// cf-bpt: multi-step win-back — day 0 miss-you, day 7 deal, day 21 last call` — accurate. `delayHours: 0`, `168` (7×24), and `504` (21×24) match the description.

2. **Step descriptions on lines 90-92** — `'Day 0 — we miss you + exclusive offer (cf-bpt)'`, `'Day 7 — here\'s a deal (cf-bpt)'`, `'Day 21 — last chance (cf-bpt)'` — all accurate and match the new registry entry names.

3. **Registry comment block** — The `// Re-engagement` section header at line 207 covers all three entries. No misleading or stale inline comments introduced.

4. **HTML title vs. subject line discrepancy** — The HTML `<title>` tag reads `Last Chance — Your Offer Expires Soon` while the registry `subjectLine` reads `Last chance, {firstName} — your offer expires soon`. The title is used in browser/preview-in-browser view; the subject line is what the recipient sees in their inbox. The difference (capitalization, no `{firstName}`) is intentional and correct — no accuracy issue, but worth noting.

5. **HTML urgency bar says "expires in 48 hours"** — The registry has no expiry duration field, and the automation does not pass an expiry deadline as a template variable. The "48 hours" figure is hardcoded in the HTML and cannot be made dynamic without adding a variable. This is not a comment inaccuracy per se, but it is a factual claim in the email body that should be verified against the actual discount expiry logic.

6. **JSDoc on `getTemplatesBySequence` line 248** — Lists `reengagement` as a valid sequence type value. This comment is now accurate (previously only `reengagement_1` existed, making the sequence partially broken).

DreadPirateRobertz:

Reviewing test coverage quality:

**What's covered well:**
- Registry completeness: length check + id array match for all 3 reengagement templates (line 57-61)
- `getTemplatesBySequence('reengagement')`: count=3 and step sort order [1,2,3] (line 162-166)
- `resolveSubjectLine` for all three templates with `{firstName}` substitution (lines 310-323)
- `validateTemplateVariables` happy path for both new templates with all required variables (lines 417-437)
- Missing variables case for `reengagement_3` (lines 439-445)

**Coverage gaps:**

1. **`validateTemplateVariables('reengagement_2', {})` missing** — The missing-variables test only covers `reengagement_3`. The gap test for `reengagement_2` is absent. Given these were both missing from the registry, parity is expected.

2. **`discountAvailable: false` not tested** — As noted by silent-failure-hunter, the validator has a falsy-check bug. A test like `validateTemplateVariables('reengagement_2', { firstName: 'A', discountCode: 'X', discountAvailable: false, email: 'a@b.com' })` should return `valid: true` but will currently return `valid: false`. This test gap hides a real defect.

3. **`getTemplate('reengagement_2')` and `getTemplate('reengagement_3')` not tested** — The single-template lookup via `getTemplate` is untested for the new entries. A quick existence + shape check would prevent a future registry key typo from going undetected.

4. **No test for the sequence loop consuming the new templates** — The integration path (the sequence loop in `emailAutomation.web.js` iterating all 3 steps and calling `queueEmail`) is not covered by these tests. There's no test asserting that `queueEmail` is called 3 times with the correct `templateId` values when a reengagement sequence is triggered. This is the highest-risk gap since it's exactly the path that was broken.

5. **Test at line 60 is order-sensitive** — See code-simplifier note. `Object.values` insertion-order reliance could cause a brittle failure.

**Summary**: Happy-path coverage is solid. The critical gap is the missing integration test for the 3-step sequence loop and the `discountAvailable: false` validator case.

PR #1134, opened 2026-05-04
fix(cf-nbu4): gate promo lightbox with wix-storage-frontend session scope

DreadPirateRobertz:
## Summary

- **Root cause**: `masterPage.js` used browser `sessionStorage` to gate the promotional lightbox. Wix re-initializes the master page context on each client-side page navigation, resetting `sessionStorage` and causing the lightbox to fire every time.
- **Fix**: Replace raw `sessionStorage` with `wix-storage-frontend`'s `session` scope, which Wix persists across client-side page navigations within the same browser tab — the same pattern used by `MembershipPrompt.js` and `recentlyViewed.js`.
- Removes the `typeof sessionStorage !== 'undefined'` guard (no longer needed; `wix-storage-frontend` handles availability internally).

## Test Plan

- [x] New test: lightbox suppressed when `wixSession.getItem(dismissKey)` returns truthy (prior dismissal)
- [x] New test: dismiss key written to `wixSession` on overlay click
- [x] Existing 8 promo lightbox tests continue to pass (mock wires through `globalThis.sessionStorage`)

### DIFF
--- a/src/pages/masterPage.js
+++ b/src/pages/masterPage.js
@@ -505,7 +505,9 @@ async function initPromoLightbox() {
     $w('#promoOverlay').show('fade', { duration: 300 });
     _promoDialog.open();
   } catch (e) {
-    // Promo lightbox is non-critical — never block the page
+    // Promo lightbox is non-critical — never block the page, but log so
+    // Wix monitoring captures storage/API failures that silently prevent display.
+    console.error('[masterPage] initPromoLightbox failed:', e?.message);
   }
 }
 
--- a/src/pages/masterPage.js
+++ b/src/pages/masterPage.js
@@ -14,6 +14,7 @@ import { trackEvent } from 'public/engagementTracker';
 import { fireCustomEvent, initScrollDepthTracking } from 'public/ga4Tracking';
 import { colors, typography, spacing } from 'public/designTokens.js';
 import { captureInstallPrompt, canShowInstallPrompt, showInstallPrompt, isInstalledPWA } from 'public/pwaHelpers';
+import { session as wixSession } from 'wix-storage-frontend';
 import { initAppDownloadBanner } from 'public/AppDownloadBanner';
 import { reportMetrics } from 'backend/coreWebVitals.web';
 import { initFooter, initMountainDividerWithSkyWiring } from 'public/FooterSection';
@@ -475,15 +476,11 @@ async function initPromoLightbox() {
     const promo = await getActivePromotion();
     if (!promo) return;
 
-    // Check if user already dismissed this promotion in this session
+    // Check if user already dismissed this promotion in this session.
+    // Uses wix-storage-frontend session scope so the gate persists across
+    // Wix client-side page navigations (browser sessionStorage resets per nav).
     const dismissKey = `promo_dismissed_${promo._id}`;
-    if (typeof sessionStorage !== 'undefined') {
-      try {
-        if (sessionStorage.getItem(dismissKey)) return;
-      } catch (e) {
-        // sessionStorage may not be available
-      }
-    }
+    if (wixSession.getItem(dismissKey)) return;
 
     populatePromoLightbox(promo);
     initPromoCountdown(promo.endDate);
@@ -605,11 +602,7 @@ function dismissLightbox(dismissKey) {
     $w('#promoOverlay').hide('fade', { duration: 200 });
   } catch (e) {}
 
-  if (typeof sessionStorage !== 'undefined') {
-    try {
-      sessionStorage.setItem(dismissKey, '1');
-    } catch (e) {}
-  }
+  wixSession.setItem(dismissKey, '1');
 }
 
 function initPromoDismiss(promoId, dismissKey) {
--- a/src/pages/masterPage.js
+++ b/src/pages/masterPage.js
@@ -602,7 +602,7 @@ function dismissLightbox(dismissKey) {
     $w('#promoOverlay').hide('fade', { duration: 200 });
   } catch (e) {}
 
-  wixSession.setItem(dismissKey, '1');
+  try { wixSession.setItem(dismissKey, '1'); } catch (e) {}
 }
 
 function initPromoDismiss(promoId, dismissKey) {
--- a/tests/importBudget.test.js
+++ b/tests/importBudget.test.js
@@ -24,7 +24,7 @@ const KNOWN_OVERBUDGET = {
   'Category Page.js': 27,
   'Home.js': 21, // cf-nqq: +1 initChallengeOfTheWeekWidget
   'Product Page.js': 29, // CF-7byz: +1 getProductVideos; CF-06xu: +1 getProductStructuredData; hq-kgno: +1 wix-data; CF-wzv8: +1 subscribeAndSave; CF-qgg0: +1 renderSimplePrice
-  'masterPage.js': 22, // CF-qgg0: +1 formatCardPrice/renderSimplePrice from productCardHelpers; CF-e2ib: +1 initAppDownloadBanner
+  'masterPage.js': 23, // CF-qgg0: +1 formatCardPrice/renderSimplePrice from productCardHelpers; CF-e2ib: +1 initAppDownloadBanner; cf-nbu4: +1 wix-storage-frontend (promo lightbox session gate)
   'Thank You Page.js': 23, // CF-2zr3: +1 getSoftPromptConfig; CF-y7lp: +2 getWhiteGloveSlots + wixLocationFrontend
 };
 
--- a/tests/masterPage.test.js
+++ b/tests/masterPage.test.js
@@ -105,6 +105,7 @@ import { reportMetrics } from 'backend/coreWebVitals.web';
 import { isInstalledPWA, canShowInstallPrompt } from 'public/pwaHelpers';
 import { submitContactForm } from 'backend/contactSubmissions.web';
 import { initConsentGate, fireTrackedTikTokEvent } from 'public/pixelConsentService';
+import { session as wixSession } from 'wix-storage-frontend';
 const { mockIsMobile, mockGetViewport } = vi.hoisted(() => ({
   mockIsMobile: vi.fn(() => false),
   mockGetViewport: vi.fn(() => 'desktop'),
@@ -1637,6 +1638,8 @@ describe('masterPage.js', () => {
   // ── Promotional Lightbox ──────────────────────────────────────────
 
   describe('promotional lightbox', () => {
+    afterEach(() => { wixSession.clear(); });
+
     it('populates promo lightbox when active promotion exists', async () => {
       getActivePromotion.mockResolvedValueOnce({
         _id: 'promo-1',
@@ -1728,14 +1731,13 @@ describe('masterPage.js', () => {
         products: [],
       });
       // Simulate prior dismissal via wix-storage-frontend session storage
-      globalThis.sessionStorage.setItem('promo_dismissed_promo-gate-1', '1');
+      wixSession.setItem('promo_dismissed_promo-gate-1', '1');
       vi.useFakeTimers();
       elements.clear();
       await onReadyHandler();
       vi.advanceTimersByTime(3000);
       await vi.advanceTimersByTimeAsync(100);
       expect(getEl('#promoOverlay').show).not.toHaveBeenCalled();
-      globalThis.sessionStorage.clear();
       vi.useRealTimers();
     });
 
@@ -1759,8 +1761,7 @@ describe('masterPage.js', () => {
       const overlayHandler = getEl('#promoOverlay').onClick.mock.calls[0]?.[0];
       if (overlayHandler) overlayHandler();
       await vi.advanceTimersByTimeAsync(50);
-      expect(globalThis.sessionStorage.getItem('promo_dismissed_promo-gate-2')).toBe('1');
-      globalThis.sessionStorage.clear();
+      expect(wixSession.getItem('promo_dismissed_promo-gate-2')).toBe('1');
       vi.useRealTimers();
     });
 
--- a/tests/masterPage.test.js
+++ b/tests/masterPage.test.js
@@ -1718,6 +1718,52 @@ describe('masterPage.js', () => {
       vi.useRealTimers();
     });
 
+    it('does not show lightbox when promo already dismissed this session', async () => {
+      getActivePromotion.mockResolvedValueOnce({
+        _id: 'promo-gate-1',
+        title: 'Spring Sale',
+        subtitle: 'Save 20%',
+        discountCode: 'SPRING20',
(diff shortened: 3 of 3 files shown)


DreadPirateRobertz (commented):

## Summary

Reviewed the swap from browser `sessionStorage` to `wix-storage-frontend`'s `session` scope for the promo lightbox dismiss gate. The fix correctly addresses the documented bug: Wix re-initializes the master page on each client-side navigation, which clears browser `sessionStorage` and causes the lightbox to refire. Switching to `wix-storage-frontend` `session` matches the established pattern in `MembershipPrompt.js`, `recentlyViewed.js`, `ugcVoting.js`, and `fitScoreEngine.js`.

The two new tests cover the prior-dismissal gate and the dismiss-write path. The existing 8 promo lightbox tests continue to validate behavior because the mock at `tests/__mocks__/wix-storage-frontend.js` proxies through `globalThis.sessionStorage`.

No high-priority blocking issues found. One important issue worth addressing before merge:

## Important

### `dismissLightbox` no longer guards `setItem` against throws (confidence 82)

**File:** `src/pages/masterPage.js:599-606`

```js
function dismissLightbox(dismissKey) {
  try {
    $w('#promoLightbox').hide('fade', { duration: 200 });
    $w('#promoOverlay').hide('fade', { duration: 200 });
  } catch (e) {}

  wixSession.setItem(dismissKey, '1');  // not inside try/catch
}
```

Previously the bare-`sessionStorage` write was wrapped in `try { ... } catch (e) {}`. The new call is bare. `wix-storage-frontend` session storage can throw under quota-exceeded, storage-disabled (private browsing on some platforms), or transient init conditions. Because `dismissLightbox` is invoked from inside `_promoDialog.close()`'s `onClose` callback, an exception here propagates back into the dialog close path and may leave focus restoration / cleanup in a half-applied state.

The other consumers of `wixSession` in this codebase (e.g. `recentlyViewed.js`, `MembershipPrompt.js` outer `try`) all keep their writes inside a try/catch. Suggest:

```js
function dismissLightbox(dismissKey) {
  try {
    $w('#promoLightbox').hide('fade', { duration: 200 });
    $w('#promoOverlay').hide('fade', { duration: 200 });
  } catch (e) {}

  try {
    wixSession.setItem(dismissKey, '1');
  } catch (e) {
    // storage may be unavailable — non-critical
  }
}
```

Same defensive guard would also be appropriate around the `getItem` call at line 483, though that one is already safely inside `initPromoLightbox`'s outer `try { ... } catch (e) { /* non-critical */ }`, so the gate read is fine as-is.

## Notes (not blocking)

- The new test `'sets session dismiss key when lightbox is dismissed via close button'` actually exercises the overlay click path (`getEl('#promoOverlay').onClick.mock.calls[0]?.[0]`), not the close button. The test name is slightly misleading but functionally correct. Consider renaming to `'sets session dismiss key when overlay is clicked'` for clarity.
- Mock at `tests/__mocks__/wix-storage-frontend.js` already exists and proxies through `globalThis.sessionStorage`, so the test approach of seeding/asserting via `globalThis.sessionStorage` is valid.

Otherwise the change is small, focused, and matches existing conventions. LGTM with the `setItem` try/catch suggestion.

DreadPirateRobertz (commented):

---

## Error Handling Audit — PR #1134

The core fix (swapping raw `sessionStorage` for `wix-storage-frontend` session scope) is the right call and the rationale is sound. The mock wiring in `tests/__mocks__/wix-storage-frontend.js` is consistent with the peer module pattern used by `MembershipPrompt.js` and `recentlyViewed.js`. That said, I found three issues that need attention before this merges.

---

### CRITICAL — `initPromoLightbox` outer catch silently swallows all errors, including new wixSession failure

**Location:** `src/pages/masterPage.js` lines 507–509

```js
} catch (e) {
  // Promo lightbox is non-critical — never block the page
}
```

`wixSession.getItem(dismissKey)` (line 483) now runs **inside** this catch block. The old `sessionStorage` calls were also inside individual try/catch blocks with explicit suppression — so the swallow was scoped. Now that `wixSession.getItem` runs naked (no inner try/catch), if `wix-storage-frontend` throws at runtime (quota exceeded, Wix SSR context where the module is not yet hydrated, permission error, or any future regression in the Wix API), the outer catch at line 507 will silently absorb it and the lightbox will silently never appear again — indistinguishable from the gate working correctly.

The same applies to `wixSession.setItem(dismissKey, '1')` at line 605. It sits **outside** any try/catch, so if it throws (e.g. Wix storage quota, SSR context, network isolation on mobile), the exception propagates uncaught from `dismissLightbox()`. If `dismissLightbox()` is called from inside the `_promoDialog.onClose` callback (which is inside the outer try), the throw will be caught by that same silent outer catch — meaning the hide animation at lines 601–602 may have completed but the dismiss key was never written. On the next page nav the lightbox shows again, with no log to indicate why.

**Recommendation:** At minimum, log the error before returning so future debugging is possible. Since this project uses `logError` for Sentry tracking:

```js
} catch (e) {
  // Promo lightbox is non-critical — never block page load.
  // Log so Sentry can detect wixSession regressions or API breakage.
  logError(e, errorIds.PROMO_LIGHTBOX_INIT_FAILED, { dismissKey });
}
```

If `errorIds.PROMO_LIGHTBOX_INIT_FAILED` does not yet exist, a `console.warn` is still vastly better than the current comment-only catch. The lightbox being "non-critical" justifies not rethrowing — it does not justify not logging.

---

### HIGH — Test asserts against `globalThis.sessionStorage` but production code now calls `wixSession` — the new tests are testing the mock, not the contract

**Location:** `tests/masterPage.test.js` lines 1731, 1738, 1762, 1763

```js
// Test: "does not show lightbox when promo already dismissed this session"
globalThis.sessionStorage.setItem('promo_dismissed_promo-gate-1', '1');
// ...
globalThis.sessionStorage.clear();

// Test: "sets session dismiss key when lightbox is dismissed via close button"
expect(globalThis.sessionStorage.getItem('promo_dismissed_promo-gate-2')).toBe('1');
globalThis.sessionStorage.clear();
```

The tests work only because `tests/__mocks__/wix-storage-frontend.js` proxies through `globalThis.sessionStorage`. They are asserting on the underlying test-only proxy, not on the actual `wixSession` module interface. If someone refactors the mock to use an in-memory store (which would be a legitimate simplification), both new tests silently break — the assertion `globalThis.sessionStorage.getItem(...)` returns `null`, the `expect` fails, and the reviewer has no idea the breakage is in the mock proxy rather than the feature.

The tests should assert through the mock itself:

```js
import { session as wixSession } from 'wix-storage-frontend';

// setup:
wixSession.setItem('promo_dismissed_promo-gate-1', '1');

// teardown:
wixSession.clear();

// assertion:
expect(wixSession.getItem('promo_dismissed_promo-gate-2')).toBe('1');
```

This also makes the test intent clearer: "the production code writes through the wix-storage-frontend session API", not "the production code ends up touching globalThis.sessionStorage".

---

### HIGH — `dismissLightbox` calls `wixSession.setItem` outside any error boundary — uncaught throw silently skips hide animation

**Location:** `src/pages/masterPage.js` lines 599–606

```js
function dismissLightbox(dismissKey) {
  try {
    $w('#promoLightbox').hide('fade', { duration: 200 });
    $w('#promoOverlay').hide('fade', { duration: 200 });
  } catch (e) {}          // <-- hide errors silently suppressed

  wixSession.setItem(dismissKey, '1');  // <-- outside any guard
}
```

The original code had `sessionStorage.setItem` inside its own try/catch (lines removed in the diff). That guard is gone. If `wixSession.setItem` throws synchronously (storage quota, Wix API unavailability, malformed key), the exception exits `dismissLightbox` uncaught. Depending on the call site:

- Called from `_promoDialog.onClose` callback (inside the outer `initPromoLightbox` try) → caught silently, no log.
- Called from `initPromoDismiss` → `dismissHandler` → not inside any try/catch, propagates to Wix's event system → silently swallowed by Wix, no user feedback, no log.

The user has clicked to dismiss, the overlay may or may not have animated away (race with the hide), and the key was never written, so the lightbox reappears on next navigation. There is zero signal in the logs.

**Recommendation:**

```js
function dismissLightbox(dismissKey) {
  try {
    $w('#promoLightbox').hide('fade', { duration: 200 });
    $w('#promoOverlay').hide('fade', { duration: 200 });
  } catch (e) {}

  try {
    wixSession.setItem(dismissKey, '1');
  } catch (e) {
    console.warn('[masterPage] dismissLightbox: failed to write dismiss key, lightbox may re-appear:', e?.message);
  }
}
```

---

### MEDIUM — `initPromoLightbox` outer catch comment does not match severity — should be explicit about what is and is not caught

**Location:** `src/pages/masterPage.js` lines 507–509

The comment "Promo lightbox is non-critical — never block the page" is accurate in intent but misleading in scope. This catch swallows: `getActivePromotion` failures, `populatePromoLightbox` throws, `setupAccessibleDialog` throws, `wixSession.getItem` throws, `$w` selector failures, and every sub-init. A future developer adding logic inside `initPromoLightbox` will not notice this catch silently suppresses their new code too. The comment should enumerate what category of errors is expected, or the outer try should only wrap the non-critical UI calls while letting `getActivePromotion` failures be logged:

```js
let promo;
try {
  promo = await getActivePromotion();
} catch (e) {
  console.warn('[masterPage] getActivePromotion failed, skipping promo lightbox:', e?.message);
  return;
}
if (!promo) return;
// ... rest of init inside the existing broad catch is acceptable for UI ops
```

---

### Summary

| # | Location | Severity | Issue |
|---|---|---|---|
| 1 | `masterPage.js:507-509` | CRITICAL | Outer catch silently swallows wixSession failures — no log, no Sentry signal |
| 2 | `masterPage.test.js:1731,1762` | HIGH | Tests assert on `globalThis.sessionStorage` proxy instead of `wixSession` interface — mock-coupled |
| 3 | `masterPage.js:605` | HIGH | `wixSession.setItem` outside any guard — uncaught throw silently prevents dismiss key write |
| 4 | `masterPage.js:507-509` | MEDIUM | Catch comment understates scope — hides future developer intent |

Items 1 and 3 are pre-existing silent failure patterns that were present before this PR, but this PR's changes land squarely inside them. Given that this is the exact path that gates the promo lightbox, it is worth fixing before merge rather than accruing as tech debt. Item 2 is new and should be corrected in the test file before this lands.

DreadPirateRobertz (commented):

## Test Coverage Analysis: fix(cf-nbu4) promo lightbox wix-storage-frontend gate

### Summary

The two new tests provide correct behavioral coverage for the core fix. The mock wiring is sound and the tests validate observable outcomes rather than implementation details. Overall coverage quality is good, but there are two meaningful gaps worth addressing.

---

### How the mock wires through

`wix-storage-frontend` is resolved via the auto-discovered alias in `vitest.config.js` to `tests/__mocks__/wix-storage-frontend.js`, which delegates `session.getItem/setItem` directly to `globalThis.sessionStorage`. This means both new tests exercise the real code path without any per-test `vi.mock()` declaration — correct and consistent with the project pattern.

---

### Critical Gaps

**1. Dismiss write goes through `_promoDialog.close()` → `onClose` → `dismissLightbox`, not directly** (Criticality: 7/8)

The "sets session dismiss key" test fires `#promoOverlay`'s `onClick` handler. In production, with `_promoDialog` set (which it will be, since `setupAccessibleDialog` is loaded via `vi.importActual`), the overlay handler calls `_promoDialog.close()`, which calls `onClose()`, which calls `dismissLightbox(dismissKey)`, which calls `wixSession.setItem`. The test validates the correct end state, so this path IS covered — but only via the overlay click. The `#promoDismiss` button path (which goes through `makeClickable`) and the Escape-key path (wired by `setupAccessibleDialog` to `document.addEventListener('keydown', escHandler)`) both reach `dismissLightbox` too and write the session key. If either of those dismiss vectors broke, the test suite would not catch it.

A test exercising one of the non-overlay dismiss vectors (e.g., simulating the `#promoDismiss` click or a `keydown` Escape event) would guard against a regression where the key is written for overlay clicks but silently skipped for keyboard-initiated dismissals. Criticality: 6/10 — this is a real behavioral gap but the fix itself only touched `wixSession.setItem` and `.getItem`, not the dismiss routing logic.

**2. No test for `wixSession.getItem` returning a falsy-but-non-null value** (Criticality: 5/10)

The gate is `if (wixSession.getItem(dismissKey)) return`. The existing test seeds `'1'` (truthy string). The `wix-storage-frontend` mock coerces all values to `String(value)` on `setItem`. If a future caller inadvertently stores `'0'` or `''`, the gate would silently fail to suppress. This is a low-probability edge case but it would be a silent failure — the lightbox would reappear unexpectedly. A test seeding an empty string and asserting the lightbox still shows would document the contract clearly.

---

### Important Improvements

**3. Suppress-if-dismissed test uses a weak negative assertion** (Criticality: 5/10)

```js
expect(getEl('#promoOverlay').show).not.toHaveBeenCalled();
```

This is correct as far as it goes, but it would also pass if `getActivePromotion` threw an error and the entire `initPromoLightbox` was swallowed by the outer `catch`. Consider adding a spy or small assertion confirming `getActivePromotion` was called (it resolved successfully) before asserting suppression — otherwise a misconfigured `getActivePromotion` mock returning `null` would produce a false-green test. The existing "does not show promo lightbox when no active promotion" test has the same shape, so this is a pre-existing pattern — flagging for awareness rather than blocking.

**4. `globalThis.sessionStorage.clear()` cleanup is inline rather than in `afterEach`** (Criticality: 3/10)

Both new tests call `globalThis.sessionStorage.clear()` at the end of the test body. If the assertion throws before reaching `clear()`, storage leaks into the next test. The existing promo tests that need a clean slate rely on this — the "populates promo lightbox" test is immediately adjacent and could be polluted by a prior test failure. Using an `afterEach` (or the existing `beforeEach`/`afterEach` block if the describe scope has one) is safer. This is a low-criticality pattern issue, not a current bug.

---

### Positive Observations

- The mock delegation to `globalThis.sessionStorage` is clean and consistent with other storage mocks in the codebase. No need for per-test `vi.mock` boilerplate.
- Both tests correctly use `vi.useFakeTimers()` with the 3-second `setTimeout` advance pattern established by the existing promo tests — no timer leaks introduced.
- Asserting `#promoOverlay`.show was not called (rather than checking a storage flag in isolation) tests the user-visible behavior, not the internal branching. This is the right level of abstraction for the suppress test.
- Removing the `typeof sessionStorage !== 'undefined'` guard and the nested `try/catch` is a net simplification that reduces dead branches — fewer untestable code paths.
- The dismiss-write test uses `vi.waitFor` to confirm the lightbox was shown before attempting to dismiss, which correctly sequences async state rather than assuming timing.

---

### Verdict

The two new tests cover the advertised scenarios correctly and follow project conventions. The gaps identified (alternate dismiss vectors, falsy-value gate contract, cleanup robustness) are improvements rather than blockers. The fix is well-scoped and the test quality matches the existing suite standard.

PR #1138, opened 2026-05-05
fix(cf-foo0): preserve sizeOfInterest in contactSubmissions HTTP wrapper

DreadPirateRobertz:
## Summary

Per @melania review on the Velo PR (DreadPirateRobertz/carolina-futons-stage3-velo#15): the cfw `ContactRequest` carries an optional `sizeOfInterest` (twin/full/queen/king) from the contact form's size radio, and `post_contactSubmissions` was silently dropping it before calling `sendEmail`. Store needs to see size in their inbox so they can respond meaningfully.

**Fix:** when `sizeOfInterest` is a non-empty string, prepend `[Size: <value>] ` to the subject. `sendEmail` has no first-class size field, but the subject flows through to both the triggered email body (Wix Triggered Email variable) and the ContactSubmissions CMS row.

## Tests

`tests/httpFunctionsCoverage.test.js` — added 2 cases in the `post_contactSubmissions` describe block:
- `prepends [Size: <sizeOfInterest>] to the subject so the store sees it` — asserts `crmEmailLog()[0].options.variables.subject === '[Size: queen] Frame question'`
- `omits the size prefix when sizeOfInterest is absent` — backstops the conditional

All 10 `post_contactSubmissions` tests pass locally (`vitest run -t 'post_contactSubmissions'`).

## Companion PR

The Velo deploy repo (`carolina-futons-stage3-velo`) has the same one-liner amended onto its open PR #15. They must land together for the fix to take effect — the cfutons monorepo edit alone won't reach production until the velo repo publishes via Wix CLI.

## Out of scope (per melania)

- (2) phone-length cap mismatch (cfw uncapped vs backend 20 char) — file as cfw-* follow-up bead.
- (3) standalone wrapper test (this PR adds 2 in the existing coverage file; if a separate file is preferred, follow-up).

Closes follow-up review on cf-foo0.

### DIFF
--- a/src/backend/http-functions.js
+++ b/src/backend/http-functions.js
@@ -2624,6 +2624,19 @@ export async function get_weeklyBlogDigestCron(request) {
 // CORS allowlist (see backend/utils/cors): prod Vercel domain,
 // project-scoped Vercel domain, branch preview URLs, localhost dev.
 
+// Whitelist of accepted bed sizes — mirrors cfw `BedSize` type
+// (src/lib/contact/contact-schema.ts). The wrapper rejects any other value
+// silently (treats it as if `sizeOfInterest` were absent) so a malicious
+// client can't smuggle arbitrary text into the subject prefix.
+const SIZE_OF_INTEREST_WHITELIST = new Set(['twin', 'full', 'queen', 'king']);
+
+// sendEmail caps the subject at 300 chars via validateSchema. The size prefix
+// (`[Size: queen] ` = 14 chars max) plus a max-length subject would exceed
+// that cap and fail validation — the user would see a generic 400. Truncate
+// the combined subject so the prefix is preserved end-to-end and only the
+// trailing user copy is clipped if necessary.
+const SUBJECT_MAX_LEN = 300;
+
 /**
  * @function post_contactSubmissions
  * @route POST /_functions/contactSubmissions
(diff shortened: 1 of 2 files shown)


DreadPirateRobertz:
## 5-Agent Review — FIXES REQUIRED

### code-reviewer (HIGH)
- **300-char subject cap interaction.** `sendEmail` (emailService.web.js:124) validates `subject` with maxLength 300. The prefix `[Size: <value>] ` adds 14–17 chars. A subject at/near the cfw cap will now fail validation with 400 where it previously passed. JSDoc still advertises ≤300. Either cap `rawSubject` to `300 - prefixLen` pre-concat, or document the new effective cap.
- Whitespace-only `rawSubject` diverges between prefixed/unprefixed branches (`.trim()` only on prefixed path). Minor.

### silent-failure-hunter (HIGH)
- **Other passthrough fields.** `sendEmail` destructures only `{name, email, phone, subject, message}`. If cfw's `ContactRequest` carries `productOfInterest`, `preferredContactMethod`, `urgency`, etc. — they silently drop. The PR fixes one silent-drop but adds no guard against the next. Recommend either a `console.warn` for unknown body keys or an audit confirming cfw schema.
- Non-string `sizeOfInterest` (object/number) silently dropped — given the bug being fixed *is* a silent drop, a `console.warn` here is on-theme.

### code-simplifier
- No HIGH-priority findings.

### comment-analyzer (HIGH)
- **Stale JSDoc.** Function header at L2627-2645 documents the 5 original params but not `sizeOfInterest`, even though the body now reads it. New callers won't know the field is supported. Add a `@param {string} [request.body.json.sizeOfInterest]` line.

### pr-test-analyzer (HIGH)
- **Missing CMS-row assertion.** PR claims "flows through to both the triggered email body AND the ContactSubmissions CMS row." Tests only assert `crmEmailLog()[0].options.variables.subject`. No test confirms the CMS-row subject column receives the prefix. Either add a CMS-row assertion or document why existing coverage is equivalent.
- **Missing 300-char boundary test.** No test covers a subject at/near 300 chars combined with `sizeOfInterest`. Add a boundary test pinning expected behavior.
- **Missing non-string `sizeOfInterest` test** (number, null, whitespace-only) — small but locks down the trim/typeof guard.

---

**Verdict: FIXES REQUIRED.** Two runtime concerns worth addressing pre-merge: subject cap interaction + stale JSDoc + decision on unknown-field guard. Test gaps could land as a follow-up bead if you accept the runtime fixes alone.

PR #1140, opened 2026-05-05
docs(cf-icww): email touchpoint audit — static review of welcome + touch emails

DreadPirateRobertz:
## Summary
Static audit of every Velo email touchpoint (`triggeredEmails.emailContact`) across `emailService.web.js`, `emailAutomation.web.js`, `newsletterService.web.js`, `emailQueueService.web.js`, `http-functions.js` cron endpoints, and `events.js`. Report at `docs/email-audit-2026-05-04.md`.

### Findings (P0/P1)
- **F3 (P0)** `order_confirmation` never sends — `events.js#wixEcom_onOrderCreated` does not call `sendOrderConfirmation`; the call lives in a dead duplicate handler in `emailAutomation.web.js`.
- **F4 (P0)** shipping notifications never fire — `wixEcom_onFulfillmentCreated` defined only in `emailAutomation.web.js`. Wix auto-registers handlers from `events.js` only.
- **F5 (P0)** delivery confirmation + Day-14 review reward + NPS survey never fire — `wixEcom_onOrderDelivered` defined only in `emailAutomation.web.js`.
- **F1 (P1)** welcome series for exit-intent + member-self-trigger fails — `recipientContactId` enqueued as `''` and `sendQueuedEmail` throws.
- **F2 (P1)** `subscribeToNewsletter` does not queue welcome despite docs/comments saying it does.
- F6/F7 (P2) — no contact-form auto-reply; swatch-confirmation skipped for new visitors.

### Verification gaps (requires STAGING_SITE)
Static analysis cannot prove HTML render, literal `{{var}}` left-overs, CTA URL correctness, or send-window throttling. A per-row staging test plan is included at the bottom of the report.

## Test plan
- [ ] Stilgar (or owner) walks through the 8-step staging checklist in the report to confirm F1/F3/F4/F5/F6/F7 reproduce and ✅ rows actually render.
- [ ] Mayor/melania prioritise fix beads in the recommended order at the bottom of the report.

### DIFF
--- a/docs/decisions/cf-3l0d-subscribeToNewsletter-decision.md
+++ b/docs/decisions/cf-3l0d-subscribeToNewsletter-decision.md
@@ -1,0 +1,83 @@
+# cf-3l0d — `subscribeToNewsletter` welcome-flow decision
+
+**Status:** awaiting Stilgar/Mayor pick
+**Owner:** rennala
+**Filed:** 2026-05-05
+**Refs:** cf-icww F2, cf-xdji (resolveContactId helper)
+
+## Problem
+
+The styleQuiz.web.js comment claims:
+
+> // subscribeToNewsletter deduplicates silently and triggers the welcome flow.
+
+But `subscribeToNewsletter` does NOT queue any `welcome_series_*` row. It only:
+1. Inserts a `NewsletterSubscribers` row
+2. Fires-and-forgets `_syncToESPInternal(...)` to Klaviyo
+
+The welcome series is queued only by `captureExitIntentEmail`, which the exit-intent UI calls *separately* alongside `subscribeToNewsletter`.
+
+## Caller inventory
+
+### Bucket 1 — explicitly composes both calls (welcome IS queued)
+| Caller | Pairs `subscribeToNewsletter` with `captureExitIntentEmail`? |
+|---|---|
+| `src/public/exitIntentCapture.js#submitExitCapture` | yes (lines 209–222) |
+
+### Bucket 2 — calls `subscribeToNewsletter` only (welcome NOT queued, contrary to comment)
+| Caller | Source label |
+|---|---|
+| `src/public/exitIntentCapture.js#submitExitIntentEmail` | `exit_intent_popup` |
+| `src/public/FooterSection.js` | `footer_newsletter` |
+| `src/pages/Blog.js` | `blog_newsletter` |
+| `src/pages/Blog Post.js` | `blog_post` |
+| `src/pages/Home.js` | (default — no source) |
+| `src/pages/masterPage.js` | `exit_intent_popup` |
+| `src/pages/Newsletter.js` | `newsletter_page` |
+| `src/pages/Thank You Page.js` | `thank_you_page` |
+| `src/backend/styleQuiz.web.js#captureQuizLead` | `style_quiz` |
+| `src/backend/http-functions.js#post_mailingListSignups` | webhook source |
+
+**9 of 10 callers expect welcome behaviour but don't get it.** Only `submitExitCapture` is wired correctly.
+
+## Options
+
+### Option A — rename to honest scope, callers compose explicitly
+- Rename `subscribeToNewsletter` → `subscribeToMailingList` (or similar)
+- Update the styleQuiz comment to remove the false claim
+- Each Bucket-2 caller decides: do they want welcome? if yes, they call `triggerWelcomeSeries(email, '')` (post-cf-xdji) themselves
+- **Pro:** API name truthfully reflects what the function does; explicit composition is auditable
+- **Con:** 10 caller sites to fix; future callers will probably forget the second call again ("rename" doesn't prevent the same mistake from recurring); historical name strongly implies welcome behaviour
+
+### Option B — auto-trigger welcome inside `subscribeToNewsletter` (recommended)
+- After cf-xdji lands, `subscribeToNewsletter` itself calls the welcome trigger (with the `resolveContactId` helper bridging the anonymous-capture → CRM contact gap)
+- `submitExitCapture` drops its explicit `captureExitIntentEmail` call (now a duplicate; the EmailQueue dedup guard would catch it anyway, but better to remove the redundant code)
+- **Pro:** one place to own welcome behaviour; Bucket-2 callers' implicit assumption becomes correct; matches the historical name
+- **Con:** depends on cf-xdji landing first; couples `subscribeToNewsletter` (Anyone permission) to `triggerWelcomeSeries` (currently SiteMember) — needs a privileged internal helper or a permission relax with explicit safeguards (rate limit + dedup are already there)
+
+### Option C — keep `subscribeToNewsletter` thin, add wrapper `subscribeAndWelcome`
+- Hybrid: leave `subscribeToNewsletter` semantics narrow (CMS + ESP), introduce `subscribeAndWelcome(email, opts)` that does subscribe + welcome queue
+- Bucket-2 callers switch to `subscribeAndWelcome`
+- **Pro:** preserves narrow primitive while making the common case ergonomic
+- **Con:** still requires updating all 9 caller sites; two near-duplicate webMethods to maintain
+
+## Recommendation: **Option B**
+
+- 9/10 callers' implicit assumption becomes correct without per-call changes
+- Single source of truth for welcome behaviour
+- Existing dedup guard (`EmailQueue` query keyed on `recipientEmail + sequenceType + sequenceStep`) makes redundant calls safe
+- Couples cleanly to cf-xdji, which already needs to solve the anonymous-contactId problem for F1+F7
+
+**Sequencing:** wait for cf-xdji to merge → import `resolveContactId` in `newsletterService.web.js` → in `subscribeToNewsletter` success branch, after the CMS insert, resolve contactId then call `triggerWelcomeSeries(cleanEmail, '')` (or call `enqueueEmail` directly with the resolved contactId for each welcome step). Update styleQuiz.web.js comment to reflect reality. Remove the now-redundant `captureExitIntentEmail` call from `submitExitCapture`.
+
+## Risk if we wait too long
+
+`subscribeToNewsletter` continues to look like it triggers welcome but doesn't, so any new caller (e.g. a fresh signup form on a new page) silently joins Bucket 2. Cost grows linearly with new callers.
+
+## Asks for Stilgar / Mayor
+
+1. **Pick A / B / C.** Default = B unless you object.
+2. If B: do we want `triggerWelcomeSeries` permission relaxed to Anyone, or do we add an internal-only `_triggerWelcomeSeriesInternal` helper that is callable from `subscribeToNewsletter`'s `Anyone` context? (I recommend the internal helper — keeps the public webMethod gated to SiteMember.)
+3. Confirm OK to remove the redundant `captureExitIntentEmail` call from `submitExitCapture` once welcome is auto-triggered.
+
+Reply on cf-3l0d or `gt mail` with pick + answers; I'll implement once cf-xdji is in.
--- a/docs/email-audit-2026-05-04.md
+++ b/docs/email-audit-2026-05-04.md
@@ -1,0 +1,140 @@
+# Email Touchpoint Audit — 2026-05-04
+
+**Bead:** cf-icww
+**Auditor:** rennala (cfutons crew)
+**Method:** Static analysis of `src/backend/`. Live STAGING_SITE / production sends were not verified — see *Verification gaps* at bottom.
+**Scope:** all welcome + touch emails dispatched via `triggeredEmails.emailContact` (Wix CRM Triggered Emails).
+
+## Legend
+
+- **Trigger** — is the trigger function actually wired to fire?
+- **Send** — does the send call reach `triggeredEmails.emailContact` with the required arguments?
+- **Render** — does a template ID exist in the registry?
+- **Data-bind** — are all variable names referenced by the trigger free of `undefined`/missing-key risk?
+
+`✅` = looks good statically. `⚠️` = works but has a gap. `❌` = broken / dead path.
+**Live render verification** for every row needs STAGING_SITE access (Stilgar) — flagged at end.
+
+## Summary table
+
+| # | Email                                  | Template ID                         | Trigger | Send | Render | Data-bind | Notes |
+|---|----------------------------------------|-------------------------------------|---------|------|--------|-----------|-------|
+| 1 | Welcome drip — exit-intent capture     | `welcome_series_1..3`               | ✅      | ❌   | ✅     | ⚠️       | F1 — empty `recipientContactId` blocks send |
+| 2 | Welcome drip — `triggerWelcomeSeries`  | `welcome_series_1..5`               | ✅      | ❌   | ✅     | ⚠️       | F1 — same root cause |
+| 3 | Welcome drip — member signup           | `welcome_series_1..5`               | ✅      | ✅   | ✅     | ✅        | Wired via `events.js#wixMembers_onMemberCreated` |
+| 4 | Welcome drip — `subscribeToNewsletter` | (none — never queued)               | ❌      | n/a  | n/a    | n/a       | F2 — webMethod inserts subscriber but never queues welcome |
+| 5 | Order confirmation                     | `order_confirmation`                | ❌      | ❌   | ✅     | ✅        | F3 — only-wired handler in `events.js` does not call `sendOrderConfirmation`; the call lives in dead-code copy of `wixEcom_onOrderCreated` inside `emailAutomation.web.js` |
+| 6 | Shipping notification (parcel)         | `order_shipped`                     | ❌      | ✅   | ✅     | ✅        | F4 — `wixEcom_onFulfillmentCreated` defined ONLY in `emailAutomation.web.js`; not wired in `events.js` |
+| 7 | Shipping notification (LTL freight)    | `freight_shipped`                   | ❌      | ✅   | ✅     | ✅        | F4 — same |
+| 8 | Delivery confirmation                  | `delivery_confirmation`             | ❌      | ✅   | ✅     | ✅        | F5 — `wixEcom_onOrderDelivered` defined ONLY in `emailAutomation.web.js` |
+| 9 | Post-purchase Day 3 / 7 / 30 / referral| `post_purchase_1..3`, `post_purchase_referral` | ⚠️ | ✅   | ✅     | ✅        | Fires from `events.js#wixEcom_onOrderCreated` so countdown starts at *order* not *delivery* (CF-nkau spec says delivery) |
+| 10| Post-purchase Day-14 review reward     | `post_purchase_review_reward`       | ❌      | ✅   | ✅     | ✅        | F5 — depends on `wixEcom_onOrderDelivered` |
+| 11| Cart abandonment recovery (3-step)     | `cart_recovery_1..3`                | ✅      | ✅   | ✅     | ✅        | Cron `triggerCartRecoveryCron` + `cartRecovery.web.js` |
+| 12| Browse recovery                        | `browse_recovery_1`                 | ✅      | ✅   | ✅     | ✅        | Cron `triggerBrowseRecoveryCron` |
+| 13| Re-engagement (Day 0 / 7 / 21)         | `reengagement_1..3`                 | ✅      | ✅   | ✅     | ✅        | Cron `triggerReengagementCron` |
+| 14| Winback (UTM-based)                    | (lifecycle templates)               | ✅      | ✅   | ✅     | ✅        | Cron `scanAndTriggerWinbackCron` |
+| 15| Review-request (orders + 7d)           | `post_purchase_2`                   | ✅      | ✅   | ✅     | ✅        | Cron `runReviewRequestEmailsCron`; also wired via `jobs.config` |
+| 16| Wishlist price-drop alerts             | (in `wishlistAlerts.web.js`)        | ✅      | ✅   | ✅     | ✅        | Cron `checkWishlistAlerts` |
+| 17| Tier milestone (gamification)          | `tier_*_approach`/`*_achieved`      | ✅      | ✅   | ✅     | ✅        | Wired via `events.js#wixEcom_onOrderCreated` post-purchase block (CF-8onx) |
+| 18| Contact form (owner notification)      | `contact_form_submission`           | ✅      | ✅   | ✅     | ✅        | `emailService.sendEmail` — sends to `SITE_OWNER_CONTACT_ID` |
+| 19| Contact form (customer auto-reply)     | (none)                              | ❌      | n/a  | n/a    | n/a       | F6 — no customer-side reply email sent (godfrey's PR #15 path) |
+| 20| Swatch request (owner notification)    | `contact_form_submission`           | ✅      | ✅   | ✅     | ✅        | `submitSwatchRequest` |
+| 21| Swatch request (customer confirmation) | `swatch_confirmation`               | ⚠️      | ✅   | ✅     | ✅        | F7 — only fires if a CRM contact already exists for the email; new visitors get nothing |
+| 22| Swatch follow-up sequence              | `swatch_followup_arrived/decide`    | ✅      | ✅   | ✅     | ✅        | Day 3/10 post-ship via cron |
+| 23| Consultation follow-up                 | `consultation_followup`             | ✅      | ✅   | ✅     | ✅        | CF-tcj5 |
+| 24| Newsletter signup confirmation         | (Klaviyo handles via ESP)           | ✅      | ✅   | n/a    | n/a       | `subscribeToNewsletter` calls `_syncToESPInternal` (best-effort) |
+| 25| Birthday reward                        | `birthday_reward_*`                 | ✅      | ✅   | ✅     | ✅        | `birthdayRewardService.web.js` |
+| 26| Gift card delivery                     | (in `giftCards.web.js`)             | ✅      | ✅   | ✅     | ✅        | Two send sites |
+| 27| Fabric sample confirmation/fulfillment | `fabric_sample_*`                   | ✅      | ✅   | ✅     | ✅        | `fabricSampleService.web.js` |
+| 28| Product Q&A owner alert                | `OWNER_EMAIL_TEMPLATE`              | ✅      | ✅   | ✅     | ✅        | `productQA.web.js` |
+| 29| Restock notification                   | `restock_notification`              | ✅      | ✅   | ✅     | ✅        | Triggered from `wixStores_onInventoryVariantUpdated` |
+| 30| Review thank-you                       | `review_thank_you`                  | ✅      | ✅   | ✅     | ✅        | `reviewsService.web.js` |
+| 31| Content schedule cron                  | (publisher, not email)              | ✅      | n/a  | n/a    | n/a       | `processContentScheduleCron` schedules CMS content; no email side-effect |
+
+## Findings (P0/P1) — broken or dead
+
+### F1 — Welcome series for anonymous + member-self-trigger paths fails to send (P1)
+**Where:**
+- `src/backend/newsletterService.web.js:453` (`captureExitIntentEmail`) inserts `recipientContactId: ''`
+- `src/backend/emailAutomation.web.js:487` (`triggerWelcomeSeries`) inserts `recipientContactId: ''`
+
+**Why it breaks:** `processEmailQueue` calls `sendQueuedEmail` (`emailAutomation.web.js:1485`) which `throw new Error('No contact ID for recipient')` when `recipientContactId` is empty. Items retry until `MAX_RETRIES`, then status='failed'.
+
+**Result:** Exit-intent welcome series and member-self-trigger welcome series never deliver. Member-signup path (F3 ✅) works because `wixMembers_onMemberCreated` passes `member._id` as `contactId`.
+
+**Fix:** Resolve a Wix CRM contact via `contacts.appendOrCreateContact({ emails: [{ email }] })` before queueing, then store the returned `contactId` on the queue row.
+
+### F2 — `subscribeToNewsletter` does not actually queue the welcome series (P1)
+**Where:** `src/backend/newsletterService.web.js:333..389` and the comment in `src/backend/styleQuiz.web.js:327` ("subscribeToNewsletter deduplicates silently and triggers the welcome flow").
+
+**Why it breaks:** `subscribeToNewsletter` inserts a `NewsletterSubscribers` row and fires-and-forgets `_syncToESPInternal(...)`. There is no call to `triggerWelcomeSequence`, `triggerWelcomeSeries`, or `enqueueEmail` for `welcome_series_*`. The misleading comment in styleQuiz means callers believe a welcome flow runs.
+
+**Fix:** Either rename `subscribeToNewsletter` to make its scope honest, or call `triggerWelcomeSeries(cleanEmail, '')` (after F1 is fixed) at the success branch.
+
+### F3 — `order_confirmation` template never sends (P0)
+**Where:** Two competing definitions of `wixEcom_onOrderCreated`:
+- `src/backend/events.js:214` — wired by Wix; calls `triggerPostPurchaseSequence` (Day-3+ care guide), **does NOT call `sendOrderConfirmation`**.
+- `src/backend/emailAutomation.web.js:183` — dead code; would call `sendOrderConfirmation` but Wix only auto-registers handlers exported from `backend/events.js`.
+
+**Why it breaks:** Wix Stores' built-in receipt email may still fire (platform-level), but the Velo `order_confirmation` template — the branded one with `firstName`/`orderNumber`/`itemSummary`/`estimatedDays` variables — is dormant. If Velo template was meant to replace the platform default, customers receive a less branded receipt; if it was additive, customers receive *one* receipt instead of two.
+
+**Fix:** Add `sendOrderConfirmation({contactId, email, firstName, orderNumber, total, itemSummary})` at the top of `events.js#wixEcom_onOrderCreated` (after the `if (!email) return` guard). Delete the dead-code copy in `emailAutomation.web.js` to prevent further confusion.
+
+### F4 — Shipping notifications never fire (P0)
+**Where:** `wixEcom_onFulfillmentCreated` is defined ONLY in `src/backend/emailAutomation.web.js:218`. `events.js` has no such handler.
+
+**Why it breaks:** Wix auto-registers event handlers from `backend/events.js`. `.web.js` files expose webMethods — exporting a `wix*_on*` name from them is not an event registration. Both `order_shipped` and `freight_shipped` templates never get triggered by tracking-info creation.
+
+**Fix:** Move `wixEcom_onFulfillmentCreated` body into `events.js` (or have `events.js` import-and-call). The send-side functions (`sendShippingNotification`, `sendFreightShippingNotification`) are correct.
+
+### F5 — Delivery confirmation, Day-14 review reward, NPS survey never fire (P0)
+**Where:** `wixEcom_onOrderDelivered` is defined ONLY in `src/backend/emailAutomation.web.js:264`. `events.js` has no such handler.
+
+**Why it breaks:** Same as F4. Affects:
+- `delivery_confirmation` template (sendDeliveryConfirmation)
+- `post_purchase_review_reward` (CF-qy79 Day-14 review prompt)
+- NPS survey scheduling (CF-1mlj)
+- The "post-purchase sequence starting from delivery date" requirement (CF-nkau spec) — currently the sequence starts from order *creation* via the `events.js` order-created handler.
+
+**Fix:** Add `wixEcom_onOrderDelivered` to `events.js`, copying logic verbatim from `emailAutomation.web.js:264`. Then audit whether the order-created path should still queue post-purchase or whether the Day-3/7/30 timing should reset to delivery date.
+
+### F6 — Contact form has no customer-side reply (P2)
+**Where:** `src/backend/emailService.web.js#sendEmail` (line 105+).
+
+**Why it's a gap:** Only the site owner gets `contact_form_submission`. The submitter sees a UI confirmation ("Thanks!") but no email confirmation lands in their inbox — common reason customers re-submit the form. Bead description references "godfrey's PR #15 path" which suggests an auto-reply was planned.
+
+**Fix:** Resolve/append a CRM contact, then call `triggeredEmails.emailContact('contact_form_auto_reply', contactId, { variables: { customerName, originalMessage } })` after the owner notification. Template needs to be created.
+
+### F7 — Swatch confirmation skipped for new visitors (P2)
+**Where:** `src/backend/emailService.web.js:259..280`.
+
+**Why it's a gap:** `submitSwatchRequest` fires the customer-side `swatch_confirmation` only if `contacts.queryContacts().eq('primaryInfo.email', cleanEmail)` already returns a hit. First-time submitters (no prior CRM contact) get no email. The owner-side notification fires either way.
+
+**Fix:** Use `contacts.appendOrCreateContact({ emails: [{ email: cleanEmail }] })` first to guarantee a contactId, then send.
+
+## Verification gaps — require STAGING_SITE access
+
+Static analysis cannot prove:
+- HTML template renders (no blank emails) — requires triggering each template on staging and inspecting received HTML.
+- Variables actually bind (no literal `{{firstName}}` left over) — requires received-email inspection.
+- CTA links resolve to the correct production URLs with UTM params intact.
+- Send-window throttling (`isInSendWindow`) does not delay transactional emails past their useful TTL.
+- Klaviyo ESP sync fires on `subscribeToNewsletter` (best-effort `_syncToESPInternal` swallows errors silently).
+
+Suggested STAGING_SITE acceptance test plan (hand-off to whoever runs it):
+1. New visitor → exit-intent popup → confirm `welcome_series_1` arrives. **Expect F1 to surface.**
+2. New member signup → confirm `welcome_series_1` arrives. **Expect ✅.**
+3. Test order placement → confirm `order_confirmation` arrives. **Expect F3 to surface (template silent).**
+4. Mark fulfillment with tracking → confirm `order_shipped` or `freight_shipped` arrives. **Expect F4 (silent).**
+5. Mark order delivered → confirm `delivery_confirmation` + Day-14 reward + NPS survey scheduled. **Expect F5 (silent).**
+6. Add to cart, abandon → wait 1 hr → confirm `cart_recovery_1`. Should ✅.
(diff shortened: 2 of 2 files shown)


DreadPirateRobertz:
## 5-Agent Review — FIXES REQUIRED

### code-reviewer (BLOCKING)
- **F5 / Row #6 / Row #10 are now factually wrong.** Audit claims `wixEcom_onOrderDelivered` is "defined ONLY in emailAutomation.web.js" with no events.js handler. Verified at `src/backend/events.js:373`: `export async function wixEcom_onOrderDelivered` exists and dynamic-imports `handleOrderDelivered`. The neighborhood comment even cites the cf-jmmk fix (PR #1141, just shipped). Merging this audit as-written will create false-positive bead trails for delivery_confirmation, Day-14 review reward, and NPS. **F5 must be deleted, re-scoped to the legitimate Row #9 concern (countdown starts at order not delivery), or re-verified against current main.**
- F3 (sendOrderConfirmation never wired) — accurate.
- F4 (no wixEcom_onFulfillmentCreated in events.js) — accurate.
- F1 (recipientContactId='' silent fail) — accurate, but line number drifted (audit says 487, actual 498).
- F2 (subscribeToNewsletter doesn't enqueue welcome_series) — accurate.

### silent-failure-hunter (HIGH)
- **Row #14 (Winback) and Row #16 (Wishlist alerts) are opaque** — "lifecycle templates" and "in wishlistAlerts.web.js" with no template ID, no fix path, no verify step.
- **Row #24 contradiction.** Marked ✅ for Klaviyo path but the verification-gaps section admits `_syncToESPInternal` swallows errors silently. Should be ⚠️.
- **Row #31 (content schedule cron)** — no email side-effect; either remove or justify scope.
- **F-numbering collision** — Row #5 is F3, but rows #8 and #10 both labeled F5 without obvious grouping signal.

### code-simplifier (HIGH)
- F5 fix sentence is now self-contradictory ("add to events.js" — already there).
- Verification-gaps section claims HTML render / variable binding / CTAs / throttling all need staging, but plan only covers triggers + arrival. Mismatch.
- "Live render verification flagged at end" appears twice (legend + verification-gaps).
- "godfrey's PR #15 path" in F6 — opaque internal reference; link or describe.

### comment-analyzer (HIGH)
- **Line numbers in F1/F2/F3/F4/F5 are load-bearing.** F1 already drifted (487 → 498). Recommend grep-stable anchors (function names + nearest comment) instead of raw line numbers.
- **Central thesis** "Wix only auto-registers handlers exported from backend/events.js" is uncited. Link the Velo docs page that states this contract — it's the load-bearing claim tying F3/F4/F5 together.

### pr-test-analyzer (HIGH)
- **15+ of 31 rows have no acceptance test in the 8-step plan.** Rows 2, 9, 12–17, 22, 23, 25–30 are asserted ✅ in the table without runtime evidence. Either expand the plan or scope it explicitly as "P0/P1 only."
- **Step 5 expects "F5 (silent)"** — but F5 is wired in main. When the staging tester runs step 5 and delivery_confirmation DOES arrive, the plan will mislead them.
- **No HTML/variable inspection step** despite verification-gaps calling out `{{firstName}}` leftover risk. Add explicit "view raw email source, grep for `{{` / `undefined`".
- **No UTM/link/throttling steps** despite all three being named as concerns in verification-gaps.
- **Step 6 cart-abandonment "wait 1 hr"** — no staging knob to short-circuit the timer; tester hits a wall.

---

**Verdict: FIXES REQUIRED.**

Blocker: F5 is based on an outdated reading of events.js — cf-jmmk wired the handler in PR #1141 (just landed). Merging the audit as-is will corrupt downstream bead trails. Re-verify against current main, then resolve the test-plan / row-coverage mismatches.

DreadPirateRobertz:
Refresh pushed (`47102a81`) addressing the 5-agent review:

**code-reviewer (BLOCKING):**
- F5 — kept ❌ for current main (PR #1141 still OPEN, not merged) but added §"Status delta vs. PRs in flight" so the reader sees the impending flip and the new Row 9 duplicate-queue concern when both `wixEcom_onOrderCreated` and `wixEcom_onOrderDelivered` enqueue post-purchase. F5 will flip to ✅ on the same merge that ships #1141; the audit explicitly says so.
- F1/F2/F3/F4 — replaced all line numbers with grep-stable symbol anchors (function/method names). Future drift won't invalidate citations.

**silent-failure-hunter:**
- Row 14 (winback) — `lifecycle_winback_30d/60d/90d`, anchor `lifecycleCron.web.js#scanAndTriggerWinback`.
- Row 16 (wishlist alerts) — `wishlist_price_drop`, anchor `wishlistAlerts.web.js#checkWishlistAlerts`.
- Row 24 (Klaviyo) — adjusted ✅→⚠️ for `_syncToESPInternal` swallowing errors.
- Row 31 (content schedule) — removed; no `triggeredEmails.emailContact` site, out of scope.
- F-numbering — each finding now has its own §; rows that share a finding cite the finding ID.

**code-simplifier:**
- F5 fix sentence rewritten to reflect "fix in flight via #1141" rather than the contradictory "add to events.js" (it's already there in #1141).
- §"Verification gaps" deduped — single mention, expanded into a structured table + revised 8-step plan.
- F6 — clarified "PR #15" as external/historical reference, not a code path.

**comment-analyzer:**
- Symbol anchors replace line numbers throughout.
- Wiring contract now cited (https://dev.wix.com/docs/velo/api-reference/wix-stores-backend/events) as a load-bearing claim with explicit refutation criteria.

**pr-test-analyzer:**
- Test plan scope now explicitly **P0/P1 only** — `✅` rows in the table are statically asserted, not runtime-asserted; STAGING expansion for the rest is tracked separately.
- Step 5 (was "F5 silent") — now reads "F5 (Row 8/10) — delivery + Day-14 reward + NPS" with arrival assertions for all three chained outcomes; will pass once #1141 merges.
- New explicit HTML render / `{{var}}` grep / UTM / throttling steps under §"Verification gaps" in a table form, and each numbered step references the corresponding gap.
- Cart-abandonment "wait 1 hr" replaced with two concrete fast-forward knobs (direct `EmailQueue` insert with `scheduledFor=now`, or `processEmailQueue({ now: now+3600_000 })`).

Bead-trail flag added (§"Status delta"): cf-i23b and cf-icdc are CLOSED without merged PRs on main — flagging back to mayor as drift, since the audit treats current `main` as source of truth.

Ready for re-review.

DreadPirateRobertz (commented):

## Summary
This is a docs-only PR (two `.md` files, zero code). No new tests are needed for the diff itself. The audit's job is to identify untested/dormant runtime behaviour and recommend how to verify the in-flight fixes — and on that axis the refresh at `47102a81` is in good shape. Prior pr-test-analyzer feedback from the 5-agent pass was addressed concretely. A few unit-test recommendations are still missing for the in-flight fix beads (cf-xdji, post-merge-#1141 dedup, Row 24 silent-failure observability) — these are improvements, not blockers.

## Prior pr-test-analyzer feedback — addressed?
1. "15+ rows have no acceptance test" → **Addressed** via explicit scoping ("Scope of acceptance test plan: P0/P1 only — `✅` rows are statically asserted, not runtime-asserted; STAGING expansion tracked as `cf-icww.followup-tests`"). Defensible scoping retreat, but the follow-up bead must actually get filed for the `✅` rows or this regresses.
2. "Step 5 expects F5 silent, but F5 wired in main" → **Addressed**. Step 5 now reads "F5 (Row 8/10) — delivery + Day-14 reward + NPS" with three chained arrival assertions that flip to pass once #1141 merges.
3. "No HTML/variable inspection step" → **Addressed**. Verification-gaps table now includes "grep raw HTML for `{{` / `}}` / `undefined` / `null` strings".
4. "No UTM/link/throttling steps" → **Addressed**. Verification-gaps table covers UTM survival, CTA URL hosts, and send-window throttle test against TTL.
5. "Step 6 cart-abandonment wait 1 hr unrealistic" → **Addressed**. Replaced with two concrete fast-forward knobs: direct `EmailQueue` insert with `scheduledFor=now`, or `processEmailQueue({ now: now+3600_000 })`.

## Are the testability recommendations concrete?
**Mostly yes, for the F1–F7 fixes:**
- **cf-jmmk (F5, PR #1141)**: Step 2 asserts (a) `delivery_confirmation` arrives within 5 min, (b) `EmailQueue` row has `scheduledFor = deliveredAt + 14d`, (c) `Surveys` row inserted with `deliveredAt + 7d`. Three observable invariants. Concrete.
- **cf-xdji (F1+F7 resolveContactId)**: Step 4 explicitly asserts "EmailQueue row: `recipientContactId` is non-empty (proves `resolveContactId` ran)". That's exactly the right behavioural assertion to catch the `: ''` regression. Step 5 is a regression check on the member-signup contactId path. Step 7 covers the new-visitor swatch path.
- **cf-hafn (F6 auto-reply)**: Step 6 asserts owner + customer emails both arrive and customer body echoes original subject+message.
- **F2 / F3 / F4**: Steps 1, 3, 8 all have concrete trigger + arrival + body assertions.

## Critical gaps in testability recommendations (8-10)
**8/10 — Row 9 post-merge dedup test is missing.** The audit correctly flags that once #1141 ships, both `wixEcom_onOrderCreated` and `wixEcom_onOrderDelivered` will enqueue `post_purchase_*` → duplicate Day-3/7/30 sends to every customer. The recommended fix ("dedupe by `sequenceType + orderNumber` in `queueEmail`") has no corresponding unit-test recommendation. A vitest test against the `wix-data` mock asserting "second `queueEmail` call with same `sequenceType+orderNumber` is no-op" would prevent regression. Cost: ~30 lines. Bug it prevents: every customer receiving 2× post-purchase emails for ~28 days.

## Important improvements (5-7)
**7/10 — `resolveContactId` deserves a unit test.** cf-xdji introduces a small, load-bearing helper (`appendOrCreateContact` wrapper) used at three call sites (F1 exit-intent, F1 triggerWelcomeSeries, F7 swatch). The audit specs the runtime invariant (`recipientContactId !== ''`) but not the helper's contract. Three vitest cases against the existing `wix-crm-backend` mock pattern (`tests/__mocks__/`):
1. Existing contact email → returns existing `_id`
2. New email → calls `appendOrCreateContact`, returns new `_id`
3. Empty/whitespace email → throws or returns sentinel (matching the audit's wiring contract)

Criticality: 7/10 because the runtime test in step 4 will catch the regression, but a unit test catches it pre-deploy and at the helper boundary, not after EmailQueue has already accumulated bad rows.

**6/10 — Row 24 `_syncToESPInternal` silent-failure observability test.** Audit correctly downgraded to ⚠️ per silent-failure-hunter feedback, but doesn't recommend that the swallowed `.catch(() => {})` at minimum increment a counter or log via the existing logger so failures become queryable. A test would be: force the Klaviyo fetch mock to reject, assert the failure is recorded somewhere observable (log spy or metric counter). Cost: ~20 lines. Bug it prevents: undetected ESP outages costing weeks of newsletter signups.

## Test-quality issues
None — no tests landed in this PR.

## Positive observations
- **Symbol anchors instead of line numbers** (per comment-analyzer's prior feedback) make every test assertion location-independent. F1 line drift (487→498) was the trigger; the fix generalises across all findings.
- **Test plan ordering matches Recommended Fix Order** — iterative remediation is verifiable step by step rather than as a big-bang.
- **Each test step lists trigger + arrival + HTML/UTM assertion** — unusually concrete for an audit doc; most "audits" stop at "verify X" without saying how.
- **The `recipientContactId !== ''` invariant** for F1 is exactly the right behavioural assertion (resilient to refactors of `resolveContactId`'s internals; only fails if the contract regresses).
- **Row 9 duplicate-queue concern was proactively flagged** with a recommended dedup key (`sequenceType + orderNumber`) — that's a testable spec ready for the next bead, even though the unit test itself wasn't recommended (see 8/10 gap above).
- **Status-delta table** distinguishes "audit of current main" from "in-flight-fix prediction" — prevents the false-positive bead trails miquella's code-reviewer agent flagged in the prior pass.
- **Bead-trail drift flag** (cf-i23b, cf-icdc CLOSED without merged PRs) is the correct escalation; not strictly a test concern but it prevents the audit from being read as "F3/F4 are fixed."

## Verdict
**Recommend approve from a test-coverage standpoint.** Prior pr-test-analyzer feedback was addressed substantively (not just acknowledged). The audit's runtime test plan is concrete enough that a STAGING_SITE walkthrough can actually decide each ✅/❌. The three improvement-tier gaps (cf-jmmk Row 9 dedup test, cf-xdji unit test, Row 24 observability) are best filed as their own beads against the implementation PRs, not as blocking changes to this audit doc.

DreadPirateRobertz (commented):

## Honest scoping note

This is a docs-only PR (267/0 add/del across `docs/email-audit-2026-05-04.md` + `docs/decisions/cf-3l0d-subscribeToNewsletter-decision.md`) in a Wix Velo JavaScript codebase with no TypeScript and, by my read of both files, no JSDoc `@typedef` recommendations either. So the surface for type-design analysis is narrow. I'm not going to invent findings — what follows is limited to the four type-shaped artifacts the audit actually proposes or relies on:

1. The `resolveContactId(email, firstName?) -> contactId` helper proposed by cf-xdji (referenced by F1, F6, F7, and the F2 decision doc).
2. The `EmailQueue` row contract — load-bearing for F1, the F2 decision, and the post-merge dedupe concern in Row 9.
3. The implicit `contact_form_auto_reply` variables shape mentioned in F6.
4. The "Wiring contract" — handlers must live in `events.js` (a pseudo-type contract over the file system).

---

## Type 1: `resolveContactId(email, firstName?) -> contactId`

### Invariants identified
- **Postcondition: return value is non-empty** — the entire reason this helper exists is that `EmailQueue.recipientContactId === ''` is the silent-failure condition behind F1. The helper is a value-producer whose contract is "never returns `''` or `null`/`undefined`."
- **Idempotent on email** — implied by `contacts.appendOrCreateContact(...)`: two calls with the same `email` should return the same `contactId`.
- **Email is the identity key**, `firstName` is enrichment-only — the audit's signature `(email, firstName?)` correctly encodes that `firstName` is optional and not part of identity.

### Ratings
- **Encapsulation: 7/10.** The helper hides `wix-crm-backend` plumbing behind one function. Good. But the audit doesn't say where it lives or what its visibility is — F2's Option B explicitly raises the `Anyone` vs `SiteMember` permissions question, which is a real encapsulation concern (a webMethod with `Anyone` permission that auto-creates CRM contacts is a spam vector). The decision doc's recommendation to use `_triggerWelcomeSeriesInternal` is the right shape; same pattern should apply to `resolveContactId` (internal-only, called by `Anyone` webMethods after their own rate-limit/dedup gate).
- **Invariant expression: 5/10.** In a TS codebase the signature would be `(email: Email, firstName?: string) => Promise<ContactId>` with `ContactId` a branded non-empty string. In Velo JS, the only expression mechanism is JSDoc + a runtime invariant check. The audit recommends neither. This is a missed opportunity — adding a one-line `@returns {Promise<string>} non-empty CRM contact ID` JSDoc would at least give the WIX-Studio hint surface something to display on hover.
- **Invariant usefulness: 9/10.** The non-empty postcondition prevents the exact bug the audit is filed against (F1). High signal, low cost.
- **Invariant enforcement: 4/10 (as designed).** The audit doesn't specify enforcement. Recommendation: the helper should `throw` if `appendOrCreateContact` returns falsy/empty rather than returning `''` — turn the silent-failure condition into a loud-failure condition at the boundary, so callers (`captureExitIntentEmail`, `triggerWelcomeSeries`, `submitSwatchRequest`, future F6 wiring) cannot accidentally re-introduce the F1 bug by forgetting to check the return value.

### Concrete suggestion (would not bloat the audit)
Add one paragraph under §"Recommended fix order" step 3:

> `resolveContactId(email, firstName?)` MUST throw on empty contactId rather than return `''`. JSDoc `@returns {Promise<string>}` and `@throws {Error}` so the WIX-Studio editor surfaces the contract on hover. Callers wrap with try/catch and degrade to the existing dedup/skip path; they never enqueue with an empty `recipientContactId`.

That single sentence elevates `resolveContactId` from "convenience helper" to "invariant-enforcing boundary" and closes F1 by construction rather than by convention.

---

## Type 2: `EmailQueue` row

### Invariants identified (extracted from the audit)
- `recipientContactId` is non-empty (the F1 invariant — currently violated at two call sites).
- `(recipientEmail, sequenceType, sequenceStep)` is the dedup key (per F2 Option B's reliance on the existing dedup guard).
- After post-F5 merge: `(sequenceType, orderNumber)` becomes a NEW dedup key for post-purchase rows queued by both `wixEcom_onOrderCreated` and `wixEcom_onOrderDelivered` (Row 9).
- `scheduledFor` is a future-or-now epoch — short-circuit knob in §"Cart-abandonment timing knob" is a tester escape hatch, but the production invariant is implicit.
- `status` is a finite state machine: presumably `queued` → (`sent` | `failed`) with retry < `MAX_RETRIES`. The audit references this but doesn't spell it out.

### Ratings
- **Encapsulation: 6/10.** The audit treats `EmailQueue` as a plain `wixData` collection — anyone with permission can `wixData.insert('EmailQueue', {...})` with arbitrary shape. The dedup guard lives inside `enqueueEmail` / `queueEmail` but the audit also discusses bypassing those wrappers in §"Cart-abandonment timing knob" by inserting directly. That's fine for testing but it tells you the invariant is enforced by convention, not by construction.
- **Invariant expression: 3/10.** Nothing in the doc defines the row shape. A `@typedef` block at the top of `emailAutomation.web.js` would help — `@typedef {{ recipientContactId: string, recipientEmail: string, sequenceType: string, sequenceStep: number, templateId: string, scheduledFor: number, status: 'queued'|'sent'|'failed' }} EmailQueueRow`. Optional-but-cheap.
- **Invariant usefulness: 9/10.** The dedup-key invariants are the real load-bearing constraints in this audit. They directly drive the F2 decision (Option B is safe BECAUSE the dedup guard catches redundant calls) and the post-F5 dedupe recommendation.
- **Invariant enforcement: 5/10.** The dedup guard is enforced in one wrapper but not at the collection level. Adding a single `enqueueEmail` chokepoint (no direct `wixData.insert('EmailQueue', ...)` anywhere in the codebase) would make the dedup-key invariant unbreakable.

### Concrete suggestion
The audit's Row 9 dedupe recommendation ("dedupe by `sequenceType + orderNumber` before insert in `queueEmail`") is the right move, but it should be stated as: *all `EmailQueue` writes go through `enqueueEmail`; direct `wixData.insert('EmailQueue', ...)` is forbidden going forward.* That converts a row-level invariant into a chokepoint-level invariant — much harder to violate by accident.

---

## Type 3: `contact_form_auto_reply` variables (F6)

### What the audit actually says
F6 says: "call `triggeredEmails.emailContact('contact_form_auto_reply', contactId, { variables })`. Template needs to be created."

That's it. No shape for `variables`. No `customerName` / `subject` / `message` / `replyEta` design — those four fields appear nowhere in the diff, despite being called out in the prompt. The audit punts on the variables-object design.

### Ratings
- **Encapsulation: n/a** — type doesn't exist yet.
- **Invariant expression: n/a.**
- **Invariant usefulness: n/a.**
- **Invariant enforcement: n/a.**

### Concrete suggestion
Before F6 implementation lands, the audit (or the follow-up bead `cf-icww.followup-f6`) should define the variables object explicitly: `{ customerName: string (non-empty, sanitized to ≤80), subject: string (≤120), message: string (≤2000, echoed verbatim — beware self-XSS in HTML render), replyEta: string ('within 1 business day' default) }`. The `message` field in particular has a security invariant the audit doesn't mention: it's user-supplied and rendered into an email template. The existing `sanitize(str, maxLen)` truncation pattern (known bug STORY-001 truncates 22-char gift-card codes when called as `sanitize(code, 20)`) needs to be applied with the right `maxLen` per field, and the audit should call that out so the implementer doesn't trip the same trap.

This is a real gap. F6 implementation without a defined variables-object contract will produce another `recipientContactId: ''`-class silent failure, just on a different field.

---

## Type 4: The "Wiring contract"

> Backend event handlers are auto-registered only from `backend/events.js`. Functions matching the `wix*_on*` naming pattern that live in any other backend file ... are not wired.

### Ratings
- **Encapsulation: 8/10.** Wix enforces this at the platform level. Not something the codebase can change.
- **Invariant expression: 7/10.** The audit now cites the docs URL (good — addresses comment-analyzer's concern). The naming convention `wix*_on*` is itself a cheap form of type — Wix uses the function name as the type-tag for which event to dispatch. That's a stringly-typed dispatch table, but it's at least conventional.
- **Invariant usefulness: 10/10.** This contract is the entire reason F3/F4/F5 exist. The audit is right to elevate it.
- **Invariant enforcement: 6/10.** Wix enforces the *positive* direction (handlers in `events.js` get called). It does NOT enforce the *negative* direction (handlers OUTSIDE `events.js` should not exist with the `wix*_on*` name pattern). That's how the codebase ended up with dormant duplicates in `emailAutomation.web.js` for `wixEcom_onOrderCreated`, `wixEcom_onFulfillmentCreated`, and `wixEcom_onOrderDelivered`. A simple ESLint rule (`no-wix-event-handlers-outside-events-js`) would convert this from "Wix-platform-enforced positive contract" to "ESLint-enforced bidirectional contract" and prevent F3/F4/F5-class drift from recurring.

### Concrete suggestion
Add a P3 bead alongside `cf-icww.followup-tests`: "Lint rule forbidding `^wix\w+_on\w+` exports in any `src/backend/*.web.js` file." It's the cheapest possible way to make the wiring contract enforceable by construction rather than by code review.

---

## Were prior type-design comments from miquella's 5-agent pass addressed?

Looking at the prior 5-agent review (DreadPirateRobertz comment) and the refresh response, no comment in that round was specifically a type-design comment. The closest were:

- **comment-analyzer's "line numbers are load-bearing"** — refresh replaced numeric line refs with symbol anchors throughout. That's a type-of-citation improvement, addressed.
- **silent-failure-hunter's Row 24 contradiction** — refresh moved `_syncToESPInternal` Klaviyo path from ✅ to ⚠️ to reflect the swallowed-error invariant. That's a correctness-of-status-symbol fix, addressed.

The substantive type-design surface — `resolveContactId` postconditions, `EmailQueue` row shape, `contact_form_auto_reply` variables — was not raised by miquella's pass and remains unaddressed in this PR. That's a real gap, but it's a gap in the *original review coverage*, not a regression in this refresh.

---

## Verdict

The PR is two well-written audit/decision documents in a JavaScript codebase. Type-design analysis applies thinly:

- **`resolveContactId` proposal:** signature is correct (encodes the right invariants). Should specify *throw on empty* and add JSDoc. Minor doc-only fix.
- **`EmailQueue` shape:** the audit's dedup-key reasoning is correct and load-bearing. Should recommend chokepoint-level enforcement (all writes via `enqueueEmail`).
- **`contact_form_auto_reply` variables:** undefined in the audit. Real gap. Should be specified before F6 implementation.
- **Wiring contract:** correctly cited and elevated. Should propose an ESLint rule to make it bidirectionally enforceable.

None of these are blocking for a docs PR. All four are cheap follow-ups for the implementation beads (cf-xdji, the F6 follow-up, and a new lint-rule bead). I'd merge this audit as-is and file the four follow-ups against the implementation work.

DreadPirateRobertz (commented):

## Verdict: doc PR — claims verified against `main`; audit itself surfaces the silent-failure backlog correctly. A few additional silent-failure patterns inside the audited surface are NOT called out and should be added before this is treated as the canonical email map.

### Prior 5-agent feedback (this PR has zero review comments yet — no prior round to verify)
`gh api repos/.../pulls/1140/comments` returns `[]`. The audit doc references "miquella feedback" (symbol anchors not line numbers) and "silent-failure-hunter feedback" (Row 24 downgrade `✅`→`⚠️` for the Klaviyo `.catch(() => {})`), so an earlier pass clearly happened off-PR. Both pieces of feedback are reflected in the doc as committed.

### Audit findings I verified against current `main` code paths

All six P0/P1 claims line up with code on the PR head. Receipts:

| Finding | Claim | Verified at |
|---|---|---|
| F1 | `recipientContactId: ''` blocks `sendQueuedEmail` | `newsletterService.web.js:453`, `emailAutomation.web.js:487`, `emailAutomation.web.js:1488` (`throw new Error('No contact ID for recipient')`) |
| F2 | `subscribeToNewsletter` does not queue welcome | `newsletterService.web.js:333-389` — only `wixData.insert('NewsletterSubscribers', ...)` + `_syncToESPInternal(...).catch(() => {})`. No `triggerWelcomeSeries`/`enqueueEmail` call. |
| F3 | `events.js#wixEcom_onOrderCreated` does not call `sendOrderConfirmation` | `events.js:214-309` — calls `triggerPostPurchaseSequence`, `handleOrderStatusChange`, gamification, referral, swatch attribution. No `sendOrderConfirmation`. The `emailAutomation.web.js:183` duplicate is dormant. |
| F4 | `wixEcom_onFulfillmentCreated` defined only in `.web.js` | `grep wixEcom_onFulfillmentCreated src/backend/events.js` → empty. Only definition is `emailAutomation.web.js:218`. |
| F5 | `wixEcom_onOrderDelivered` defined only in `.web.js` on `main` | `grep wixEcom_onOrderDelivered src/backend/events.js` → empty. Only definition is `emailAutomation.web.js:264`. |
| F7 | Swatch confirmation skipped when no CRM contact exists | `emailService.web.js:264-280` — `if (contactResult.items.length > 0)` skips and the `else` arm is missing entirely. No `logError`, no audit-event row, the visitor sees the owner-side success without ever knowing they're in the no-confirmation cohort. |

The "Wix auto-registers handlers from `events.js` only" wiring contract the audit hangs F3/F4/F5 on is correct per Wix Velo docs. F3 and F5 are confirmed broken on `main` even though their bead beads are closed.

### Silent-failure patterns the audit MISSED inside its own scope

**1. `unsubscribeUrl` failure swallowed, falls back to static URL** — `emailAutomation.web.js:1493-1502`
```js
try {
  const secret = await getSecret('UNSUB_TOKEN_SECRET');
  ...
  unsubscribeUrl = await buildUnsubscribeUrl(...);
} catch {
  unsubscribeUrl = 'https://www.carolinafutons.com/unsubscribe';
}
```
- Empty `catch {}` with NO logging. If the secret is rotated/missing or `buildUnsubscribeUrl` throws (signing error, malformed email), every queued email silently gets the static fallback URL — meaning recipients land on a generic page that cannot identify them, breaking one-click unsub for that campaign.
- This catch block could hide: secret-store outages, JWT signing failures, malformed email inputs, network errors loading the secret. None get reported.
- Fix: at minimum `logError('sendQueuedEmail:unsubscribeUrl', err)` and consider failing-loud rather than masking — CAN-SPAM compliance depends on a working unsub link, not a fallback that pretends to unsubscribe everyone.

**2. `EmailQueue` rows transitioning to `status='failed'` are not surfaced anywhere** — `emailAutomation.web.js:1235-1249`
```js
const newStatus = attempt >= MAX_RETRY_ATTEMPTS ? 'failed' : 'pending';
...
await wixData.update('EmailQueue', { ...item, status: newStatus, lastError: err.message ... });
```
- `lastError` is written to the row, but there is no `logError` call when an item terminally fails, no alert, no daily digest counting failed rows. F1 today produces N failed rows per affected user; the only way an operator finds out is to manually query `EmailQueue` for `status='failed'`.
- Recommend the audit add an explicit "F8 (P1) — terminal queue failures are silent" finding: route `lastError` to `logError('processEmailQueue:terminal', err, { templateId, recipientEmail, sequenceType, attempt })` on the transition into `'failed'` so Sentry catches the F1/F3-class blast radius without staging access.

**3. Table Row 24 caveat is real but understated** — `newsletterService.web.js:380`
```js
_syncToESPInternal(cleaned, source).catch(() => {});
```
- Row 24 already downgrades to `⚠️` in the table. But the same `_syncToESPInternal` is exported as the `syncToESP` webMethod and is referenced as the load-bearing path for "newsletter signup → Klaviyo". An empty `.catch(() => {})` here means a Klaviyo outage, an expired API key, a 4xx from the Profiles API, or even a `wix-fetch` import error all look identical to success from the caller's perspective. The internal function does return `{ synced: false, reason }` for known failure modes — but the OUTER `.catch(() => {})` wraps `throw`s that escape the inner try (e.g. `loadESPSecrets()` rejection, JSON parsing errors). Recommend: replace with `.catch(err => logError('subscribeToNewsletter:espSync', err, { source: cleanSource }))`.

**4. F6 fix recipe risks recreating F7-shape silent skip** — audit §F6 says "Resolve/append a CRM contact (use cf-xdji's helper once landed), then call `triggeredEmails.emailContact('contact_form_auto_reply', contactId, ...)`. Template needs to be created."
- The recipe is correct, but please add an explicit assertion in the bead description: if `resolveContactId` returns `''`/null, the implementation must `logError('contactFormAutoReply:noContactId', ...)` and surface to the user a deliverability warning — not silently skip the auto-reply the way `submitSwatchRequest` does today (F7). The audit currently treats F7 as "fix the contact resolution" but does not call out that the existing `if (length > 0)` arm is itself a silent failure pattern that should not be copy-pasted into the F6 fix.

**5. F1 fix recipe needs a guardrail clause** — audit §F1 fix-in-flight says cf-xdji's `resolveContactId` will be called before queueing.
- Please specify in the audit (or in cf-xdji acceptance criteria) what happens if `appendOrCreateContact` itself fails. Current pattern across the file is `console.error(...)` + return failure object. For the welcome-queue flow specifically, the queue insert should be GATED on a non-empty contactId — silently inserting an empty `recipientContactId` to "unblock" the visitor (so they get the discount code) just recreates F1. Audit should explicitly forbid the "insert anyway with `''`" fallback in the cf-xdji implementation note.

### Patterns the audit handles well

- The Wiring Contract section explicitly anchors the F3/F4/F5 logic to a falsifiable claim ("if this contract is wrong, F3/F4/F5 collapse") — exactly the right framing for a static analysis.
- Row 9 dedupe risk after F5 lands (post-purchase double-queue from both `onOrderCreated` and `onOrderDelivered`) is exactly the kind of forward-looking silent failure that gets missed in single-finding remediation.
- The "Status delta vs. PRs in flight" table catching that F3/F4 beads are CLOSED with no merged PR is a real operational silent failure (bead-trail drift) and the audit correctly flags it back to mayor.

### Net recommendation

Audit doc is high-quality and the findings hold up. **Land it**, but please address the five gaps above as a follow-up commit on the same PR or as cf-icww.followup-silent-fails — particularly F8 (terminal `status='failed'` rows are invisible) which is silently masking F1's blast radius today.

DreadPirateRobertz (commented):

Scope: prose / comment / cross-reference rot in `docs/email-audit-2026-05-04.md` and `docs/decisions/cf-3l0d-subscribeToNewsletter-decision.md` at HEAD `47102a81`. Verified claims against `main` and against the audit's stated base commit `4dabfd96`.

---

## Critical issues — factually wrong

### CR-1. Row #14 (Winback) — wrong file anchor + invented template IDs
**Location:** `docs/email-audit-2026-05-04.md` Row 14 of summary table.
**Audit claim:**
> Template IDs `lifecycle_winback_30d`, `lifecycle_winback_60d`, `lifecycle_winback_90d` per `lifecycleEmailTemplates.js`. Anchor `lifecycleCron.web.js#scanAndTriggerWinback`.

**Reality (verified at `4dabfd96` and on `main`):**
- `scanAndTriggerWinback` is exported from `src/backend/marketingSequences.web.js` (line 334), **not** `lifecycleCron.web.js`.
- The strings `lifecycle_winback_30d/60d/90d` appear nowhere in the repo (`grep -rn lifecycle_winback_30d src/` returns 0 matches).
- `src/backend/lifecycleEmailTemplates.js` is a HTML-template-generator file for Day-7 / Month-1 / Year-1 lifecycle care emails — it has nothing to do with the winback flow.
- Real winback templateIds are loaded dynamically by `loadActiveSteps('winback')` from the `MarketingSequenceSteps` CMS collection (`marketingSequences.web.js:301`). They are not statically declared in any source file, so the audit cannot assert `Render: ✅` from static analysis alone.

**Why this matters:** This is the kind of confidently-wrong citation that future agents will trust. If anyone tries to grep for `lifecycle_winback_30d` to fix or test this row, they will hit zero matches and assume the row is broken. Worse, the static `Render: ✅` claim is not actually defensible — the templateIds live in a Wix CMS collection that static analysis cannot inspect.

**Suggested fix:** Replace anchor with `marketingSequences.web.js#scanAndTriggerWinback`. Replace the template-ID column with "(loaded from `MarketingSequenceSteps` CMS — sequenceType `winback`)". Demote `Render` from ✅ to ⚠️ since static analysis cannot verify CMS-driven template registration.

---

### CR-2. Row #17 (Tier milestone) — minor template-ID hand-wave
**Location:** Row 17.
**Audit claim:** Template ID `tier_*_approach`/`*_achieved`.
**Reality:** Actual IDs are `tier_mountain_guide_approach`, `tier_mountain_guide_achieved`, `tier_summit_master_approach`, `tier_summit_master_achieved` (`emailAutomation.web.js:121-124`). Glob-style ID is fine for a summary, but a future maintainer trying to find these will grep `tier_*_approach` and get nothing. Cite at least one real ID or the literal `SEQUENCES.tierMilestone.steps` anchor.

---

## Improvement opportunities

### IM-1. "Status delta vs PRs in flight" cf-i23b / cf-icdc drift claim
**Location:** §"Status delta", and F3/F4 fix-in-flight sentences.
**Current state:** Claims cf-i23b (F3) and cf-icdc (F4) are "bead CLOSED 2026-05-05; no merged PR observed on main." This is a strong assertion that, if wrong, makes the whole audit look careless. From this radahn session I cannot directly query the bead system (Dolt) to verify state, and the audit gives the reader no way to verify either.
**Suggestion:** Either (a) include the bead-state evidence inline (e.g. "`bd show cf-i23b` ⇒ status=closed, PRs:[]" output snippet), or (b) reframe as "bead-trail audit forwarded to mayor — outcome pending." The current wording mixes conclusion and accusation without a citation.

### IM-2. "as of 2026-05-04 / refreshed 2026-05-05" date stamps will rot
**Location:** Title (`Email Touchpoint Audit — 2026-05-04 (refreshed 2026-05-05)`), and references to "as of 2026-05-04" / "as of refresh" elsewhere.
**Issue:** This is an audit document — its findings are point-in-time by design, so the date is load-bearing. But once cf-jmmk / cf-xdji land, F5/F1/F7 will flip and the title will look stale even though the audit was correct. There is no built-in way for a future reader to know whether this audit is still authoritative.
**Suggestion:** Add a §"Freshness" line near the top: "Re-verify before citing if any of cf-jmmk, cf-xdji, cf-i23b, cf-icdc landings have shipped to main. Quick check: grep `events.js` for `wixEcom_onOrderDelivered` and `wixEcom_onFulfillmentCreated` exports."

### IM-3. Wiring-contract URL — needs anchor specificity
**Location:** §"Wiring contract":
> Source: Wix Velo "Backend Events" reference (https://dev.wix.com/docs/velo/api-reference/wix-stores-backend/events) — handlers are registered by exporting from `backend/events.js`.

**Issue:** The cited URL is a generic API index. The load-bearing claim ("auto-registration only from `backend/events.js`") is not actually stated explicitly on that page in the form the audit quotes. miquella's comment-analyzer pass flagged exactly this ("Link the Velo docs page that states this contract") and the refresh added a URL but not a quote/page-section that actually says it. The contract is empirically true (verified by every event handler in events.js working and the duplicates in `.web.js` files being dormant), but the citation does not back the wording.
**Suggestion:** Either quote a specific paragraph from the linked page, OR link the Wix Velo "Web Modules" doc + the `backend/events.js` doc as two separate sources (web modules = SPI, events.js = subscriptions). Or hedge to "empirically observed; Wix has not formally documented the contract."

### IM-4. Decision doc cf-3l0d Bucket-2 caller list is mixed scope
**Location:** `docs/decisions/cf-3l0d-subscribeToNewsletter-decision.md` "Caller inventory" table.
**Issue:** The table mixes page files (`Blog.js`, `Home.js`, `masterPage.js`, `Newsletter.js`, `Thank You Page.js`) with backend modules (`styleQuiz.web.js`, `http-functions.js`) and public modules (`exitIntentCapture.js`, `FooterSection.js`). Page-file callers may or may not actually call `subscribeToNewsletter` directly — page files often delegate to public/backend modules. I did not verify each one. If even one of the 9 page-file callers is wrong, the "9 of 10 callers expect welcome behaviour but don't get it" headline number rots.
**Suggestion:** Verify each page-file row by grepping `import.*subscribeToNewsletter` and citing the actual import statement. If the callers compose via a wrapper, cite the wrapper, not the page file.

### IM-5. Row #19 F6 — "PR #15" parenthetical is still confusing
**Location:** Row 19 + F6 finding.
**Audit refresh wording:** "The 'PR #15' reference in the original cf-icww description is an external/historical issue reference, not a code path on current main."
**Issue:** This is better than the prior "godfrey's PR #15" but still leaves the reader wondering: external to what? If it refers to a third-party tool or an issue tracker, name it. If it was a typo for an internal PR number, say so. The current wording prompts curiosity without satisfying it.
**Suggestion:** Either remove the parenthetical entirely (the audit doesn't depend on it), or link the actual external reference.

### IM-6. Row #9 "duplicate post-purchase queue once cf-jmmk merges"
**Location:** Row 9 notes + F5 finding.
**Issue:** The audit correctly notes that once cf-jmmk merges, both `wixEcom_onOrderCreated` and `wixEcom_onOrderDelivered` will queue the post-purchase sequence. But it gives this only one sentence and buries it in Row 9 / F5 prose. This is arguably a P0 of its own (duplicate emails to every customer post-cf-jmmk-merge). It deserves its own labeled finding (F8?) so the merge gate doesn't lose it.
**Suggestion:** Promote to a labeled finding. Otherwise reviewers will skim and miss it.

---

## Recommended removals

### RM-1. (none — every section earns its place after the refresh)

---

## Prior 5-agent (miquella) findings — addressed status

**Addressed in 47102a81:**
- comment-analyzer: line numbers replaced with symbol anchors throughout — verified (e.g. F1 now cites `newsletterService.web.js#captureExitIntentEmail` not line 487).
- comment-analyzer: wiring contract URL cited — addressed in form, but see IM-3 (citation does not back the precise wording).
- code-reviewer: F5 status clarified with §"Status delta" — addressed correctly. F5 is genuinely still ❌ on `main` (`git show main:src/backend/events.js | grep wixEcom_onOrderDelivered` returns no matches; PR #1141 is OPEN, not merged).
- silent-failure-hunter: Row 24 demoted ✅→⚠️ — verified in diff.
- silent-failure-hunter: Row 31 (content schedule) removed — verified.
- silent-failure-hunter: Row 14 / Row 16 templateIds — addressed for Row 16 (`wishlist_price_drop` is correct, anchor verifies). **Row 14 is wrong** — see CR-1 above. The refresh substituted plausible-looking but invented IDs.
- pr-test-analyzer: P0/P1 test scope explicitly stated — addressed.
- pr-test-analyzer: Step 5 rewritten — addressed.
- pr-test-analyzer: HTML/{{var}}/UTM/throttling steps — addressed via the verification-gaps table.
- pr-test-analyzer: cart-abandonment "wait 1 hr" replaced with fast-forward knobs — addressed.
- code-simplifier: F5 self-contradiction — addressed.
- code-simplifier: dedup of "Live render verification flagged" — addressed.
- code-simplifier: F6 "PR #15" — partially addressed; see IM-5.

**NOT addressed (or addressed but introducing new error):**
- Row 14 templateIds — the refresh "fixed" silent-failure-hunter's complaint by inventing IDs that do not exist in the codebase. This is worse than the original opaque "lifecycle templates" wording because it looks authoritative.
- comment-analyzer central-thesis citation (IM-3) — URL added but does not back the quote.
- pr-test-analyzer 15+ rows without acceptance test — refresh sidestepped by scoping the test plan to P0/P1 only and tracking the rest as "cf-icww.followup-tests." This is a defensible scoping choice, but the unverified ✅ rows are still asserted in the table; future readers may treat ✅ as runtime-asserted unless they read the scope footnote carefully.

---

## Verdict

**Refresh substantially addressed the prior round.** Symbol anchors, the §"Status delta" PR-flip table, and the verification-gaps table are real improvements that protect the audit against rot.

**One blocker** (CR-1) and **one architectural concern** (IM-6) before merge:
1. Row 14 winback claims must be corrected — anchor file is wrong, template IDs are invented.
2. Promote the post-cf-jmmk-merge double-queue concern to a labeled finding so the merge gate doesn't lose it.

The remaining IM-* items are improvements, not blockers.

DreadPirateRobertz (commented):

## Prior 5-agent feedback (miquella): PARTIALLY addressed

| Miquella finding | Status |
|---|---|
| Line numbers → symbol anchors | ADDRESSED |
| Status delta vs. PRs in flight section | ADDRESSED |
| Wix docs citation for wiring contract | ADDRESSED |
| F5 scoped to "on main at audit time"; PR #1141 still open | ADDRESSED — claim accurate |
| Rows 14/16 template IDs and fix paths | ADDRESSED |
| Row 24 Klaviyo changed from ✅ to ⚠️ | ADDRESSED |
| Test plan expanded (HTML/UTM/throttling) | ADDRESSED |
| Duplicate .web.js handler problem | PARTIAL — citation added, but the active duplicate handler block is not called out as dead code that must be removed |

## New findings from independent codebase pass

**Confidence 90 — Duplicate event handler exports not flagged in remediation guidance**

`src/backend/emailAutomation.web.js` exports three Wix platform handlers that are also in `events.js`: `wixMembers_onMemberCreated` (l.167), `wixEcom_onOrderCreated` (l.183), `wixEcom_onOrderCanceled` (l.309). Per Wix Velo docs, only `events.js` exports register. The entire event handler block of `emailAutomation.web.js` (~lines 160–320) is dormant legacy code. The audit correctly identifies that F3/F4/F5 handlers are stranded in `.web.js`, but the remediation guidance does not say these dead exports must be deleted. Any engineer who reads "fix F3 by wiring the order-confirmation handler" may edit `emailAutomation.web.js::wixEcom_onOrderCreated` (which calls `sendOrderConfirmation`) instead of `events.js::wixEcom_onOrderCreated` (which does not). The fix documentation should state explicitly: add `sendOrderConfirmation()` call inside `events.js::wixEcom_onOrderCreated`, and delete the dead event handler block from `emailAutomation.web.js`.

**Confidence 85 — Double-queue secondary concern is misdescribed**

The audit says the post-purchase sequence "will double-queue once F5 merges." The actual risk is: `events.js::wixEcom_onOrderCreated` already queues the sequence anchored to order date; `wixEcom_onOrderDelivered` (F5) will queue it anchored to delivery date. These are not exact duplicates, but they produce overlapping sends for the same customer. Before F5 merges, verify `triggerPostPurchaseSequence` deduplicates on `(recipientEmail, sequenceType, orderNumber)` — not just `(recipientEmail, sequenceType)` — or customers will receive duplicate Day-3/7/30 emails for the same order.

**Confidence 80 — F3 claim fully verified accurate; fix path needs specificity**

Confirmed: `events.js::wixEcom_onOrderCreated` calls `triggerPostPurchaseSequence` and the mobile push webhook, but NOT `sendOrderConfirmation`. The order confirmation email is only in the dormant `emailAutomation.web.js` handler. F3 is correct. The remediation note should name the exact call site to add.

## Merge readiness

**(a) Prior 5-agent feedback: PARTIALLY addressed.** All miquella findings resolved except the duplicate handler dead-code problem is noted but not carried into the remediation guidance.

**(b) NOT merge-ready.** One gate remains: the F3/F4/F5 remediation sections need a sentence clarifying that the `emailAutomation.web.js` event handler block is dead code to be removed, not merely "moved." Without this, whoever implements the fixes is likely to edit the wrong file. A one-line addition per finding (e.g., "fix: add call in `events.js::wixEcom_onOrderCreated`; delete dormant block in `emailAutomation.web.js`") is sufficient to unblock merge.

PR #1147, opened 2026-05-05
fix(cf-89xn): get_deliveryZone — don't spread service error envelope into 200 (cf-tvbi lying-status)

DreadPirateRobertz:
## Summary

Per cf-89xn (radahn's 5-agent review on PR #18 of stage3-velo). The webMethod `getDeliveryZone` has an input-validation branch returning `{ error: '...' }`. The wrapper's `ok({ body: JSON.stringify({ success: true, ...result }) })` spreads that into the success envelope — so a service rejection produced HTTP 200 + `{ success: true, error: '...' }`, the exact cf-tvbi lying-status pattern.

**Fix:** when `result.error` is a string, return 400 + `{ success: false, error: result.error }` so cfw can branch on status. Successful zone shapes (zone/label/rate/eta/distanceMiles, plus the `outofrange` zone with its `message` field) keep producing 200 + `success:true`.

## Reachability

The wrapper already validates `zip` format before calling the service (`/^\d{5}$/`), so the service's own validation branch is currently unreachable from cfw in production. Defensive handling is warranted because:
- The cfutons monorepo's twin in stage3-velo had the same bug actively flagged in radahn's review.
- Future callers / tests may construct service results directly.
- Cost is one if-block.

## Tests

`tests/deliveryZone.test.js` — added 1 case asserting the service `{ error }` envelope path returns 400 + `success:false`. 12/12 pass locally.

## Companion PR
- stage3-velo (paired): `fix/cf-89xn-stage3-followups` — covers all 5 cf-89xn issues including this same deliveryZone fix. Both must land before next Wix CLI publish.

### DIFF
--- a/src/backend/http-functions.js
+++ b/src/backend/http-functions.js
@@ -3004,7 +3004,12 @@ export async function get_deliveryZone(request) {
     // { error: '...' } on its own input validation, which when spread into a
     // 200 envelope produces { success: true, error: '...' } (cf-tvbi
     // lying-status pattern). Surface as 400 so cfw can branch on status.
-    if (result && typeof result.error === 'string') {
+    // Truthy check rejects empty-string `error` (which falls through to 200
+    // since an empty string isn't a real service rejection signal). Logged
+    // so ops can distinguish wrapper-validation 400s from service-validation
+    // 400s without a stack.
+    if (result && typeof result.error === 'string' && result.error.length > 0) {
+      console.warn('[deliveryZone] service emitted error envelope:', result.error);
       return badRequest({
         body: JSON.stringify({ success: false, error: result.error }),
         headers: JSON_HEADERS,
--- a/src/backend/http-functions.js
+++ b/src/backend/http-functions.js
@@ -3000,6 +3000,16 @@ export async function get_deliveryZone(request) {
   }
   try {
     const result = await _getDeliveryZone(zip);
+    // cf-89xn: don't spread service result blindly — `getDeliveryZone` returns
+    // { error: '...' } on its own input validation, which when spread into a
+    // 200 envelope produces { success: true, error: '...' } (cf-tvbi
+    // lying-status pattern). Surface as 400 so cfw can branch on status.
+    if (result && typeof result.error === 'string') {
+      return badRequest({
+        body: JSON.stringify({ success: false, error: result.error }),
+        headers: JSON_HEADERS,
+      });
+    }
     return ok({ body: JSON.stringify({ success: true, ...result }), headers: JSON_HEADERS });
   } catch (err) {
     console.error('HTTP function error (deliveryZone):', err);
--- a/tests/deliveryZone.test.js
+++ b/tests/deliveryZone.test.js
@@ -91,6 +91,19 @@ describe('get_deliveryZone — error handling', () => {
     const res = await get_deliveryZone(makeRequest({ zip: '28792' }));
     expect(res.status).toBe(500);
   });
+
+  it('returns 400 + success:false when service returns { error } envelope (cf-89xn lying-status)', async () => {
+    // The webMethod has its own input-validation branch that returns
+    // { error: 'Please enter a valid 5-digit zip code.' }. Spreading that
+    // into the 200 envelope used to produce { success: true, error: '...' } —
+    // exact cf-tvbi lying-status pattern. Wrapper now surfaces 400.
+    getDeliveryZone.mockResolvedValue({ error: 'Please enter a valid 5-digit zip code.' });
+    const res = await get_deliveryZone(makeRequest({ zip: '28792' }));
+    expect(res.status).toBe(400);
+    const body = JSON.parse(res.body);
+    expect(body.success).toBe(false);
+    expect(body.error).toMatch(/5-digit zip/i);
+  });
 });
 
 describe('options_deliveryZone — CORS preflight', () => {
--- a/tests/deliveryZone.test.js
+++ b/tests/deliveryZone.test.js
@@ -104,6 +104,40 @@ describe('get_deliveryZone — error handling', () => {
     expect(body.success).toBe(false);
     expect(body.error).toMatch(/5-digit zip/i);
   });
+
+  it('keeps the OUT-OF-RANGE envelope (no `error` field) at 200 + success:true', async () => {
+    // Regression guard for cf-89xn: the lying-status fix narrows the 400
+    // branch to results with a non-empty `error` STRING. Out-of-range
+    // results carry a `message` field but no `error` — those are legitimate
+    // success responses (zip is valid, just not deliverable) and must keep
+    // the 200 + success:true envelope so cfw can render the "out of range"
+    // copy without branching on status.
+    getDeliveryZone.mockResolvedValue({
+      zone: 'outofrange',
+      label: 'Out of Range',
+      rate: null,
+      eta: null,
+      distanceMiles: 120,
+      message: 'We deliver within 60 miles of our Hendersonville, NC showroom.',
+    });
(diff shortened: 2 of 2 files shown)


DreadPirateRobertz:

**Verdict: ship as-is.**

Reviewed the deliveryZone guard:

```js
if (result && typeof result.error === 'string') {
  return badRequest({
    body: JSON.stringify({ success: false, error: result.error }),
    headers: JSON_HEADERS,
  });
}
```

- Three lines, single-purpose, reads top-to-bottom. Extracting a helper (`asLyingStatusError(result)`) would add indirection without reducing complexity at the call site — net negative for clarity.
- `result && typeof result.error === 'string'` is the right shape: covers null/undefined and non-string error fields without overreach.
- Comment block accurately describes the cf-tvbi pattern; future readers won't have to spelunk.
- Test mirrors the production shape (mock returns `{ error }`, asserts 400 + `success:false`).

No high-bar simplifications in scope. The cross-repo `parseJsonBody(request, route)` helper noted on PR #16 is still the only meaningful reuse opportunity — out of scope here, lives as a separate bead.

DreadPirateRobertz:

## Scope
Analyzed the new comment block added to `get_deliveryZone` and the matching test comment.

## Verification

**`// cf-89xn: don't spread service result blindly...`** (src/backend/http-functions.js:3003-3006)
- Rationale is **accurate**: the underlying `getDeliveryZone` webMethod does return `{ error: '...' }` on its own input-validation branch (verified). Spreading into `{ success: true, ...result }` does indeed produce the lying-status `{ success: true, error: '...' }` envelope.
- `cf-tvbi` lying-status pattern reference is **valid** — the cf-tvbi bead is the canonical reference for this anti-pattern in the codebase.
- Calls out the cfw consumer concern (branching on status) which is the actual "why."

**Test comment** (tests/deliveryZone.test.js:96-99)
- Accurately describes the historical bug and the fix. Solid regression-test rationale.

## Findings

### Critical Issues
None.

### Improvement Opportunities
- **src/backend/http-functions.js:3003** — Comment says "input validation" but doesn't enumerate which validation paths produce the envelope. A future reader has to chase `_getDeliveryZone` to discover this. One short clause naming the trigger (e.g. "non-NC zip / unmapped county") would make the comment self-contained. Minor.

### Recommended Removals
None.

### Positive Findings
- The comment explains **why** (lying-status anti-pattern, downstream cfw branching) instead of restating the code. That's the kind of comment that earns its rent.
- Bead-ID anchoring (cf-89xn, cf-tvbi) is precise and lets future maintainers reconstruct context.

## Verdict
Comments are accurate, useful, and unlikely to rot. PR is comment-clean.

DreadPirateRobertz:

## Verdict: APPROVE

Tight, targeted fix. Wrapper now distinguishes service-validation envelope from success envelope, eliminating the cf-tvbi lying-status pattern.

### Verified
- **`outofrange` path still returns 200 + success:true** — `_getDeliveryZone` returns `{ zone:'outofrange', label, rate:null, eta:null, distanceMiles, message }` with no `error` field, so the new `typeof result.error === 'string'` guard is correctly skipped (deliveryZoneService.web.js:82-104).
- **Test coverage**: new `tests/deliveryZone.test.js` case asserts 400 + `success:false` + error preservation. Other existing tests (zip-validation, success cases, 500 path) untouched.
- **No spread collision**: by short-circuiting before the `{ success: true, ...result }` spread, downstream callers (cfw `useDeliveryZone`) get a clean status-based branch.

### Notes (non-blocking)
- Twin PR (`stage3-velo` #20) carries the identical fix — verified diffs match.
- Consider a follow-up to make `getDeliveryZone` itself throw on invalid input rather than return `{ error }` — would let the wrapper drop the special-case entirely. Out of scope for cf-89xn.

No high-confidence findings. Ship it.

DreadPirateRobertz:

Scope: `get_deliveryZone` lying-status fix. Net positive — this closes the cf-tvbi pattern. A few sharper edges remain.

**HIGH — empty-string `error` falls through to 200.** `typeof result.error === 'string'` is true for `''` but the spread `{ success: true, ...result }` would emit `{ success: true, error: '' }` — same lying-status shape, just with a falsy error. The current service never produces this, but the whole point of the guard is defensive shape-checking. Recommendation: tighten to `if (result && typeof result.error === 'string' && result.error.length > 0)` and decide explicitly what `error: ''` means (probably treat as 500 + log, since the service would be violating its contract).

**MEDIUM — co-existing valid zone + `error` swallowed as 400.** If service ever returned `{ zone: 'red', label: 'Red Zone', error: 'partial-warning' }` (e.g. distance computed but eta lookup failed), the new branch returns 400 and discards `zone`/`label`. Today's service can't produce this, but the radahn finding that prompted this fix was about defensive shapes. Either narrow the guard to "error AND no zone" (`if (result.error && !result.zone)`) or document in a comment that the contract is mutually exclusive.

**LOW — `outofrange` message is not surfaced via status.** `outofrange` returns `{ zone: 'outofrange', message: '...' }` with no `error`, so it correctly stays 200 — but cfw clients have to inspect `zone` rather than status. That's by design here, just flagging that the new "branch on status" model is partial.

**Logging gap — every input-validation 400 from the service is now silently mapped to 400 with no log.** When this fires in production, ops has no signal that the service-level validator (vs the wrapper-level `/^\d{5}$/` validator above) is being hit. Recommend `console.warn('[deliveryZone] service rejected zip:', zip, result.error)` before the `badRequest` return so misuse patterns surface.

Test suite: the new test covers the happy 400 path. Missing: empty-error path, mixed zone+error path, and a log-emission assertion for the validator branch. Add at least the empty-string test — it's the cheapest hedge against future service-side regressions.

DreadPirateRobertz:

## Summary
The fix is correct and the new test exercises the exact lying-status pattern. 12/12 in `tests/deliveryZone.test.js` pass. One regression-guard gap worth backfilling.

## Important Improvement (rated 6/10)

**Add an OUT-OF-RANGE keep-200 regression test.** The fix uses `typeof result.error === 'string'` to discriminate, which correctly skips the OUT-OF-RANGE branch (which returns `{ zone: 'outofrange', message: '...' }`, no `error` key). But there is no test asserting that. A future refactor could plausibly broaden the guard to `result.zone === 'outofrange'` or `result.message` and silently promote a legitimate \"out of delivery range\" answer to 400 — exactly the kind of regression this PR is hardening against.

Suggested test (mirror of the new one, opposite assertion):
```js
it('returns 200 + zone:outofrange when service returns out-of-range envelope (no error key)', async () => {
  getDeliveryZone.mockResolvedValue({
    zone: 'outofrange', label: 'Out of Range', rate: null, eta: null,
    distanceMiles: 120,
    message: 'We deliver within 60 miles of our Hendersonville, NC showroom. Contact us for a delivery quote.',
  });
  const res = await get_deliveryZone(makeRequest({ zip: '90210' }));
  expect(res.status).toBe(200);
  const body = JSON.parse(res.body);
  expect(body.success).toBe(true);
  expect(body.zone).toBe('outofrange');
  expect(body.message).toMatch(/60 miles/);
});
```

This pins down the contract: \"`error` envelope → 400; `message`-only envelope → 200\". Without it, the keep-200 path is implicitly tested only by the existing `success` cases (which never carry a `message`).

## Positive Observations
- New test is DAMP — name encodes both behavior (\"returns 400 + success:false\") and bead reference (\"cf-89xn lying-status\")
- Body-shape assertions check `success: false` AND error message — would catch a partial fix that returned 400 but kept `success: true`
- Mock setup reuses the exact string the real webMethod returns, so a copy-paste rename in the service would surface
- Existing 11 tests cover happy paths, query validation, and 500 path — good coverage perimeter

PR #1157, opened 2026-05-09
feat(qa): post-publish smoke suite for the 5 Velo /_functions endpoints

DreadPirateRobertz:
## Summary

Per @melania's directive 2026-05-05. Three artifacts:

- **`docs/qa/velo-functions-smoke-2026-05-05.md`** — manual spec with case tables for each endpoint (contactSubmissions, mailingListSignups, sampleRequests, deliveryZone, unsubscribe + bonus health), expected status + body shape, plus a "Known-broken paths" section with an env-var contract for current infra gaps.
- **`scripts/qa/velo-functions-smoke.mjs`** — standalone .mjs runner. Hits staging by default (`VELO_SMOKE_BASE` overrides for prod). 27 cases, TAP-ish output, exit 1 on any unexpected status. Three env vars (`CONTACT_VALID_EXPECT`, `NEWSLETTER_VALID_EXPECT`, `UNSUB_INVALID_TOKEN_EXPECT`) flip expectations as underlying infra fixes ship.
- **`.github/workflows/velo-smoke.yml`** — nightly cron at 09:00 UTC, plus `workflow_dispatch` with overridable inputs and `pull_request` scoped to the spec/script/workflow itself. On schedule/dispatch failure: comments on or creates a single tracking issue (label `velo-smoke`) with a pointer to the env-var contract — so we don't fan out across N runs.

## What the first run surfaced

3 production gaps matching the known-broken contract:

1. **mailingListSignups valid path** returns 400 `'Subscription failed. Please try again.'` — newsletter service unhealthy. Root cause TBD (likely missing CMS collection or ESP creds). Tracked as cf-89xn-followup.
2. **get_unsubscribe + post_unsubscribe invalid-token paths** return 500 — same `SecretNotFoundError` pattern cf-9ieq found for `SITE_OWNER_CONTACT_ID`; `UNSUB_TOKEN_SECRET` is missing in Wix Secrets Manager. The bare-catch in `get_unsubscribe` (silent-failure-hunter flagged on PR #18) hides the actual exception. Worth filing two follow-ups: add the secret, and remove the bare-catch.
3. **contactSubmissions valid path** 500s pending cf-c6g5 (already tracked).

All three are encoded as env-var defaults today so 27/27 pass with current reality. When each fix lands, flip the env var (workflow_dispatch input or commit change) and the smoke starts catching regressions on the new contract.

## Verification

- [x] `node scripts/qa/velo-functions-smoke.mjs` against staging — 27/27 pass.
- [x] Workflow file passes basic shape inspection.
- [ ] First scheduled run will fire at 09:00 UTC tomorrow (or trigger manually via `workflow_dispatch` to validate).

## Follow-ups to file (need melania / prefix)

- Newsletter service `'Subscription failed. Please try again.'` root cause investigation.
- Add `UNSUB_TOKEN_SECRET` to Wix Secrets Manager.
- Remove bare-catch in `get_unsubscribe` (silent-failure pattern, deferred from cf-89xn).
- Once cf-c6g5 lands → flip `CONTACT_VALID_EXPECT` to 200.

### DIFF
--- a/.github/workflows/velo-smoke.yml
+++ b/.github/workflows/velo-smoke.yml
@@ -1,0 +1,103 @@
+name: Velo /_functions smoke
+
+# Hits the 5 Velo /_functions endpoints on staging nightly + on-demand,
+# alerting via a GitHub Issue when expected !== actual status.
+#
+# Spec:   docs/qa/velo-functions-smoke-2026-05-05.md
+# Script: scripts/qa/velo-functions-smoke.mjs
+
+on:
+  schedule:
+    - cron: '0 9 * * *'  # nightly 09:00 UTC
+  workflow_dispatch:
+    inputs:
+      base:
+        description: 'Velo origin (staging by default)'
+        required: false
+        default: 'https://chrisdealglass.wixstudio.com/my-site'
+      contact_valid_expect:
+        description: 'contactSubmissions valid path expected status (500 today; 200 once cf-c6g5 lands)'
+        required: false
+        default: '500'
+      newsletter_valid_expect:
+        description: 'mailingListSignups valid path expected status (400 today; 200 once newsletter healthy)'
+        required: false
+        default: '400'
+      unsub_invalid_expect:
+        description: 'unsubscribe invalid-token paths expected status (500 today; 400 once UNSUB_TOKEN_SECRET added)'
+        required: false
+        default: '500'
+  pull_request:
+    paths:
+      - 'scripts/qa/velo-functions-smoke.mjs'
+      - 'docs/qa/velo-functions-smoke-*.md'
+      - '.github/workflows/velo-smoke.yml'
+
+permissions:
+  contents: read
+  issues: write
+
+jobs:
+  smoke:
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Use Node.js 22
+        uses: actions/setup-node@v4
+        with:
+          node-version: 22
+
+      - name: Run Velo smoke
+        id: smoke
+        env:
+          VELO_SMOKE_BASE: ${{ inputs.base || 'https://chrisdealglass.wixstudio.com/my-site' }}
+          CONTACT_VALID_EXPECT: ${{ inputs.contact_valid_expect || '500' }}
+          NEWSLETTER_VALID_EXPECT: ${{ inputs.newsletter_valid_expect || '400' }}
+          UNSUB_INVALID_TOKEN_EXPECT: ${{ inputs.unsub_invalid_expect || '500' }}
+        run: |
+          set +e
+          node scripts/qa/velo-functions-smoke.mjs | tee smoke.tap
+          echo "exit_code=${PIPESTATUS[0]}" >> "$GITHUB_OUTPUT"
+
+      - name: Surface failures in job summary
+        if: always()
+        run: |
+          {
+            echo "## Velo smoke results"
+            echo
+            echo '```'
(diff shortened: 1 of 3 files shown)


DreadPirateRobertz:
**5-agent aggregate review** (dispatched by godfrey)

Combined review across correctness, silent-failures, simplification, comments, and tests. Verified against staging by running `node scripts/qa/velo-functions-smoke.mjs` — **27/27 green** at the current default expectations (`CONTACT_VALID_EXPECT=500`, `NEWSLETTER_VALID_EXPECT=400`, `UNSUB_INVALID_TOKEN_EXPECT=500`).

### 1. Correctness — PASS
- Script's case set matches the 6 endpoint tables in the spec (health was added as bonus row #6 and is wired in script).
- Honeypot path verified: `subscribeToNewsletter` returns `{ success: true, discountCode }` when `options.honeypot` truthy (newsletterService.web.js:341-344), and `post_mailingListSignups` translates `success:true` to `ok()` (200). Spec's "honeypot silent 200" is correct.
- All `expectStatus` values align with the canonical wrappers in `src/backend/http-functions.js` (badRequest=400, ok=200, serverError=500, OPTIONS via `corsPreflight` returns 204).
- The 429 fallback on the newsletter valid-path (`expectStatus: [NEWSLETTER_VALID_EXPECT, 429]`) is a thoughtful guard against per-email rate-limit collisions on repeat smoke runs.

### 2. Silent failures — one minor finding
- **`Number('abc')` → NaN, and the body-match ternary `UNSUB_INVALID_TOKEN_EXPECT === 400 ? /…/ : /…/`** silently picks the 500-side regex on a typo'd env var. Status comparison is fail-closed (NaN never matches any real status), so the case fails loudly — but the body-match branch silently misroutes. Low severity: the case will fail on status anyway, so the misrouted regex never gets a chance to mask anything. Optional hardening: `Number.isFinite(parsed) ? parsed : THROW`.
- `gh issue create … || gh issue create …` fallback: the second invocation runs under `set -euo pipefail` so a true failure exits non-zero and the alert step itself fails the run — acceptable.
- The `gh issue list` line uses `|| echo ""` which swallows auth failures and silently creates a duplicate issue rather than commenting. Acceptable for a nightly alert (better to over-notify than silently lose).

### 3. Simplification — none required
27 cases-as-data is the right shape. Extracting a per-endpoint helper would obscure more than it would save; the `probe()` abstraction already does the heavy lifting.

### 4. Comments / spec doc — accurate
- Bead IDs cf-foo0, cf-w1lg, cf-9ieq, cf-c6g5, cf-fuvd cited in the doc all exist (cf-foo0/cf-w1lg/cf-89xn closed; cf-9ieq/cf-c6g5/cf-fuvd open per linked-beads section, consistent with current state).
- "First run surfaced 3 production gaps" still accurate — running today against staging with the documented defaults (500/400/500) is 27/27 green, confirming the 3 gaps are still present and the contract document reflects reality.

### 5. Tests — appropriately omitted
The script *is* the test. A meta-test asserting the case array length would be theater; the live run is the only meaningful verification, and the workflow runs it nightly + on-PR-touching-itself.

### Hook-aware GHA security — PASS
- `inputs.*` flow through `env:` block, never inlined into `run:` shell strings.
- `github.event_name`, `github.run_id`, `github.repository`, `github.server_url`, `secrets.GITHUB_TOKEN` are all constrained/trusted values.
- No `${{ github.event.pull_request.title }}` or other user-controllable interpolation in `run:` blocks.
- `permissions:` is correctly minimized (`contents: read`, `issues: write`).

### Verdict: APPROVE
Solid net-new QA infrastructure. One nit (NaN env-var hardening) is optional — does not block merge.

PR #1164, opened 2026-05-09
proposal(cf-vtx5): module-dispatcher pattern for cfw→Velo routing

DreadPirateRobertz:
## Summary
Pairing with godfrey on cf-vtx5 (P0 — 22 cfw→Velo routes 404'ing). I wrote the cf-jqkg audit; godfrey owns the rollout. Posting design proposal + reference implementation so godfrey can pick the pattern and execute.

### Two patterns considered
- **A** — 22 individual `post_<name>` handlers (cf-w1lg shape). Requires cfw to refactor `callVelo` to drop the module prefix; method-name collisions (e.g. `getLeaderboard` exists in both `gamificationCore` and `loyaltyService`) need rename.
- **B** — one dispatcher per module reading `request.path[0]` and forwarding `body.args` into the matching webMethod. No cfw change. Cuts wrapper count from 22 to 9 (6 module dispatchers + 3 standalone for the cf-jqkg flat-URL gaps).

**Recommendation: B.** See `docs/cf-vtx5-dispatcher-design.md`.

### Reference implementation: `wishlistService`
- `src/backend/http-functions.js` — `post_wishlistService` + `options_wishlistService` (~80 lines). Per-module allowlist Set names the 5 cfw-callable methods; Admin-only `updateWishlistStock` deliberately omitted as cfw-boundary enforcement.
- `tests/wishlistServiceDispatcher.cfvtx5.test.js` — 9 tests covering happy path, args spread, allowlist rejection, body parse failures, server-error errorId correlation, and pass-through of `{success:false}` envelope as HTTP 200 (reserves 500 for unexpected throws).

### Permission model
webMethods called from backend code bypass Wix's client-side Permissions gating, so each method must self-guard via `currentMember.getMember()` if it requires SiteMember. The allowlist itself doubles as the "what cfw can call" audit surface — exactly what would have prevented cf-jqkg's silent-404 surface in the first place.

## Test plan
- [x] 9/9 wishlistServiceDispatcher tests pass.
- [x] 350/350 broader http test surface clean (regression check).
- [ ] godfrey reviews + picks pattern (A vs B vs hybrid).
- [ ] If B: godfrey replicates the shape across `gamificationCore`, `loyaltyService`, `referralService`, `styleQuiz`, `pushNotificationService`. Then folds in the 3 flat-URL gaps (`recordSpinGrant`, `submitSurvey`, `submitCommunityPhoto`).
- [ ] Stilgar repeats the 4-route smoke probe; expects 200 instead of 404.

### DIFF
--- a/docs/cf-vtx5-dispatcher-design.md
+++ b/docs/cf-vtx5-dispatcher-design.md
@@ -1,0 +1,131 @@
+# cf-vtx5 design proposal — module dispatcher pattern
+
+**Bead:** cf-vtx5
+**Author:** rennala (paired with godfrey)
+**Status:** proposal — godfrey picks pattern, executes the rest
+**Refs:** cf-jqkg audit (PR #1160), godfrey's cf-w1lg (5 wrappers) and cf-foo0 fix patterns
+
+## Problem recap
+
+cfw's `callVelo` posts to `/_functions/<module>/<method>` with body `{args: [...]}`. 22 routes are 404-ing because no Velo `get_<module>` / `post_<module>` handlers exist. Every loyalty / gamification / wishlist / referral / quiz / push call from cfw is currently broken.
+
+Affected modules and method counts:
+
+| Module                    | Methods called from cfw                                         | Webmethod self-guards? |
+|---------------------------|------------------------------------------------------------------|------------------------|
+| `gamificationCore`        | `getActiveChallenges`, `getActivityFeed`, `getLeaderboard`, `receiveGamificationEvent`, `recordChallengeProgress` | yes (most) |
+| `loyaltyService`          | `getLeaderboard`, `getMyActivity`, `getMyLoyaltyAccount`, `redeemReward` | yes |
+| `wishlistService`         | `addToWishlist`, `getWishlist`, `getWishlistByMemberId`           | yes — `currentMember.getMember()` first line |
+| `referralService`         | `getMyReferralCode`, `getReferralByCode`                         | mixed |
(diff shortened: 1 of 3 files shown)


DreadPirateRobertz:
Addressed all 4 review items in d4a2d26b on the design doc:

1. **One-dispatcher-per-module** — confirmed direction; no impl change (Pattern B already shipped).
2. **CORS via `backend/utils/cors.js`** — reference impl already uses `corsHeaders` + `corsPreflight`; design doc now makes it a hard requirement so godfrey doesn't hand-roll Access-Control-Allow-Origin in any new dispatcher.
3. **Field-name shim pattern** — worked example for `post_submitSurvey` showing cfw `{score, comments, orderId}` → `submitSurveyResponse({orderId, npsScore, comment})`. Each shim line gets a `// cfw "<field>" → Velo "<field>"` comment for auditability. If a shim grows past 5 lines, escalate to fix cfw payload at source.
4. **Auth gating** — three-class table (authenticated / anonymous-rate-limited / mixed) by cfw call shape, with the dispatcher's duty for each. Spot-check discipline documented for adding methods to a new module's allowlist (grep for `currentMember.getMember()` or confirm intentional Anyone).

   Bonus gotcha flagged: cfw's `survey.ts` currently fires WITHOUT forwarding the access token, so the `submitSurvey` wrapper would return "Authentication required" even after wiring. cfw-side forward-the-token fix recommended; flagged for the cf-vtx5 follow-up so the wrapper isn't shipped half-broken.

Reference implementation (wishlistService dispatcher + 9 tests) unchanged — already aligned with all four addenda. Ready for radahn's 5-agent pass.

PR #1193, opened 2026-05-10
chore(cf-kg1x): delete dead fabricSampleService + FabricSampleRequest pair

DreadPirateRobertz:
## Summary
- Deletes \`src/backend/fabricSampleService.web.js\`, \`src/public/FabricSampleRequest.js\`, and their two test files
- Updates \`docs/cf-66ne-phase-b-audit-2026-05-09.md\` with the verdict

## Why (cf-kg1x decision)
Per the cf-66ne Phase B audit, \`FabricSampleRequest.js\` imports two symbols (\`getAvailableSwatches\`, \`submitFabricSample\`) that do not exist in \`fabricSampleService.web.js\` — the backend exports only \`submitFabricSampleRequest\`. Any caller would throw.

cf-kg1x decision criteria: \`grep -rn 'FabricSampleRequest|initFabricSampleRequest|fabricSampleSection' src/pages/\` returned **zero callers**. The live swatch flow is \`Fabric Swatches.js\` → \`swatchRequest.web.submitSwatchRequest\` (cf-w1lg PR #15), entirely independent. Both files were dead-on-arrival.

**Verdict: Option C — delete.**

## Verification
- \`grep -rn 'fabricSampleService|FabricSampleRequest|getAvailableSwatches|submitFabricSample' src/ tests/\` post-delete = **0 results**
- Doc references in \`docs/cf-66ne-phase-b-audit-2026-05-09.md\` retained as historical context (the verdict + action are now recorded in the §1 entry)
- Full vitest sweep: 40,253 pass / 1 unrelated pre-existing failure (\`tests/noDoMockInTestBody.test.js\` against \`communityPhoto.test.js\` — not introduced by this PR)

## Refs
Bead: cf-kg1x. Discovered from cf-66ne Phase B audit (PR #1190).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1196, opened 2026-05-10
fix(cf-unxw): fabricSampleService — getAvailableSwatches + submitFabricSample alias

DreadPirateRobertz:
## Summary

Per @morgott's cf-66ne Phase A chunk 2 audit + cf-unxw bead. `src/public/FabricSampleRequest.js` (Velo widget) imports `getAvailableSwatches` and `submitFabricSample` from `backend/fabricSampleService.web`; backend originally only exported `submitFabricSampleRequest` and had **no swatch-list method at all**. Production widget was silently broken — modal opens but the swatch repeater never populates, and submit calls undefined → caught by FE try/catch → "Something went wrong".

Same pattern class as cf-hpb2. Stilgar precedent is to add backend aliases rather than touch the Wix Velo Editor.

### (1) `getAvailableSwatches` webMethod

Queries `FabricSwatches` CMS (sortOrder ASC, limit 100), returns the FE-shape envelope:

```js
{ success: true, data: { swatches: [{ _id, swatchId, swatchName, swatchImage, colorFamily, colorHex, material }] } }
```

Coerces legacy `name` field → `swatchName` for older CMS rows. Failure returns `{success: false, error}` so the FE disables the request button.

### (2) `submitFabricSample = submitFabricSampleRequest` alias

Backend-side alias so the FE `submitFabricSample(...)` import resolves. Same wrapper, same `Permissions.Anyone` behavior. Test asserts `submitFabricSample === submitFabricSampleRequest` (single wrapper, two names).

## Known follow-up (not blocking — file as cf-unxw.fu1)

The FE call shape `submitFabricSample({swatchIds, contactInfo: {name, email, address}, productId})` doesn't match the backend's expected `{contactInfo: {firstName, lastName, email, address1, city, state, zip}, productSlug}`. Aliasing unblocks the **import resolution** failure; the **shape mismatch** still makes submit reject with a validation error. Either (a) FE rewrite to match canonical, or (b) backend shape-shimming wrapper that maps flat → structured. Documented in the alias's JSDoc.

## Tests

5 new cases in `tests/fabricSampleService.test.js`:
- `getAvailableSwatches`: happy path, legacy `name` → `swatchName` coercion, empty CMS still returns `success:true`
- `submitFabricSample`: alias is the same wrapper as canonical (`===`), routes through to canonical impl

**66/66 pass in the file.**

## Companion PR

`carolina-futons-stage3-velo` — mirrors the backend file change (no test duplication per stage3's single-source posture).

Closes cf-unxw.

### DIFF
--- a/src/backend/fabricSampleService.web.js
+++ b/src/backend/fabricSampleService.web.js
@@ -260,3 +260,65 @@ export const submitFabricSampleRequest = webMethod(
     }
   }
 );
+
+// ── cf-unxw — FE/BE name reconciliation ──────────────────────────────────────
+//
+// src/public/FabricSampleRequest.js imports `getAvailableSwatches` and
+// `submitFabricSample` from this module, but the backend originally only
+// exported `submitFabricSampleRequest` and had no swatch-list method.
+// Production Velo widget was silently broken (modal opens but swatches
+// never render; submit throws). Same pattern class as cf-hpb2 referralService
+// rename — Stilgar precedent is to add backend aliases rather than touch
+// Wix Velo Editor pages.
+
+/**
+ * List swatches available for the customer-facing fabric sample request
+ * widget. Returns the catalog ordered by `sortOrder` for stable display.
+ *
+ * @function getAvailableSwatches
+ * @returns {Promise<{success: boolean, data?: {swatches: Array}, error?: string}>}
+ *   On success: `{success:true, data:{swatches:[…]}}` with each swatch
+ *   carrying `_id`, `swatchId`, `swatchName`, `swatchImage`, `colorFamily`,
+ *   `colorHex`, `material`.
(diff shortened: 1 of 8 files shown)


DreadPirateRobertz:
Superseded by cf-kg1x PR #1193 (e91eb97b) — that PR deleted the fabricSampleService backend + FabricSampleRequest public widget pair entirely (Option C from the cf-kg1x bead), which makes this PR's Option-A 'add aliases' approach moot for cfutons.

Note: stage3-velo PR #33 (mirror) already merged BEFORE this conflict surfaced, so stage3 still has the fabricSampleService files + my new aliases. That's drift — if cfutons treats the subsystem as dead, stage3 should follow with a deletion. Flagging for melania.

cf-unxw bead can close (the FE/BE name mismatch is no longer a real concern since the FE module itself is gone in cfutons).

PR #1216, opened 2026-05-10
docs(cf-3qt.9): consolidated Wix Studio retirement checklist

DreadPirateRobertz:
## Summary

Phase 9 pre-work doc for the cf-3qt epic — the canonical retirement checklist for the Wix Studio cutover, gated on cf-3qt.8 (DNS) + 24 h prod stability + Stilgar sign-off.

Consolidates two closed sibling research docs from morgott (cf-3qt.9.1 + cf-3qt.9.2) into a single document at `docs/cf-3qt-9-wix-retirement-checklist.md` covering all four areas in melania's nudge:

1. Production carolinafutons.com unpublish (§1) — pre-flight, UI steps, risks, rollback
2. Services that stay active vs retire (§3) — Wix Stores / Headless API / CMS / Velo all stay; rendering layer + editor go
3. Wix Premium downgrade evaluation (§4) — Wix Headless plan ~$13/mo vs assumed Business Unlimited ~$25/mo (~$144/yr savings, pricing re-verification flagged)
4. EDITOR_HOOKUP_GUIDE archive (§5) — file moves + RETIRED banner template

Plus staging-site (§2), Stilgar sign-off gate (§6), execution log (§7), post-unpublish smoke-test playbook (§8), and linked-bead index (§9).

**Doc only — no live changes. No execution until §6 fully signed.**

## Test plan
- [x] All four melania-nudged areas covered
- [x] References morgott's prior research (cf-3qt.9.1 + cf-3qt.9.2)
- [x] No execution path can run before Stilgar sign-off in §6
- [x] Smoke-test playbook covers Stores API + Velo HTTP + cfw routes + order placement
- [x] Cross-reference to cf-xe2 prevents future-scope creep

### DIFF
--- a/docs/cf-3qt-9-wix-retirement-checklist.md
+++ b/docs/cf-3qt-9-wix-retirement-checklist.md
@@ -1,0 +1,348 @@
+# cf-3qt.9 — Wix Studio Retirement Checklist
+
+**Phase:** cf-3qt.9 (final phase of the cf-3qt epic — full migration to Next.js + Wix Headless)
+**Status:** Pre-work — DOCUMENTATION ONLY. Execution gated on cf-3qt.8 (DNS cutover) + 24 h prod stability.
+**Audience:** Stilgar (sign-off), melania (PM), engineering (execution)
+**Predecessors:** This consolidates the retirement research from cf-3qt.9.1 (`crew/melania/retirement-plan.md` — production focus) and cf-3qt.9.2 (`crew/melania/wix-retirement-checklist.md` — staging focus). Both authored by morgott on 2026-05-04.
+
+> ⚠️ **Do NOT execute any step in this document until:**
+> 1. cf-3qt.8 (DNS cutover) is complete AND `carolinafutons.com` has served from Vercel for ≥ 24 h with zero P0/P1 rollback triggers, AND
+> 2. Stilgar has signed every box in §6 below.
+
+---
+
+## 0. Scope at a glance
+
+| What gets retired | What stays active |
+|---|---|
+| Wix Studio rendering layer (the page editor + published front-end) | Wix Stores (catalog, inventory, orders, payments, fulfillment) |
+| `EDITOR_HOOKUP_GUIDE.html` as the live developer reference | Wix Headless APIs (`@wix/stores`, `@wix/auto_sdk_*`) consumed by Next.js |
+| The `cf-yw0m` editor-hookup standing order | Wix CMS collections (read-only data source for cfw) |
+| Free-tier staging site `My Site 5` (metaSiteId `3af610bf`) | Velo backend (`src/backend/**`) — webhooks, cron, custom APIs |
+| | Production carolinafutons.com (DNS already pointing at Vercel) |
+
+The migration retires the *front-end rendering* and *page editor* — not Wix as a backend.
+
+---
+
+## 1. Production Wix Studio Unpublish (carolinafutons.com)
+
+> Detail: `crew/melania/retirement-plan.md §1`. Summary follows.
+
+### 1.1 Pre-flight checks
+
+Execute every check **before** clicking Unpublish:
+
+- [ ] `curl -I https://carolinafutons.com` → header `server: Vercel` (NOT Wix)
+- [ ] `curl https://carolinafutons.com/api/healthcheck` → `200 OK`
+- [ ] Vercel dashboard: production deploy stable for ≥ 24 h, no traffic anomalies
+- [ ] Wix dashboard → **Analytics** → confirm zero organic traffic on Wix-rendered URLs in the last 24 h
+- [ ] No active Wix Studio A/B tests, no scheduled publishes
+- [ ] Sentry P0/P1 rate at baseline for ≥ 24 h post-cutover
+
+### 1.2 Unpublish sequence (UI)
+
+1. Log in at **wix.com** with the Carolina Futons owner account.
+2. Open the **Carolina Futons** site dashboard.
+3. **Site Actions** → **Manage Site** (or **Settings** depending on plan UI).
+4. **Publishing** → **Unpublish Site** → confirm.
+5. Verify `https://carolinafutons.com` still serves the Next.js storefront via Vercel.
+6. Verify Wix dashboard → **Store Products** still loads (API still functional).
+7. Record timestamp in §7 Execution Log.
+
+### 1.3 What unpublish does (and doesn't)
+
+| Effect | After unpublish |
+|---|---|
+| Wix Studio page rendering | ❌ Disabled |
+| Wix Stores REST/SDK API | ✅ Still live |
+| Velo backend (`src/backend/**`) | ✅ Still live |
+| Custom domain `carolinafutons.com` | ✅ Unaffected — DNS routes to Vercel |
+| Wix dashboard / order management | ✅ Still functional |
+
+### 1.4 Risks + mitigations
+
+| Risk | Likelihood | Mitigation |
+|---|---|---|
+| Wix Stores API breaks because rendering is off | Low | Stores API is plan-tier, not render-tier. Verified per Wix docs. Smoke-test §8. |
+| Velo HTTP functions stop firing | Low | Velo is plan-tier, not render-tier. Test in §8. |
+| Wix-hosted checkout page breaks | Low | Checkout is backend; not tied to Wix Studio rendering. |
+| SEO regression — Google still has Wix-rendered URLs in index | Medium | Vercel pages must be in Search Console coverage report ≥ 24 h pre-unpublish. |
+| Email triggers (price-match, Q&A) stop reaching ops | Low | Webhooks live in Velo; un-coupled from rendering. Verify §8. |
+| Wix dashboard login breaks | None | Dashboard is decoupled from site publish status. |
+
+### 1.5 Rollback
+
+Rollback is one click: **Wix dashboard → Publishing → Publish Site**. No DNS changes needed.
+
+---
+
+## 2. Staging Site Unpublish (`My Site 5`, metaSiteId `3af610bf`)
+
+> Detail: `crew/melania/wix-retirement-checklist.md`.
+
+`My Site 5` is a free-plan dev sandbox under the halworker85 account at
+`https://chrisdealglass.wixstudio.com/my-site`. **Not** the same site as production.
+
+### 2.1 Pre-flight
+
+- [ ] Confirm `My Site 5` has no active integrations expecting it to be published
+- [ ] Confirm no Velo webhooks under `My Site 5` receive production traffic
+
+### 2.2 Unpublish (UI Path A — recommended)
+
+1. `https://manage.wix.com/dashboard/3af610bf/home`
+2. Sidebar → **Site & Mobile App** → **Site Actions** → **Unpublish Site**
+3. Confirm dialog
+4. Verify `chrisdealglass.wixstudio.com/my-site` returns the "site is not published" page
+
+### 2.3 Notes
+
+- No Stilgar sign-off needed for `My Site 5` — it's a staging/dev site with no production traffic.
+- No subscription action: it's already on the free tier.
+- The production STAGING_SITE used for editor work pre-migration is on free/Lite (Upgrade CTA visible in dashboard) — same treatment applies.
+
+---
+
+## 3. What Wix services stay active post-cutover
+
+> Required by Phase 9 to ensure no service we depend on gets dropped.
+
+### 3.1 Active dependencies
+
+| Service | What we use it for | Where it's wired |
+|---|---|---|
+| **Wix Stores** | Product catalog, variants, inventory, orders, payment processing, fulfillment records | cfw `actions/*` calling Wix Headless SDK |
+| **Wix Headless API (`@wix/stores`, `@wix/auto_sdk_stores_*`)** | Cart, checkout, member sessions, redirect-to-checkout flow | cfw `lib/wix-client.ts` + `actions/cart.ts` |
+| **Wix CMS collections** | Static-ish content (Blog, FAQs, ContactSubmissions, NewsletterSubscribers, etc.) | cfw `lib/wix-cms.ts`; Velo `*.web.js` for writes |
+| **Velo HTTP functions (`src/backend/http-functions.js`)** | Server-side endpoints for webhooks, rate-limited writes (contact form, swatch request, etc.), price-match orchestrator | cfw calls via `/_functions/*` |
+| **Velo events (`src/backend/events.js`, `*.events.js`)** | `wixEcom_onOrderCreated`, `wixEcom_onOrderApproved`, `wixEcom_onOrderFulfilled`, `wixEcom_onFulfillmentCreated` (post cf-fovb) — drive the email/SMS notification orchestrator | Velo runtime |
+| **Wix Secrets Manager** | WWEX SOAP credentials, Turnstile keys, mailer creds | Velo runtime via `wix-secrets-backend` |
+| **Wix dashboard** | Manual order management, customer support, Stores admin | Stilgar / ops |
+
+### 3.2 Services we explicitly stop using
+
+| Service | Replacement |
+|---|---|
+| Wix Studio page editor | n/a — Next.js source in `carolina-futons-web` repo |
+| Wix Studio rendering layer (HTML/CSS/JS bundle Wix shipped to browsers) | Vercel-hosted Next.js |
+| `EDITOR_HOOKUP_GUIDE.html` as the live reference | `docs/SITE-OWNER-GUIDE.md` + `docs/TESTING-GUIDE.md` |
+| Wix Studio analytics dashboard | Vercel Analytics + custom events |
+| Wix Studio A/B testing tool | Application-layer feature flags / experiments |
+
+### 3.3 Future scope (NOT in cf-3qt.9)
+
+Tracked separately in **cf-xe2** (full Wix exit — Velo → Vercel Functions, Stores → headless commerce, CMS → Sanity/Payload). Do not pull forward into Phase 9.
+
+---
+
+## 4. Wix Premium plan downgrade evaluation
+
+> Detail: `crew/melania/retirement-plan.md §3`.
+
+### 4.1 Why downgrade
+
+Once Wix Studio rendering is unpublished, the only Wix infrastructure we still use is:
+- Wix Stores API (catalog + checkout backend)
+- Wix CMS API (read-only data source)
+- Velo backend execution (webhooks, cron, custom APIs)
+
+No part of that requires the rendering tier.
+
+### 4.2 Plan options (verify current pricing at wix.com/upgrade)
+
+> ⚠️ Pricing last verified ~August 2025 per morgott's research. **Re-verify current tiers at wix.com/upgrade before committing.**
+
+| Plan | ~Monthly (annual) | Wix Stores API | Velo | Wix Studio Rendering | Notes |
+|---|---|---|---|---|---|
+| **Wix Headless** | ~$13 | ✅ | ✅ | ❌ (we don't need it) | **Recommended** — purpose-built for our use case |
+| Business Basic | ~$17 | ✅ | ✅ | ✅ (unused) | Pays for rendering we won't use |
+| Business Unlimited | ~$25 | ✅ | ✅ | ✅ (unused) | Likely current plan |
+| Business VIP | ~$35 | ✅ | ✅ | ✅ (unused) | Overprovisioned |
+
+### 4.3 Estimated cost delta
+
+| Scenario | ~Monthly (annual) |
+|---|---|
+| Current (assumed Business Unlimited) | $25 |
+| Target (Wix Headless) | $13 |
+| **Estimated savings** | **~$12/mo (~$144/yr)** |
+
+Exact savings depend on the actual current plan and whether legacy/promotional pricing applies. Have Stilgar pull the current subscription cost from **Wix Dashboard → Billing & Subscriptions** before committing.
+
+### 4.4 Downgrade path (UI)
+
+1. Wix Dashboard → **Billing & Subscriptions**
+2. Identify the current active plan (record exact name + monthly cost in §7)
+3. **Change Plan** → select **Wix Headless**
+4. Wix may prompt for cancellation confirmation; review what the downgrade removes (rendering tier features) — for us, nothing required
+5. Confirm
+6. Record in §7 Execution Log
+
+### 4.5 Downgrade risks
+
+| Risk | Likelihood | Mitigation |
+|---|---|---|
+| Headless plan removes a feature we still rely on | Low | Verify Wix Headless plan page lists Stores API + Velo before committing |
+| Velo execution quotas lower on Headless | Medium | Pull current Velo quota usage. Compare to Headless quota. Test in staging if uncertain. |
+| Wix Stores checkout flow breaks | Low | Checkout is backend-only and plan-tier-agnostic. |
+| Webhook delivery breaks | Low | Velo is included in Headless; smoke-test §8. |
+| Pricing changes by the time we execute | Medium | Re-verify at wix.com/upgrade before clicking |
+
+### 4.6 Stilgar approval gate
+
+Downgrade does NOT execute until Stilgar (a) confirms current plan + cost, (b) reviews the Wix Headless feature page, (c) signs §6.
+
+---
+
+## 5. Guide archival
+
+> Detail: `crew/melania/retirement-plan.md §2`.
+
+### 5.1 Files to archive
+
+| Source path | Archive destination |
+|---|---|
+| `cfutons/EDITOR_HOOKUP_GUIDE.html` | `docs/archive/EDITOR_HOOKUP_GUIDE.html` |
+| `docs/EDITOR-HOOKUP-GUIDE.md` (if present) | `docs/archive/EDITOR-HOOKUP-GUIDE.md` |
+
+### 5.2 Archive steps (executed by dev — single PR)
+
+```bash
+mkdir -p docs/archive
+
+# Header banner: prepend a RETIRED notice to each file before moving.
+# (HTML — insert above <body>, MD — top-of-file blockquote.)
+
+git mv cfutons/EDITOR_HOOKUP_GUIDE.html docs/archive/EDITOR_HOOKUP_GUIDE.html
+git mv docs/EDITOR-HOOKUP-GUIDE.md docs/archive/EDITOR-HOOKUP-GUIDE.md  # if present
+
+git commit -m "chore(cf-3qt.9): archive EDITOR_HOOKUP_GUIDE — Wix Studio retired (see cf-3qt)"
+```
+
+### 5.3 Banner template
+
+Markdown header to prepend to each archived doc:
+
+```markdown
+> **🚫 RETIRED — DO NOT USE FOR NEW WORK**
+>
+> This document described the Wix Studio editor hookup pattern, which was retired
+> as part of [cf-3qt](../bd/cf-3qt) (full migration to Next.js + Wix Headless).
+> Archived on YYYY-MM-DD for historical reference. The current developer reference
+> is [SITE-OWNER-GUIDE.md](../SITE-OWNER-GUIDE.md) +
+> [TESTING-GUIDE.md](../TESTING-GUIDE.md).
+```
+
+HTML equivalent (top of `<body>`):
+
+```html
+<div style="background:#fee;border:2px solid #c33;padding:1em;margin-bottom:1em;font-weight:bold">
+  🚫 RETIRED — Do not use for new work. Wix Studio was retired as part of cf-3qt.
+  Archived YYYY-MM-DD. Current reference: docs/SITE-OWNER-GUIDE.md.
+</div>
+```
+
+### 5.4 Crew worktree copies
+
+Copies of `EDITOR_HOOKUP_GUIDE.html` exist in some `crew/*/` dirs. These are crew-local files, not in the published doc tree, and can be left in place or deleted at crew discretion.
+
+### 5.5 Standing-order removal
+
+After archive lands:
+
+- [ ] Remove `cf-yw0m` (editor hookup standing order) from melania's auto-memory
+- [ ] Remove related instructions from any active `CLAUDE.md` / `AGENTS.md` files
+- [ ] Mail godfrey + remaining crew about the retirement
+
+---
+
+## 6. Stilgar sign-off
+
+**Instructions:** Each box must be checked before any §1, §4, or §5 step executes.
+
+### 6.1 Pre-conditions (gate execution)
+
+- [ ] **DNS stable:** `carolinafutons.com` served from Vercel for ≥ 24 h with no errors
+- [ ] **cf-3qt.8 complete:** Phase 8 PRs all merged + confirmed live
+- [ ] **30-day stability report green:** uptime, order rate, performance, P0/P1 incident review (cf-3qt.9 acceptance gate)
+- [ ] **Order history confirmed durable:** Verified in Wix Dashboard, NOT solely in the Studio rendering layer
+- [ ] **SEO baseline captured:** Search Console shows Vercel pages indexed; no significant Wix-only URLs in coverage
+
+### 6.2 Per-step approvals
+
+- [ ] **Stilgar approves Wix Studio production unpublish** (§1)
+- [ ] **Stilgar approves staging site unpublish** (§2) — *can be self-approved by ops if not gated*
+- [ ] **Stilgar approves EDITOR_HOOKUP_GUIDE archive** (§5) — or requests a copy emailed first
+- [ ] **Stilgar confirms current Wix plan + monthly cost** (§4.3 — pulled from Billing dashboard)
+- [ ] **Stilgar approves Wix Headless downgrade** (§4) — or chooses to defer to a separate decision
+
+### 6.3 Post-execution sign-offs
+
+Run **after** unpublishing + downgrading. Each must pass.
+
+- [ ] Wix Stores API smoke test passes (§8 below)
+- [ ] cfw `/shop/futon-frames`, `/shop/mattresses`, `/products/<slug>` all 200 from Vercel
+- [ ] Wix dashboard order management still functional (place test order)
+- [ ] Price-match form submission still reaches email (ContactSubmissions webhook fires)
+- [ ] Q&A submission still reaches Velo (`submitQuestion` webhook fires)
+- [ ] No Sentry P0/P1 spike in 24 h post-unpublish
+- [ ] Tag `v-wix-studio-retired` in both `carolina-futons` and `carolina-futons-web` repos
+- [ ] Retrospective doc opened (will live at `docs/cf-3qt-retrospective.md`)
+
+---
+
+## 7. Execution log
+
+| Date | Action | Executed by | Notes |
+|---|---|---|---|
+| — | Pre-flight checks complete | — | — |
+| — | Stilgar §6.1 sign-off | — | — |
+| — | Stilgar §6.2 sign-off | — | — |
+| — | Production carolinafutons.com unpublished | — | — |
+| — | Staging `My Site 5` unpublished | — | — |
+| — | EDITOR_HOOKUP_GUIDE archived | — | — |
+| — | Wix plan downgraded | — | from `<plan>` ($X) → Wix Headless ($Y), savings $Z/mo |
+| — | Tag `v-wix-studio-retired` pushed | — | both repos |
+| — | Retrospective opened | — | — |
+| — | cf-yw0m standing order removed from memory | — | — |
+
+---
+
+## 8. Smoke-test playbook (post-unpublish)
+
+Run within 1 h of unpublishing. Each must pass.
+
+```bash
+# 1. Vercel storefront still serving
+curl -I https://carolinafutons.com                                     # 200, server: Vercel
+curl -I https://carolinafutons.com/shop/futon-frames                   # 200
+curl -s https://carolinafutons.com/products/savannah | grep -c name    # >= 1
+
+# 2. Velo HTTP functions still firing
+curl -X POST https://carolinafutons.com/_functions/wishlistService \
+  -H 'Content-Type: application/json' -H 'Origin: https://carolinafutons.com' \
+  -d '{"args":[]}'                                                      # 200 or 401 (NOT 404)
+
+# 3. Wix Stores API still serving from cfw
+curl -I https://carolinafutons.com/api/healthcheck                     # 200
+
+# 4. Order placement (manual, 1 cent test product)
+#    - place order via cfw checkout flow
+#    - confirm appears in Wix Dashboard within 60s
+#    - confirm wixEcom_onOrderCreated → confirmation email received
+#    - confirm cfw success page renders
+```
+
+If any step fails → **rollback immediately** per §1.5 and §4 cancellation flow.
+
+---
+
+## 9. Linked beads
+
+- **Parent:** cf-3qt (epic — full migration to Next.js + Wix Headless)
+- **Predecessor:** cf-3qt.8 (DNS cutover)
+- **Closed siblings (research that fed this doc):**
+  - cf-3qt.9.1 (`crew/melania/retirement-plan.md` — production unpublish + downgrade research)
+  - cf-3qt.9.2 (`crew/melania/wix-retirement-checklist.md` — staging site unpublish)
+- **Future scope:** cf-xe2 (full Wix exit — backend → Vercel Functions, Stores → headless commerce, CMS → Sanity/Payload)
--- a/docs/cf-3qt-9-wix-retirement-checklist.md
+++ b/docs/cf-3qt-9-wix-retirement-checklist.md
@@ -1,13 +1,15 @@
 # cf-3qt.9 — Wix Studio Retirement Checklist
 
 **Phase:** cf-3qt.9 (final phase of the cf-3qt epic — full migration to Next.js + Wix Headless)
-**Status:** Pre-work — DOCUMENTATION ONLY. Execution gated on cf-3qt.8 (DNS cutover) + 24 h prod stability.
+**Status:** Pre-work — DOCUMENTATION ONLY. Execution gated on cf-3qt.8 (DNS cutover) + the 30-day stability report defined in the cf-3qt.9 bead (uptime, order rate, performance, P0/P1 incident review).
 **Audience:** Stilgar (sign-off), melania (PM), engineering (execution)
-**Predecessors:** This consolidates the retirement research from cf-3qt.9.1 (`crew/melania/retirement-plan.md` — production focus) and cf-3qt.9.2 (`crew/melania/wix-retirement-checklist.md` — staging focus). Both authored by morgott on 2026-05-04.
+**Predecessors (research provenance — do NOT redirect operators back to these):** cf-3qt.9.1 (`crew/melania/retirement-plan.md`) and cf-3qt.9.2 (`crew/melania/wix-retirement-checklist.md`), both authored by morgott on 2026-05-04. This document is the single canonical source for execution; predecessors are kept for audit trail only.
 
 > ⚠️ **Do NOT execute any step in this document until:**
-> 1. cf-3qt.8 (DNS cutover) is complete AND `carolinafutons.com` has served from Vercel for ≥ 24 h with zero P0/P1 rollback triggers, AND
+> 1. cf-3qt.8 (DNS cutover) is complete AND `carolinafutons.com` has served from Vercel for the **30-day stability window** (cf-3qt.9 acceptance gate) with zero unaddressed P0/P1 rollback signals, AND
 > 2. Stilgar has signed every box in §6 below.
+>
+> The "≥ 24 h" wording from the morgott predecessor docs is a *sub-gate* of the 30-day window (no execution within 24 h of any rollback signal), NOT a substitute. **§6 wins** if any other section appears to disagree on the gate.
 
 ---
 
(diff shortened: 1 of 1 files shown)


DreadPirateRobertz:
## Review by code-simplifier (dispatched by godfrey for cf-3qt.9)

**Verdict: Useful consolidation, but the "Detail: see predecessor" pattern is at odds with the doc's actual job.** The new file is the canonical executor's script for a high-stakes cutover — the predecessors are now research artifacts. A handful of redundancies and pointer-mismatches are worth tightening before this becomes Stilgar's runbook.

Doc clarity > brevity stipulated — these are the spots where I think trimming or restructuring genuinely helps the executor, not the byte count.

### 1. The four "> Detail: predecessor file" pointers are misleading (§1, §2, §4, §5)

Each section opens with e.g. `> Detail: crew/melania/retirement-plan.md §1`, then re-states essentially all the substantive content. That pattern works for a research index, not an execution checklist. During cutover Stilgar should not have to context-switch to a second doc to verify a step.

Two options, either is fine:
- **Drop the pointers entirely** — this doc is now self-contained.
- **Reframe as a single "Research provenance" line** in the header (already present at line 6) and remove the four inline pointers. Saves four out-of-band redirects from the executor's mental model.

Same call for §1.5 rollback / §2.2 step list: keep the prose here, don't tell the executor to consult another file mid-step.

### 2. §4.6 is pure restatement of §6 (lines 192–194)

> ### 4.6 Stilgar approval gate
> Downgrade does NOT execute until Stilgar (a) confirms current plan + cost, (b) reviews the Wix Headless feature page, (c) signs §6.

§6.2 already lists the §4 approval gates. Either delete §4.6 or reduce to a single inline `> Gated on §6.2.` line. As written it's two redundant cross-references in one section.

### 3. §0 Scope table partially duplicates §3.1 / §3.2

`§0` is "scope at a glance" (5 rows, 2 columns: retire / stays). `§3.1` is "active dependencies" (7 rows, 3 cols, much more detail) and `§3.2` is "stop using" (5 rows, replacement column). The §0 right column is a less-informative subset of §3.1, and §0 left column ≈ §3.2.

For an executor's runbook this overlap is OK if §0 is sold as the orientation and §3 as the lookup. But the §0 left-column rows duplicate §3.2 verbatim (e.g. "Wix Studio rendering layer", "EDITOR_HOOKUP_GUIDE.html as the live developer reference", "cf-yw0m editor-hookup standing order"). Suggest either:
- Drop §0 and lead with §3 (which has the same info plus replacement column), or
- Keep §0 as a 1-paragraph at-a-glance and let §3 own the table form. Mixed-form duplication is what'll go stale first.

### 4. §6.3 post-execution sign-offs ↔ §8 smoke-test playbook

Five of the seven §6.3 bullets correspond 1:1 to §8 commands (Stores API smoke, cfw shop routes, order placement + email, price-match webhook, Q&A webhook). The remaining two (Sentry spike, tag push, retrospective) are §6.3-only. Right now they're presented as parallel lists with no link.

Tighter framing for an executor:
- §8 = "run these commands, capture output"
- §6.3 = "check these boxes after §8 passes" — and call §8 explicitly: `- [ ] §8 smoke-test playbook all green` instead of restating each curl.

That removes the ambiguity of "did I do the §6.3 item or the §8 item?" mid-cutover.

### 5. §4.2 plan table + §4.3 cost-delta table can be one table

§4.3's three rows (current / target / savings) are derived entirely from two cells in §4.2 ($25 and $13). Collapse into §4.2 with a footnote line below: `Estimated delta: ~$12/mo (~$144/yr) Business Unlimited → Headless`. One source of truth, one place to update if pricing re-verifies differently.

### 6. Minor info loss vs predecessor

`crew/melania/wix-retirement-checklist.md` documents Path B (Unpublish via Wix Studio Editor top-bar dropdown) for the staging site. The new doc only carries Path A. If the editor path is a viable fallback when the dashboard sidebar layout changes (Wix UI does shift), worth a one-line note in §2.2: `Editor fallback: top-bar Publish ▾ → Unpublish`. Cheap insurance.

### Not redundant — keep as-is

- Two execution log tables (one here, one in each predecessor): predecessors are read-only post-merge. Keep §7.
- §1.5 rollback one-liner: load-bearing during execution; correct to repeat even if predecessor has it.
- §1.3 "what unpublish does (and doesn't)" table: distinct framing from §3.1 (effect-of-action vs current-dependencies). Keep.
- §1.4 risks table + §4.5 risks table: section-scoped, correctly separated.

### Net suggestion

Roughly −20 to −30 lines (≈10%) achievable by (1)+(2)+(5) without touching the executor's situational awareness. (3) is a structural call for melania. (4) and (6) are clarity wins, not length wins.

The doc's bones are sound — gating, sign-off, smoke tests, rollback, log are all where they should be and ordered correctly for execution. The redundancy is concentrated in the "consolidating-from-predecessors" seams, which is exactly where you'd expect it.

DreadPirateRobertz:
## Review by silent-failure-hunter (dispatched by godfrey for cf-3qt.9)

I focused on the five concerns from the dispatch. The doc is well-structured and the gating is mostly sound, but I found several silent-failure modes — checks that look protective but could pass while the real failure mode they're meant to catch is occurring. Severity-tagged findings below.

---

### CRITICAL findings

**C1. §1.5 rollback claim is materially wrong — "one click" hides the destructive intermediate state.**
Quote: "Rollback is one click: Wix dashboard → Publishing → Publish Site. No DNS changes needed."

This is misleading in a way that will burn ops at exactly the wrong moment. Re-publishing a Wix Studio site that has been unpublished for any non-trivial period:
- Re-asserts Wix's primary-domain mapping for `carolinafutons.com` in the Wix dashboard. Because DNS still points at Vercel, *the user-visible site* will still serve from Vercel — but Wix internal routing tables, sitemaps, and bots can begin treating Wix as authoritative again, which is exactly the SEO regression risk §1.4 names but does not connect to rollback.
- Republishes whatever was last in the Wix Studio editor — which by §5 will have been EDITOR_HOOKUP_GUIDE-driven content that has been drifting from cfw for weeks/months. If anyone has made even an "experimental" save in the editor since the last canonical publish, that drifted content goes live (in the Wix render path) on republish.
- Does not restore A/B test or scheduled-publish state.
- Does not undo §4 (plan downgrade). If §4 already executed and the Wix Headless plan strips the rendering tier, the "one click" Publish button may be greyed out or fail with a plan-tier error — and the rollback path silently doesn't exist.

The rollback section needs to (a) state the order-of-operations constraint (you can only rollback §1 cleanly if §4 has not yet executed, OR you must first re-upgrade the plan), (b) acknowledge editor-content drift risk, (c) include a verification step that re-published content matches the last-known-good Wix Studio snapshot. Right now this reads like "no risk, one click" and that's a silent-failure trap.

**Fix:** Rewrite §1.5 as a sequenced rollback playbook with order-dependent steps and explicit verification, and link it from §4.5 (plan downgrade risk table).

---

**C2. §6.2 sign-off gate has a self-approval bypass.**
Quote: "Stilgar approves staging site unpublish (§2) — *can be self-approved by ops if not gated*"

"If not gated" is the entire question, and the doc never resolves it. Any agent reading this can choose to interpret their own action as not-gated and skip the sign-off. The §2.1 pre-flight only asks the executor to "confirm" two things — there is no independent verifier. An agent who unpublishes `My Site 5` while it actually has a Velo webhook receiving production traffic (because someone left a misrouted endpoint in Wix Secrets Manager pointing at the staging metaSiteId) would silently break production with full doc compliance.

**Fix:** Remove the self-approval clause OR require a concrete, falsifiable check ("grep `3af610bf` across `src/backend/**` and `crew/*/secrets*` — must return zero matches; paste output into §7"). The current escape hatch is exactly the silent-failure pattern this gate is supposed to prevent.

---

**C3. §8 smoke-test #2 (Velo HTTP) accepts 401 as success — this hides auth-broken-by-unpublish.**
Quote: `curl -X POST https://carolinafutons.com/_functions/wishlistService ... # 200 or 401 (NOT 404)`

Treating 401 as a passing signal is a textbook silent failure. A 401 means *something* responded — but if the unpublish accidentally invalidated the Velo HTTP function's auth context (e.g., session cookie, member token, or per-route permission rule), every authenticated call will return 401 regardless of payload. The smoke test passes; production is broken for every signed-in user.

The check is supposed to prove "Velo is still firing." A 401 does not prove that. It proves "something on the path returned an HTTP error code." Unpublishing can plausibly change which middleware runs in front of `/_functions/*` — and Wix has a documented history of plan-tier changes affecting Velo permission enforcement.

**Fix:** Use an endpoint that takes an authenticated payload that this script can sign, OR use a function known to return 200 on empty input (most `submit*` Velo handlers return 400 on missing fields, which is a positive proof of execution). At minimum, distinguish "404 = function gone" from "401 = auth path changed unexpectedly" — treat both as failures, not just 404.

---

### HIGH findings

**H1. §1.1 pre-flight `curl -I` check for `server: Vercel` header is bypassable.**
Quote: `curl -I https://carolinafutons.com → header server: Vercel (NOT Wix)`

Vercel does emit `server: Vercel` — but so does anything behind a CDN that strips/rewrites the header, and Wix's edge can serve cached HTML with arbitrary upstream headers if the page was edge-cached during the cutover window. This check proves "the response went through some node that wrote `server: Vercel`" — it does not prove the *origin* is Vercel. The stronger check is `curl -sI ... | grep -i 'x-vercel-id\|x-vercel-cache'` — those headers are always present and non-spoofable from an outside observer.

**Fix:** Replace the `server:` header check with `x-vercel-id` (always present) and add a second check on response body fingerprint (`grep -c 'next/static\|_next/'`).

---

**H2. §1.1 pre-flight "zero organic traffic on Wix-rendered URLs" — unfalsifiable as written.**
Quote: "Wix dashboard → Analytics → confirm zero organic traffic on Wix-rendered URLs in the last 24 h"

What counts as a "Wix-rendered URL" post-DNS-cutover? Once DNS points to Vercel, the Wix analytics dashboard reports zero traffic *because Wix is no longer serving the requests*, not because no one is hitting Wix-rendered content. This check will pass trivially in 100% of cases — including the case where Wix is still serving an old cached page on a stale CDN PoP somewhere.

**Fix:** Drop this check, or rewrite as: "Search Console → check for indexed URLs whose response fingerprint matches Wix (presence of `wix-public-html-renderer` in HTML); zero such URLs in the last 24h crawl."

---

**H3. §1.4 risk table downplays SEO regression as "Medium" with hand-wavy mitigation.**
Quote: "SEO regression — Google still has Wix-rendered URLs in index | Medium | Vercel pages must be in Search Console coverage report ≥ 24 h pre-unpublish"

24 hours of Search Console coverage is dramatically insufficient for any non-trivial-traffic site. Google's recrawl cycle for product pages can be days to weeks; "in the coverage report" only means Google saw the URL once — it does not mean the Wix variant has been deindexed or that the canonical tag has propagated. If Wix-rendered URLs are still indexed when §1 executes, they will start returning whatever Wix's unpublished-site placeholder is (typically a Wix-branded "site is not published" page), which Google will index as the new content and rank below the Vercel canonical.

This risk should be HIGH and the mitigation should require: (a) `site:carolinafutons.com` returns Vercel-fingerprinted pages for the top 50 product URLs, (b) rel-canonical tags from any remaining Wix surface point to the Vercel domain, (c) sitemap.xml has been re-submitted from the Vercel origin.

**Fix:** Bump severity to HIGH, replace mitigation with concrete URL-list verification.

---

**H4. §1.4 "Email triggers stop reaching ops" — likelihood "Low" undercounts the actual failure mode.**
Quote: "Email triggers (price-match, Q&A) stop reaching ops | Low | Webhooks live in Velo; un-coupled from rendering. Verify §8."

The forms that *trigger* those emails live in the rendering layer. If any form-submission flow still routes through a Wix Studio-rendered intermediate page (e.g., a Wix-hosted "thank you" page that fires the analytics ping that triggers the email), unpublishing kills the email. §8 only smoke-tests two specific endpoints — it does not exercise the actual user form-submission paths.

**Fix:** Add explicit verification in §8 that each of these flows is invoked end-to-end from the cfw side: price-match form, Q&A submission, contact form, swatch request, newsletter signup. Currently §6.3 has two of these as post-execution checks but they're not wired into §8.

---

**H5. §4.5 "Velo execution quotas lower on Headless" mitigation is conditional and untested.**
Quote: "Pull current Velo quota usage. Compare to Headless quota. Test in staging if uncertain."

"If uncertain" is a silent-failure trap. The default behavior is "don't test" — and quota differences are exactly the kind of thing where a low-traffic test won't surface the issue until peak load. The mitigation should be unconditional ("MUST test under at-or-above-current-peak load") or this risk should bump to HIGH.

**Fix:** Remove "if uncertain"; require quota comparison + load test or escalate severity.

---

### MEDIUM findings

**M1. §1.1 pre-flight has no Sentry quantitative threshold.**
Quote: "Sentry P0/P1 rate at baseline for ≥ 24 h post-cutover"

What is "baseline"? Without a number, this passes whenever the executor *feels* it passes. Specify a threshold (e.g., "< 0.5 P0/P1 events per hour averaged over 24h, or zero if baseline is zero").

**M2. §6.1 "30-day stability report green" — no link to where the report is defined.**
The acceptance gate is referenced but the actual report criteria (what "green" means) are not defined or linked. An executor could rubber-stamp this without knowing what it requires.

**M3. §4.6 sign-off gate phrasing — "or chooses to defer to a separate decision".**
This lets §1 execute without §4. That may be intentional, but if §1 executes first and then §4 is later attempted, see C1: the rollback path narrows. Document the order-of-operations constraint between §1 and §4 explicitly.

**M4. §8 smoke #4 (order placement) is manual and has no failure-detection threshold.**
"confirm appears in Wix Dashboard within 60s" — what if it appears in 90s? The check is binary as written but the failure mode (delayed webhook delivery) is gradient. Add: "If > 60s, treat as PARTIAL FAILURE; investigate before declaring §8 passed."

**M5. §5.5 standing-order removal not gated on §1 success.**
The doc allows removing the cf-yw0m editor-hookup standing order (§5.5) while §1.5 rollback is still possible. If §1 has to roll back, the editor-hookup standing order is suddenly relevant again — but the memory has already been wiped. Recommend gating §5.5 on "§1 execution +14 days with no rollback."

---

### What the doc does well (rare but worth noting)

- §0 scope-at-a-glance table is unusually clear about what's NOT being retired (Velo, Stores API, CMS) — this directly prevents a class of "I thought we were retiring Velo" silent-failure conversations later.
- §3.3 explicit "Future scope (NOT in cf-3qt.9)" with cf-xe2 link — good scope-creep firewall.
- §6.3 post-execution sign-off list including the Sentry P0/P1 24h check is the right shape.
- §7 execution log table forces a paper trail.

---

### Summary

The structural gating (§6) is the right idea and the smoke-test (§8) is the right idea — but several specific checks are written in a way that lets the failure mode they're targeting slip through. The biggest concerns are C1 (rollback claim is wrong-by-omission about ordering with §4 and editor drift), C2 (self-approval bypass on §2 sign-off), and C3 (401 accepted as success in Velo smoke test). Fixing those three closes the worst silent-failure traps; H1–H5 and the M-tier items are quality-of-evidence improvements.

This is a doc-only PR so none of these findings block merge — but every one of them should be resolved before §1 of the checklist is actually executed. Recommend: merge the doc, then open a follow-up bead "cf-3qt.9 checklist hardening" that addresses C1/C2/C3 + H1–H5 before execution day.

— silent-failure-hunter

DreadPirateRobertz:
**Review by code-reviewer (dispatched by godfrey for cf-3qt.9)**

Doc-only PR. Scope check vs melania's nudge: all four areas are present (§1 unpublish, §3 services that stay active, §4 plan downgrade, §5 guide archival). Cross-references to morgott's `crew/melania/retirement-plan.md` (cf-3qt.9.1) and `crew/melania/wix-retirement-checklist.md` (cf-3qt.9.2) faithfully reproduce their substance — pricing, metaSiteId `3af610bf`, the ~$12/mo / ~$144/yr delta, the rollback mechanic, and the `My Site 5` UI path all match. The execution gate (§6 Stilgar sign-off) reads correctly: §1, §4, §5 are explicitly held until §6 boxes are checked, and the document's own preamble (lines 8–10) restates the gate. No execution path can fire prematurely from following this doc.

That said, two of the smoke-test commands in §1 / §8 will give **false signals** — one a guaranteed false negative, one a guaranteed false positive. Both should be fixed before this is used as a Stilgar-facing playbook.

---

**Critical (90–100)**

**1. `/_functions/wishlistService` is not an HTTP-function endpoint — line 323-325 will always 404** (confidence 95)

§8 step 2 (lines 322–325) curls `/_functions/wishlistService` and instructs the operator to interpret "200 or 401 (NOT 404)" as Velo-up. But `wishlistService` lives in `src/backend/wishlistService.web.js` (a `webMethod` module callable only from in-Wix code via `wix-web-module`), not in `src/backend/http-functions.js`. Only exports of the form `get_<name>` / `post_<name>` from `http-functions.js` are reachable at `/_functions/<name>`. Grep of `src/backend/http-functions.js` confirms zero `wishlistService` HTTP exports.

Effect: this curl will return 404 unconditionally even when Velo is fully healthy, causing the smoke test to falsely fail and triggering §1.5 / §4 rollback against a healthy system.

Fix: replace with an actual HTTP function. Good candidates that already exist in `src/backend/http-functions.js`:
- `curl -I https://carolinafutons.com/_functions/health` (expect 200 — `get_health` is exported)
- `curl -I https://carolinafutons.com/_functions/sitemapXml` (expect 200 — `get_sitemapXml` is exported)
- `curl -I https://carolinafutons.com/_functions/manifest` (expect 200 — `get_manifest`)

Any of those exercises Velo's HTTP runtime end-to-end without needing a body.

**2. `https://carolinafutons.com/api/healthcheck` does not exist — lines 37 and 328 will always 404** (confidence 92)

Both §1.1 ("Vercel storefront serving" pre-flight, line 37) and §8 step 3 ("Wix Stores API still serving from cfw", line 328) curl `/api/healthcheck` and expect `200 OK`. There is no such route in the Next.js app — `carolina-futons-web/src/app/api/` contains `admin`, `auth`, `cart`, `cross-rig`, `delivery-zone`, `email`, `newsletter`, `notify-me`, `order-lookup`, `revalidate`, `search`, `swatch-request`, `wishlist`. No `healthcheck` directory. The same path was also used in `crew/melania/retirement-plan.md §1` pre-flight — the predecessor doc had the same gap, but consolidating it forward propagates the bug.

The §8 comment "Wix Stores API still serving from cfw" is also conceptually muddled — the Wix Stores API lives at `*.wixapis.com`, not on the cfw domain. What that step is actually trying to verify is "cfw still rendering and able to call Wix Headless from the server."

Fix options (pick one, ideally both):
- Add a real `src/app/api/healthcheck/route.ts` to cfw (returns 200 + ping Wix Stores SDK) and land that PR before this doc is executed; then keep the curl as-is.
- Or replace the curl with checks that already exist: `curl -fI https://carolinafutons.com/shop/futon-frames` (line 319 already covers Vercel-serving) plus a verified Velo endpoint (see issue #1) for "backend live."
- For "Wix Stores API still serving" the truthful test is to load a PDP and assert it renders (which already pulls a Headless SDK call): `curl -fs https://carolinafutons.com/products/savannah | grep -q "add to cart"`.

Either fix is acceptable — what is not acceptable is shipping the playbook with a 404 step that the operator is told means "rollback immediately" (line 337).

---

**Important (80–89)**

**3. 24 h vs 30-day stability gate inconsistency — preamble (line 9) and §6.1 (line 268) disagree** (confidence 85)

Line 9 says execution is gated on "≥ 24 h with zero P0/P1 rollback triggers." Line 268 §6.1 requires "**30-day stability report green**: uptime, order rate, performance, P0/P1 incident review." Those are two very different bars (one day vs one month) and Stilgar will not know which one you mean. Either:
- The 30-day window is the real cf-3qt.9 acceptance gate (per the parenthetical "cf-3qt.9 acceptance gate"), in which case the preamble §0 needs to say 30 days, not 24 h, or
- The 24 h window is the gate and §6.1 should drop the "30-day" framing.

Pick one, restate consistently, and state explicitly whether 24 h is the *minimum to start* and 30 days is the full retrospective window, or whether they're alternatives.

**4. §0 retirement table claims `My Site 5` is being retired, but §2.3 says no Stilgar gate — clarify what "retired" means for staging** (confidence 80)

The "What gets retired" column at line 21 lists `My Site 5` (metaSiteId `3af610bf`) alongside production rendering, but §2.3 (line 101) immediately downgrades the bar: "No Stilgar sign-off needed for `My Site 5` — it's a staging/dev site with no production traffic." That's defensible, but a reader following §6 will see no §6.2 box that requires staging unpublish before production. The §6.2 entry at line 275 mentions staging "*can be self-approved by ops if not gated*" — which contradicts §2.3's flat "no sign-off needed."

Tighten one of these. Either staging is gated and §2.3 is wrong, or staging is ungated and §6.2 should drop the bullet (it adds noise to a sign-off list whose whole purpose is to be the gate). Recommendation: drop the §6.2 bullet (line 275) and have §2 stand alone as a self-service section, with a single forward-pointer "ops may execute §2 at any time, no §6 dependency."

**5. §3.1 `wixEcom_onFulfillmentCreated (post cf-fovb)` — confirm cf-fovb is in scope before this doc executes** (confidence 80)

Line 119 lists `wixEcom_onFulfillmentCreated` with the parenthetical "(post cf-fovb)." If cf-fovb has not landed by the time §1 executes, the smoke test in §6.3 line 286 ("Wix dashboard order management still functional (place test order)") and §8 step 4 (lines 330–334) won't exercise that handler. Either drop the "(post cf-fovb)" hedge from this checklist (it's not the right place to track open work — cf-xe2 / cf-fovb beads are) or add a §6.1 pre-condition box: `[ ] cf-fovb merged + onFulfillmentCreated wired (or explicitly out of scope)`.

---

**Minor (low-confidence, FYI not blocking)**

- Line 231 (banner template): the link `[cf-3qt](../bd/cf-3qt)` resolves to `docs/bd/cf-3qt` from `docs/archive/<file>.md`. There is no such filesystem path — beads are not files. Either drop the link or point at a real artifact (e.g. the bead's GitHub issue / a cf-3qt retrospective page once §6.3's `docs/cf-3qt-retrospective.md` exists). Confidence 65 (banner template only, easy to fix later).
- Line 84 / line 17 reference `chrisdealglass.wixstudio.com/my-site` — predecessor doc says the production STAGING_SITE and `My Site 5` share the same subdomain prefix in the Sites list. Worth a verify-by-screenshot before unpublish to make sure the operator clicks the right card. Confidence 55.
- Line 320 `grep -c name` on `/products/savannah` — `name` will match plenty of incidental HTML (`name=`, `<meta name>`, etc.) and produce a misleading non-zero count even on an error page. Use `grep -q "savannah"` or `grep -q "add to cart"` for a meaningful liveness signal. Confidence 70.

---

Bottom line: structure, gating, and PM coverage are solid. The two smoke-test endpoint bugs (`_functions/wishlistService`, `/api/healthcheck`) are the only items that materially affect operator outcomes — both will misfire 100% of the time on a healthy system and should land fixed before Stilgar uses this as the playbook.

DreadPirateRobertz:
## Review by comment-analyzer (dispatched by godfrey for cf-3qt.9)

**Scope:** docs/cf-3qt-9-wix-retirement-checklist.md (new, +348). progress-report.md update is incidental.

This is a doc-as-comment review: I treated every factual claim as a comment to verify against ground truth (codebase, predecessor docs, beads, cfw repo).

**Summary:** Strong consolidation doc with one critical accuracy bug, two factual gaps worth fixing, and a few wording precisions. Most cross-references and code-shape claims check out cleanly.

---

### Critical Issues

**1. `docs/SITE-OWNER-GUIDE.md` and `docs/TESTING-GUIDE.md` do not exist**
- Location: §3.2 (replacement column) and §5.3 banner template
- Issue: The doc tells future Stilgar/devs that the *current* developer reference lives at `docs/SITE-OWNER-GUIDE.md` + `docs/TESTING-GUIDE.md`. Neither path exists in the repo today (verified at `/Users/hal/gt/cfutons/docs/`). Closest extant files are `docs/brenda-admin-guide.md` and `docs/TESTING-PROCEDURE.md` (plus `docs/testing-master.md`).
- Why this matters: This is the load-bearing pointer the RETIRED banner template will paste into the archived `EDITOR_HOOKUP_GUIDE.html` itself. Shipping that banner verbatim (per §5.3) creates a permanent broken-link in the archive — the very document people will reach for when historical context is needed.
- Suggestion: Either (a) update the references to the correct file names (`docs/brenda-admin-guide.md` + `docs/TESTING-PROCEDURE.md`), or (b) add a §5.3a step "before archiving, confirm/create the canonical replacement docs at the referenced paths". Inherited from the predecessor doc — `crew/melania/retirement-plan.md §2` makes the same incorrect claim.

---

### Improvement Opportunities

**2. §3.1 row "Wix Secrets Manager" is true but understates Velo coupling**
- Location: §3.1 active-dependencies table
- Current state: The doc claims `wix-secrets-backend` is wired in Velo runtime. Verified — `src/backend/http-functions.js` imports `wix-secrets-backend` 10× (line 33 plus 9 lazy `await import()` sites for WWEX/Turnstile/mailer paths). Claim is accurate.
- Suggestion: Add a callout to §4.5 risks that "Velo execution quotas lower on Headless" specifically affects the lazy-imported secret retrievals on hot paths (cron + webhook). If quota is shared across imports, 10 import sites is a non-trivial fan-out worth flagging for the downgrade smoke test.

**3. §3.1 cfw wiring line items omit the actual file paths**
- Location: §3.1 "Where it's wired" column
- Current state: Doc says cfw `lib/wix-client.ts`, `lib/wix-cms.ts`, `actions/cart.ts`. The cfw repo uses `src/` prefix — actual paths are `src/lib/wix-client.ts`, `src/lib/wix/velo-client.ts` (the actual `/_functions/*` dispatcher), and various `src/app/**/actions.ts` + `src/app/api/**/route.ts` files. There is no `lib/wix-cms.ts` at any path I can find.
- Suggestion: Update paths to reflect the `src/` prefix and replace the non-existent `lib/wix-cms.ts` with `src/lib/wix/velo-client.ts` (the actual Velo dispatcher) and `src/lib/wix/products.ts` (the @wix SDK consumer). Future maintainers grepping for these paths will fail otherwise.

**4. §8 `/_functions/wishlistService` smoke-test endpoint — verified, but the example payload will not exercise it**
- Location: §8 step 2
- Current state: Verified `post_wishlistService` exists at `src/backend/http-functions.js:2938` (RPC dispatcher landed via PR #1164). However the example body `{"args":[]}` is missing the `method` field the dispatcher requires (line ~2968: `const fn = wishlistServiceModule[method]`). With no method, the call returns 400, not 200 or 401 as documented.
- Suggestion: Either change the comment to `# 400 (NOT 404) — confirms route exists` or supply a real method+args, e.g. `{"method":"getWishlistByMemberId","args":["test"]}` which would return 401 (auth) and exercise the actual code path.

---

### Recommended Removals

**5. §1.4 "Wix dashboard login breaks — Likelihood: None"**
- Location: §1.4 risks table
- Rationale: "None" is not a likelihood, and this row exists only to reassure that something *won't* break. Tables of risks should list actual risks; reassurances belong in §1.3 (already covered: "Wix dashboard / order management — Still functional"). The duplication adds noise without changing the action plan.
- Suggestion: Drop the row, or merge it as a footnote under §1.3.

---

### Verified Findings (no action needed)

- **metaSiteId `3af610bf` for staging:** Matches `crew/melania/wix-retirement-checklist.md` exactly (multiple occurrences including dashboard URL `manage.wix.com/dashboard/3af610bf/home`).
- **Plan tier names + prices in §4.2:** Match `crew/melania/retirement-plan.md §3` exactly (Wix Headless ~$13, Business Basic ~$17, Business Unlimited ~$25, Business VIP ~$35; ~$144/yr savings calc). The "verify at wix.com/upgrade" caveat is properly carried forward.
- **Velo Stores events in §3.1:** All four event handlers (`wixEcom_onOrderCreated`, `_onOrderApproved`, `_onOrderFulfilled`, `_onFulfillmentCreated`) verified present in `src/backend/events.js` (lines 214, 319, 338, 375). The `(post cf-fovb)` parenthetical for `onFulfillmentCreated` is appropriate context.
- **cfw `@wix/stores` / `@wix/auto_sdk_stores_*` claim:** Verified — `src/lib/wix-client.ts` imports `@wix/auto_sdk_stores_products` and `@wix/auto_sdk_stores_collections`. SDK use is real.
- **cfw `/_functions/*` routes:** Verified across `src/lib/wix/velo-client.ts`, `src/lib/wix/custom-events.ts`, `src/app/community-gallery/actions.ts`, `src/app/contact/actions.ts`, `src/app/actions/notify-me.ts`, `src/app/actions/swatch-request.ts`, `src/app/api/*/route.ts`, etc. Heavy real usage.
- **cf-3qt.9 acceptance criteria reference (§6.1 "30-day stability report green"):** Matches the bead's Acceptance section ("Stability report signed off by Stilgar + mayor"). The "30-day" framing matches the bead's "30 days post-Phase 8 cutover" prereq.
- **Cross-bead links (§9):**
  - cf-3qt — valid (epic, OPEN, P1)
  - cf-3qt.8 — valid (OPEN, P1, assigned millicent, blocks cf-3qt.9)
  - cf-3qt.9.1 — valid (CLOSED, morgott, retirement-plan.md output)
  - cf-3qt.9.2 — valid (CLOSED, morgott, wix-retirement-checklist.md output)
  - cf-xe2 — valid (epic, P4, DEFERRED). The doc's framing of cf-xe2 as "full Wix exit" matches the bead's title and description verbatim. Good scope-creep guard.
- **Banner template HTML/Markdown:** Self-consistent, just relies on the broken paths in finding #1.

---

### Positive Findings

- §0 "Scope at a glance" table is a model summary — the *retire vs stay-active* split is exactly the question Stilgar will ask, and answering it on page 1 saves a 348-line read.
- §6 sign-off gate is properly tri-phased (preconditions / per-step / post-execution) and reuses checkbox state machine — easy for Stilgar to drive.
- §3.3 explicitly fences cf-xe2 work *out of scope*. Carrying that forward into §9 is good belt-and-suspenders against future-scope creep.
- §1.5 single-line rollback is honest and accurate (verified — Wix unpublish/republish toggles render layer only, doesn't touch DNS).
- The "Re-verify pricing at wix.com/upgrade" warning appears 3× (§4 header, §4.2 callout, §4.5 risk). Belt-and-suspenders done right; pricing is the most-likely-to-rot fact in the doc.

---

**Bottom line:** Ship-blocker is finding #1 (broken doc paths in the RETIRED banner). Fix that and tighten finding #4's smoke-test payload, and the doc is ready for Stilgar review. The other items are quality-of-life fixes safe to land in a follow-up.

DreadPirateRobertz:
**Review by pr-test-analyzer (dispatched by godfrey for cf-3qt.9)**

## Summary

Doc is well-structured and the §8 smoke-test playbook covers the four high-level integration surfaces named in the assignment (Vercel storefront, one Velo HTTP function, Wix Stores API healthcheck, manual order placement). However, §8 is materially **under-specified relative to the live integration surface area listed in §3.1** — and several risks called out in §1.4 ("email triggers stop reaching ops", "Velo HTTP functions stop firing", "webhook delivery breaks") are not actually exercised by the smoke test.

The doc is gating live infrastructure changes (production unpublish + plan downgrade) on this playbook. It should be tighter before §6 sign-off.

## Critical gaps (rated 8-10)

### 1. Velo event handlers are only verified indirectly via "place a test order" — most are not exercised at all (criticality 9)

§3.1 calls out the email/SMS notification orchestrator as wired through these triggers in `src/backend/events.js`:

- `wixEcom_onOrderCreated` — confirmation email path
- `wixEcom_onOrderApproved` — gamification earn + receipt
- `wixEcom_onOrderFulfilled` — shipped SMS via notificationOrchestrator
- `wixEcom_onFulfillmentCreated` — fulfillment email (post cf-fovb)
- `wixEcom_onFulfillmentUpdated`, `wixEcom_onOrderDelivered`, `wixEcom_onOrderCanceled`
- `wixEcom_onAbandonedCheckoutCreated` / `Recovered` — cart recovery
- `wixStores_onProductCreated` / `Updated`, `wixStores_onInventoryVariantUpdated`
- `wixMembers_onMemberCreated` / `Updated` — birthday + welcome
- `wixBlog_onPostPublished`
- `wixData_onDataItemUpdated`

§8 step 4 only manually verifies `wixEcom_onOrderCreated` (confirmation email). The fulfillment/shipped/delivered paths and the abandoned-cart recovery handler are not exercised.

**What this misses:** if unpublishing the rendering tier silently disables a subset of `wixEcom_*` event subscriptions (Wix has historically had quirks here when sites are unpublished), order confirmation still fires but ops won't notice that shipping notifications, fulfillment emails, or abandoned-cart recovery have gone dark — for days.

**Recommendation:** add to §8:
- After the test order is placed, mark it as Approved + Fulfilled in the Wix Dashboard and confirm (a) order-shipped SMS lands, (b) `wixEcom_onFulfillmentCreated` fired (check fulfillment email + audit log).
- Trigger an abandoned-cart event (add to cart, leave 1h, confirm `wixEcom_onAbandonedCheckoutCreated` enqueued the recovery email).

### 2. Velo Jobs Scheduler (cron) is not in the smoke test at all (criticality 9)

`src/backend/jobs.config` defines **22 scheduled jobs** that run on Velo's cron infrastructure — `processEmailQueue` (every 15 min), `triggerAbandonedCartRecovery` (hourly), `syncInventoryFromStore` (every 30 min), `checkWishlistPriceDrops`, `runReviewRequestEmails`, `sendWeeklyDigestEmail`, `dailySocialStories`, `refreshFacebookCatalog`, `detectPriceDrops`, `processDeliveryDayOfReminders`, etc.

§3.1 lists Velo events but **does not mention `jobs.config` at all**, and §8 has zero verification that scheduled jobs continue running post-unpublish.

**What this misses:** if any Wix billing/plan transition disables the jobs scheduler (the Headless plan downgrade in §4 is plausible — Wix has historically gated cron behind Business plans), `processEmailQueue` stops draining → entire EmailQueue silently stalls → no review-request emails, no abandoned-cart recovery, no winback, no analytics digest. Ops won't notice for hours-to-days.

**Recommendations:**
1. Add a §3.1 row: "Velo Jobs Scheduler (`src/backend/jobs.config`) — 22 cron jobs powering email queue drain, inventory sync, wishlist alerts, social story scheduling, lifecycle emails."
2. Add to §8: query EmailQueue and confirm `lastProcessedAt` advanced within 15 min post-unpublish. Equivalent: confirm `InventoryThresholds.updatedDate` advanced within 30 min.
3. Add to §4.5 risk table: "Velo Jobs Scheduler quota or feature differs on Wix Headless plan — verify before downgrade. If unavailable on Headless, the downgrade itself becomes a P0."

### 3. /_functions surface is exercised on a single endpoint (criticality 8)

`src/backend/http-functions.js` exports **81 distinct HTTP endpoints** — from public POST handlers (`contactSubmissions`, `mailingListSignups`, `notifyMe`, `submitQuestion`, `wishlistService`, `sampleRequests`, `unsubscribe`, `gamificationCore`, `loyaltyService`, `pushNotificationService`, `styleQuiz`, `submitCommunityPhoto`, `giftRegistry`, `trackCustomEvent`, `trackReferral`, `addBundleToCart`, `answerQuestion`, `recordSpinGrant`, `submitSurvey`, `gamificationEvent`, `challengeProgress`, `busEvent`) to feed/sitemap GETs (`googleShoppingFeed`, `productSitemap`, `blogSitemap`, `blogRssFeed`, `facebookCatalogFeed`, `pinterestProductFeed`, `sitemapXml`, `manifest`, `serviceWorker`, `robots`/`robotsTxt`, `topicCluster`, `loyalty`, `health`, `unsubscribe`, `productQA`, `bundles`, `activeChallenges`, `leaderboard`, `badges`, `deliveryZone`, `campaignAnalytics`) to webhooks (`stampedWebhook`, `klaviyoWebhook`) to internal cron triggers (`triggerCartRecoveryCron`, `triggerBrowseRecoveryCron`, `processEmailQueueCron`, `triggerReengagementCron`, `scanAndTriggerWinbackCron`, `runReviewRequestEmailsCron`, `processPostPurchaseCareCron`, `processContentScheduleCron`, `processNotificationQueueCron`, `cleanupRateLimitCron`, `checkWishlistAlerts`, `cmsGarbageCollect`, `sendMonthlyLoyaltyStatements`, `weeklyBlogDigestCron`).

§8 hits exactly **one** of those (`wishlistService`) and infers "the rest are fine because Velo runtime survives." That is not a smoke test — that is an assumption.

**Recommendations:**
- Add cURL probes for at least one endpoint per category: a GET feed (`googleShoppingFeed` or `sitemapXml`), a public POST (`contactSubmissions` with a synthetic payload), a webhook (`stampedWebhook` with the documented test signature), and one cron trigger (`processEmailQueueCron`). Expect 200/204/401 — not 404.
- Document the exact "404 means the function file no longer routes" failure-mode signature so the operator knows what the bad case looks like.

### 4. Wix Secrets Manager is in §3.1 but never probed in §8 (criticality 8)

§3.1 lists "Wix Secrets Manager — WWEX SOAP credentials, Turnstile keys, mailer creds." Every Velo backend that calls `wix-secrets-backend` depends on this being readable post-unpublish. Two failure modes are plausible:

1. Plan downgrade (§4) revokes the Secrets Manager API on the Headless tier (Wix has tiered access to several "advanced" features by plan).
2. Unpublishing changes the runtime context such that secrets resolve differently.

§8 has no test that any secret is still readable. Indirectly, the manual order placement step would catch the mailer breaking, but Turnstile (CAPTCHA on contact + swatch) and WWEX (freight quotes) would not be exercised.

**Recommendation:** add a §8 step:
- Submit the `/contact` form with a real Turnstile token → expect 200 (validates Turnstile secret read).
- Hit a freight-quote endpoint with a synthetic SKU → expect a quote response (validates WWEX secret read).
- Trigger one EmailQueue drain cycle → confirm a queued email actually sends (validates mailer creds read).

## Important improvements (rated 5-7)

### 5. cfw → /_functions CORS contract is not verified (criticality 7)

The cfw frontend calls Velo HTTP functions cross-origin from `carolinafutons.com` (Vercel) to `https://chrisdealglass.wixstudio.com/.../_functions/*`. `http-functions.js` defines `options_*` CORS preflight handlers for many endpoints. If unpublish breaks CORS preflight (e.g., the OPTIONS handlers stop being registered when rendering is disabled), every cfw POST will fail in browser — but server-side cURL from §8 will pass because cURL doesn't preflight.

**Recommendation:** add to §8 a browser-side verification: open `carolinafutons.com/contact` in a real browser, submit form, observe DevTools Network tab — preflight OPTIONS returns 200, POST returns 200. Server-side cURL is necessary but not sufficient.

### 6. Order placement test is loose (criticality 6)

§8 step 4 says "place order via cfw checkout flow" but does not specify:
- Which payment method (Stripe? PayPal? Wix-managed?). Each routes through a different path.
- Whether to verify `wixEcom_onOrderApproved` (gamification earn) fired, not just `wixEcom_onOrderCreated`.
- Whether the order shows up in fulfillment exports (WWEX integration) within the expected SLA.

**Recommendation:** turn step 4 into a checklist: order placed via Stripe (1¢ test product) → cfw success page renders → order in Wix Dashboard within 60s → confirmation email received → `wixEcom_onOrderApproved` audit log entry → mark Approved → confirm gamification points awarded → mark Fulfilled → confirm shipped SMS sent → mark Canceled → confirm cancellation email sent. That single end-to-end run covers 5 of the §3.1 event handlers.

### 7. No probe for Wix CMS write paths (criticality 6)

§3.1 says "Wix CMS collections — Velo `*.web.js` for writes." Many Velo modules (`emailService`, `analyticsHelpers`, `cartSessionService`, `notificationOrchestrator`, `referralService`, `swatchKitService`, `wishlistService`) write to CMS collections. §8 only verifies cfw can READ from CMS (via `/api/healthcheck` and product page rendering). A POST that ends in a CMS insert (e.g., `contactSubmissions` writing to `ContactSubmissions`) is the right probe.

### 8. No rollback dry-run rehearsal (criticality 5)

§1.5 says "Rollback is one click: Wix Dashboard → Publishing → Publish Site." This has not been rehearsed. Recommend adding a §8 dry-run step *before* unpublishing production: unpublish the staging `My Site 5` (already in §2), republish it, time the round trip. If staging rollback takes >5 min, the production rollback risk window is larger than implied.

## Test quality issues

- **§8 step 1's `grep -c name`** on `/products/savannah` is fragile — the literal string "name" appears in many unrelated contexts (HTML attributes, schema.org JSON-LD, etc.) and will return ≥1 even on a 404 page that contains the word "name." Replace with a check that looks for a slug-specific marker (the product H1 text or canonical URL).
- **§8 step 2's wishlist probe** treats 401 as success ("200 or 401 — NOT 404"). 401 is fine if the endpoint requires auth, but `wishlistService` accepts anonymous requests with a guest token. A 401 would itself be a regression. Tighten to "200" — or document why 401 is acceptable.
- **§8 step 3** (`/api/healthcheck` returning 200) only proves the cfw Next.js server is up. It says nothing about whether cfw can reach the Wix Stores API. Either point the healthcheck endpoint at a Wix API call internally (probably already does), or replace with a probe that exercises the real path: `curl https://carolinafutons.com/api/products/<known-slug>` and verify the response body contains live catalog data.

## Positive observations

- §0 scope-at-a-glance is excellent — clear retire/keep table prevents scope creep into cf-xe2.
- §1.3 "what unpublish does (and doesn't)" prevents the most likely operator confusion.
- §1.4 + §4.5 risk tables surface the real failure modes (Velo quota changes, SEO regression, plan-feature mismatch).
- §6 sign-off gate is correctly designed — pre-conditions, per-step approvals, and post-execution sign-offs are separated.
- §1.5 rollback is genuinely one-click and reversible — low blast radius.
- §3.3 explicitly excludes cf-xe2 scope, preventing creep.
- The pricing re-verification flag in §4.2 is the right caution given pricing drift since August 2025.

## Bottom line

The doc is solid as a **decision + sign-off** artifact. The §8 smoke-test playbook is the weakest section: it covers ~10% of the live integration surface (1 of 22 cron jobs, 1 of 81 HTTP functions, 1 of 14 event handlers, 0 of 3 Wix Secrets Manager-backed flows other than mailer). Before §6 sign-off, §8 should be expanded to at least cover one probe per row of the §3.1 active-dependencies table — and `jobs.config` should be added to §3.1 as its own row.

Files reviewed:
- /Users/hal/gt/cfutons/docs/cf-3qt-9-wix-retirement-checklist.md (PR diff)
- /Users/hal/gt/cfutons/src/backend/events.js
- /Users/hal/gt/cfutons/src/backend/http-functions.js
- /Users/hal/gt/cfutons/src/backend/jobs.config

DreadPirateRobertz:
## Review-fold commit eb11d877 — fold-summary

Folded all CRITICAL + HIGH findings from the 5-agent review into one consolidated commit. Doc grew 348 → 460 lines (~30% increase, mostly in §8 smoke tests + §6 sign-off + §1 risk/rollback).

### Critical findings folded

| Finding | Reviewer | Section fixed |
|---|---|---|
| Rollback NOT one-click — 3 hidden failure modes | silent-failure-hunter (C1) | §1.5 rewritten — drift, plan-ordering, A/B/scheduled state loss; 7-day soak ordering |
| §6.2 self-approval bypass clause | silent-failure-hunter (C2) | §6.2 — "No box may be self-approved" |
| §8 smoke #2 accepting 401 as success | silent-failure-hunter (C3) | §8.2 — explicit "no 'or' outcomes" rule |
| `/_functions/wishlistService` is a webMethod, not HTTP route | code-reviewer | §8.2 — `/_functions/health` (real `get_health` export) |
| `/api/healthcheck` doesn't exist in cfw | code-reviewer | §1.1 + §8.5 — removed; replaced with `/api/products` reachability |
| `docs/SITE-OWNER-GUIDE.md` + `TESTING-GUIDE.md` don't exist | comment-analyzer | §3.2 + §5.3 — `docs/brenda-admin-guide.md` + `docs/TESTING-PROCEDURE.md` (verified extant). Pre-bake link verification step added. |
| jobs.config (22 cron jobs) silently load-bearing, not in §3.1 | pr-test-analyzer | §3.1 added line item; §8.3 dedicated cron heartbeat probes |

### High findings folded

| Finding | Section fixed |
|---|---|
| `server: Vercel` header bypassable | §1.1 — `dig` + `x-vercel-id` two-signal origin check |
| "zero organic traffic on Wix-rendered URLs" unfalsifiable | §1.1 — Search Console index-list diff vs `/sitemap.xml` |
| SEO regression Medium → High with concrete mitigation | §1.4 |
| Forms may have been Wix-rendered (handler decoupled, markup not) | §1.4 + §6.1 form-rendering audit |
| Velo quota "test if uncertain" escape hatch | §4.4 — Medium → High, escape removed |
| Velo events other than `onOrderCreated` not exercised | §8.4 — full lifecycle (Created/Approved/FulfillmentCreated/Fulfilled/Canceled/AbandonedCheckoutCreated) |
| `/_functions` POST envelope shape | §8.2 — `{method, args}` |
| CORS preflight not verified | §8.2 — preflight curl |
| Wix Secrets Manager listed but not probed | §8.7 — WWEX/Turnstile/mailer indirect probes |
| No CMS write probe | §8.8 |
| No rollback dry-run rehearsal | §8.9 — pre-unpublish maintenance window |

### Important findings folded

- Header 24h vs §6.1 30-day inconsistency reconciled (§6 wins clause + 30-day primary in header)
- Staging-unpublish gating contradiction across §0/§2.4/§6.2 resolved (sign-off ceremony, no self-approval)
- §3.1 cfw paths corrected — `src/` prefix; `lib/wix-cms.ts` → "reads via Wix Headless SDK"; dispatcher path `src/lib/wix/velo-client.ts`
- §5.5 standing-order removal gated on ≥7 days post-unpublish (premature removal blocks rollback context)

### Removed (per comment-analyzer)

- §1.4 "Wix dashboard login breaks — Likelihood: None" — not a real risk row
- §4.6 redundant approval gate (now flows through §6.2)

### Deferred (low-signal)

code-simplifier's table-merging suggestions (§4.2+§4.3 partially merged; §6.3↔§8 cross-reference tightened; §0 scope-table left as-is for at-a-glance value).

PR #1215 (cf-quba.fu1) had 5/5 reviews close earlier with all factual fixes also pushed in commit 5a5a4811.

PR #1253, opened 2026-05-10
feat(cf-uwfw): queueWelcomeEmail + queueCartRecovery HTTP dispatchers

DreadPirateRobertz:
## Summary

Closes the cf-vtx5-class gap that left cfw's `/api/email/trigger` returning Velo errors against the real backend. cfw does:

```js
callVelo({ method: "queueWelcomeEmail" | "queueCartRecovery", args: [payload] })
```

→ POST `/_functions/<method>`. Wix doesn't auto-route those names to a webMethod — needs an explicit `post_<name>(request)` wrapper. Same pattern as cf-vtx5 (referralService, wishlistService, surveyService, etc.).

Pairs with **stage3-velo PR #36** for the Velo deploy mirror.

## Implementation

- **`src/backend/emailFlow.web.js`** — new file, `Permissions.Anyone` webMethods (cfw calls server-side without a Wix session)
  - `queueWelcomeEmail({type, email})`: validates email; resolves Wix contact via `contacts.appendOrCreateContact` (idempotent — Wix dedupes by email); delegates to `triggerWelcomeSequence(contactId, email, '')` (existing welcome-sequence trigger in `emailAutomation.web.js`).
  - `queueCartRecovery({type, items})`: validates payload shape; returns `success: true, acknowledged: N, note: …`. cfw payload has no email/session, so the actual cart-recovery sequence stays driven by `triggerAbandonedCartRecovery` cron over the Wix Stores `AbandonedCarts` collection (email captured at checkout). This endpoint is an explicit ACK that lets cfw's `/api/email/trigger` return 200 cleanly and serves as a future hook point.

- **`src/backend/http-functions.js`**:
  - `post_queueWelcomeEmail` / `options_queueWelcomeEmail`
  - `post_queueCartRecovery` / `options_queueCartRecovery`
  - `_veloSingleMethodDispatch` helper (variant of `_veloDispatch` for endpoints that don't need multi-method path routing)

- **Error strings** phrased to match the cf-yvs4 + cf-mgnh soft-fail classifier:
  - `"Invalid …"` → 400 (validation)
  - `"Failed to …"` → 503 (infra)
  - business-logic outcomes (no `error` string) → null → 200 (e.g. unsubscribed, already-queued)

## Tests (34 total, all passing)

- `tests/queueEmailDispatcher.test.js` (15) — cf-vtx5 dispatcher contract: invalid_json, args_must_be_array, soft-fail mapping (400 + 503 + 200), 500 + errorId with label in log, options preflight, payload pass-through.
- `tests/emailFlow.test.js` (19) — webMethod-level: payload validation, email normalization (trim + lowercase), contact reuse + creation, CRM-failure paths (throw + empty contactId), success/soft-fail envelope passthrough, every `queueCartRecovery` validation arm.

## Test plan
- [x] `vitest run tests/queueEmailDispatcher.test.js tests/emailFlow.test.js` → 34/34 pass
- [x] stage3-velo mirror PR open
- [ ] After both merge: cfw `e2e/email-triggers.spec.ts` passes against real Velo backend (`NEXT_PUBLIC_USE_FIXTURE_PRODUCTS=0`)

### DIFF
--- a/src/backend/emailFlow.web.js
+++ b/src/backend/emailFlow.web.js
@@ -1,0 +1,137 @@
+/**
+ * @module emailFlow.web
+ * @description Frontend-facing email-trigger entry points.
+ *
+ * cf-uwfw (cf-7ozz.1): cfw's `/api/email/trigger` route does
+ * `callVelo({method: 'queueWelcomeEmail' | 'queueCartRecovery', args: [payload]})`.
+ * Wix Velo doesn't auto-route /_functions/<name> to a webMethod — http-functions.js
+ * carries explicit `post_<name>` wrappers that delegate to these
+ * webMethods.
+ *
+ * Both methods are `Permissions.Anyone` because cfw calls them server-side from
+ * the /api/email/trigger Vercel route, which has no Wix session. Validation
+ * runs inline; rate limiting + de-duplication happen in the underlying
+ * implementations (`triggerWelcomeSequence` for welcome; the cron-driven
+ * `triggerAbandonedCartRecovery` for cart recovery).
+ *
+ * @requires wix-web-module
+ * @requires wix-crm-backend
+ */
+
+import { Permissions, webMethod } from 'wix-web-module';
+import { contacts } from 'wix-crm-backend';
+import { sanitize, validateEmail } from 'backend/utils/sanitize';
+import { triggerWelcomeSequence } from 'backend/emailAutomation.web';
+
+/**
+ * Queue a welcome-email sequence for a freshly-registered cfw user.
+ *
+ * cfw fires this from `/signup` after `/api/auth/register` returns
+ * `{ok: true}`. The payload is `{type: 'welcome', email}`.
+ * No firstName is sent — cfw doesn't have one at signup. The underlying
+ * triggerWelcomeSequence handles missing firstName by emitting an empty
+ * string into the template.
+ *
+ * @function queueWelcomeEmail
+ * @param {{type: 'welcome', email: string}} payload
+ * @returns {Promise<{success: boolean, error?: string, queued?: number}>}
+ *   - `{success: true, queued: N}` when N steps were queued (welcome sequence
+ *     is currently 3 steps).
+ *   - `{success: false, error: 'invalid_payload' | 'invalid_email' | 'unsubscribed' | 'already_queued' | 'contact_resolution_failed'}`
+ *     for the documented soft-fail cases (cf-yvs4 maps these to 4xx).
+ *   - `{success: false, error: <free-text>}` for other backend failures —
+ *     dispatcher catches and emits 500 + errorId.
+ * @permission Anyone — public surface; called from cfw without a session.
+ */
+export const queueWelcomeEmail = webMethod(
+  Permissions.Anyone,
+  async (payload) => {
+    // Error strings are phrased to land in the cf-yvs4/cf-mgnh
+    // soft-fail classifier buckets used by the dispatcher:
+    //   "Invalid …"          → 400 (validation)
+    //   "Failed to …"        → 503 (infra)
+    //   business-logic       → null → 200 (e.g. unsubscribed)
+    if (!payload || typeof payload !== 'object' || payload.type !== 'welcome') {
+      return { success: false, error: 'Invalid payload' };
+    }
+    const email = typeof payload.email === 'string' ? payload.email.trim().toLowerCase() : '';
+    if (!validateEmail(email)) {
+      return { success: false, error: 'Invalid email' };
+    }
+
+    let contactId = '';
+    try {
+      const result = await contacts.appendOrCreateContact({
+        emails: [{ email }],
+      });
+      contactId = result?.contactId || result?._id || '';
+    } catch (err) {
+      console.error('[emailFlow] queueWelcomeEmail contact resolution failed:', err?.message ?? err);
+      return { success: false, error: 'Failed to resolve contact' };
+    }
+    if (!contactId) {
+      return { success: false, error: 'Failed to resolve contact' };
+    }
+
+    // Delegate to the documented welcome-sequence trigger. firstName is
+    // intentionally empty — cfw doesn't pass one.
+    return triggerWelcomeSequence(contactId, email, '');
+  }
+);
+
+/**
+ * Acknowledge a cart-recovery hint from cfw.
+ *
+ * cfw fires this from CartAbandonmentTracker on visibilitychange/pagehide
+ * with `{type: 'cart-recovery', items: [{productId, quantity}]}`. The payload
+ * does NOT include an email or session identifier — at hint time the user
+ * may be anonymous.
+ *
+ * The actual abandoned-cart email sequence is driven by the
+ * `triggerAbandonedCartRecovery` cron job, which scans the
+ * `Wix Stores → AbandonedCarts` collection that Wix Stores populates
+ * automatically once a user begins checkout (where email IS captured).
+ * This webMethod is therefore an explicit ACK contract — it lets cfw's
+ * /api/email/trigger return 200 without a Velo error, and serves as a
+ * future hook point for a session-aware extension.
+ *
+ * Validates payload shape so a malformed cfw call is caught at the boundary,
+ * not silently dropped.
+ *
+ * @function queueCartRecovery
+ * @param {{type: 'cart-recovery', items: Array<{productId: string, quantity: number}>}} payload
+ * @returns {Promise<{success: boolean, acknowledged?: number, note?: string, error?: string}>}
+ * @permission Anyone — public surface; called from cfw without a session.
+ */
+export const queueCartRecovery = webMethod(
+  Permissions.Anyone,
+  async (payload) => {
+    if (!payload || typeof payload !== 'object' || payload.type !== 'cart-recovery') {
+      return { success: false, error: 'Invalid payload' };
+    }
+    if (!Array.isArray(payload.items) || payload.items.length === 0) {
+      return { success: false, error: 'items is required' };
+    }
+    // Defensive shape check — cfw schema guarantees this but fail loudly if
+    // a future schema drift sends garbage.
+    const validItems = payload.items.every(
+      (it) =>
+        it
+        && typeof it === 'object'
+        && typeof it.productId === 'string'
+        && it.productId.length > 0
+        && typeof it.quantity === 'number'
+        && Number.isFinite(it.quantity)
+        && it.quantity > 0
+    );
+    if (!validItems) {
+      return { success: false, error: 'Invalid items' };
+    }
+
+    return {
+      success: true,
+      acknowledged: payload.items.length,
+      note: 'cart-recovery emails are queued by the triggerAbandonedCartRecovery cron from the Wix Stores AbandonedCarts collection once email is captured at checkout',
+    };
+  }
+);
--- a/src/backend/emailFlow.web.js
+++ b/src/backend/emailFlow.web.js
@@ -20,17 +20,20 @@
 
 import { Permissions, webMethod } from 'wix-web-module';
 import { contacts } from 'wix-crm-backend';
-import { sanitize, validateEmail } from 'backend/utils/sanitize';
+import { validateEmail } from 'backend/utils/sanitize';
 import { triggerWelcomeSequence } from 'backend/emailAutomation.web';
 
 /**
  * Queue a welcome-email sequence for a freshly-registered cfw user.
  *
  * cfw fires this from `/signup` after `/api/auth/register` returns
- * `{ok: true}`. The payload is `{type: 'welcome', email}`.
- * No firstName is sent — cfw doesn't have one at signup. The underlying
- * triggerWelcomeSequence handles missing firstName by emitting an empty
- * string into the template.
+ * `{ok: true, redirectTo: string}` (cfw `SignUpForm.tsx`). Payload is
+ * `{type: 'welcome', email}`. No firstName is sent — cfw doesn't have one
+ * at signup. The underlying triggerWelcomeSequence handles missing
+ * firstName by emitting an empty string into the template.
+ *
+ * The welcome sequence currently queues 5 steps (immediate / Day 2 / Day 5
+ * / Day 14 / Day 21 — see SEQUENCES.welcome.steps in emailAutomation.web.js).
  *
  * @function queueWelcomeEmail
  * @param {{type: 'welcome', email: string}} payload
@@ -70,6 +73,10 @@ export const queueWelcomeEmail = webMethod(
       return { success: false, error: 'Failed to resolve contact' };
     }
     if (!contactId) {
+      // Wix CRM returned an envelope without contactId/_id. Log so a future
+      // SDK shape rename surfaces in monitoring instead of routing every
+      // welcome to a 503 with no diagnostic.
+      console.error('[emailFlow] queueWelcomeEmail contact resolution returned empty', { email });
       return { success: false, error: 'Failed to resolve contact' };
     }
 
@@ -88,12 +95,14 @@ export const queueWelcomeEmail = webMethod(
  * may be anonymous.
  *
  * The actual abandoned-cart email sequence is driven by the
- * `triggerAbandonedCartRecovery` cron job, which scans the
- * `Wix Stores → AbandonedCarts` collection that Wix Stores populates
- * automatically once a user begins checkout (where email IS captured).
- * This webMethod is therefore an explicit ACK contract — it lets cfw's
- * /api/email/trigger return 200 without a Velo error, and serves as a
- * future hook point for a session-aware extension.
+ * `triggerAbandonedCartRecovery` cron job, which scans the custom Velo
+ * `AbandonedCarts` CMS collection. That collection is populated by the
+ * `wixEcom_onAbandonedCheckoutCreated` handler in `events.js` from Wix
+ * eCommerce platform events (NOT auto-populated by Wix Stores) — which
+ * fires once a buyer has progressed far enough into checkout for Wix to
+ * capture an email. This webMethod is therefore an explicit ACK contract
+ * — it lets cfw's /api/email/trigger return 200 without a Velo error,
+ * and serves as a future hook point for a session-aware extension.
  *
  * Validates payload shape so a malformed cfw call is caught at the boundary,
  * not silently dropped.
--- a/src/backend/http-functions.js
+++ b/src/backend/http-functions.js
@@ -10,6 +10,7 @@ import { recordPriceSnapshots, checkWishlistAlerts } from 'backend/notificationS
 import { sendEmail } from 'backend/emailService.web';
 import { triggerBrowseRecovery } from 'backend/browseAbandonment.web';
 import { triggerAbandonedCartRecovery, processEmailQueue, triggerReengagement, triggerPostPurchaseSequence, getCampaignAnalytics, unsubscribeContact } from 'backend/emailAutomation.web';
+import { queueWelcomeEmail, queueCartRecovery } from 'backend/emailFlow.web';
 import { scanAndTriggerWinback, runReviewRequestEmails } from 'backend/marketingSequences.web';
 import { processContentSchedule } from 'backend/contentScheduler.web';
 import { sendWeeklyBlogDigest } from 'backend/blogDigestService.web';
@@ -3815,3 +3816,71 @@ export async function post_referralService(request) {
   return _veloDispatch(request, _REFERRAL_METHODS, 'referralService');
 }
 export function options_referralService(request) { return response(corsPreflight(request)); }
+
+// ── cf-uwfw (cf-7ozz.1): cfw email-trigger endpoints ─────────────────────────
+//
+// cfw's `/api/email/trigger` route does:
+//   callVelo({method: 'queueWelcomeEmail',  args: [{type: 'welcome', email}]})
+//   callVelo({method: 'queueCartRecovery',  args: [{type: 'cart-recovery', items}]})
+//
+// Wix Velo doesn't auto-route /_functions/<name> to a webMethod, so each name
+// needs an explicit `post_<name>(request)` wrapper. Both delegate to the
+// `backend/emailFlow.web` webMethods of the same name. Permissions on the
+// underlying methods are `Anyone` since cfw calls server-side without a Wix
+// session.
+
+async function _veloSingleMethodDispatch(request, fn, label) {
+  const JSON_HEADERS = corsHeaders(request, { 'Content-Type': 'application/json', 'Cache-Control': 'no-store' });
+
+  let body;
+  try {
+    body = await request.body.json();
+  } catch (_) {
+    return badRequest({
+      body: JSON.stringify({ success: false, error: 'invalid_json' }),
+      headers: JSON_HEADERS,
+    });
+  }
+
+  const args = Array.isArray(body?.args) ? body.args : null;
+  if (args === null) {
+    return badRequest({
+      body: JSON.stringify({ success: false, error: 'args_must_be_array' }),
+      headers: JSON_HEADERS,
+    });
+  }
+
+  try {
+    const result = await fn(...args);
+    // cf-yvs4 lying-status fix: a {success:false} envelope with a known soft-fail
+    // marker (auth/rate/notfound/infra/validation) maps to a real 4xx so cfw's
+    // velo-client throws VeloRpcError. Business-logic outcomes (status===null)
+    // stay 200 so cfw can branch on body.success.
+    if (result && typeof result === 'object' && result.success === false) {
+      const status = _veloDispatchSoftFailStatus(result.error);
+      if (status !== null) {
+        return response({ status, body: JSON.stringify(result), headers: JSON_HEADERS });
+      }
+    }
+    return ok({ body: JSON.stringify(result), headers: JSON_HEADERS });
+  } catch (err) {
+    const errorId = (typeof crypto !== 'undefined' && crypto.randomUUID)
+      ? crypto.randomUUID()
+      : `qe-${Date.now()}-${Math.random().toString(36).slice(2, 10)}`;
+    console.error(`HTTP function error (post_${label}) errorId=${errorId}:`, err);
+    return serverError({
+      body: JSON.stringify({ success: false, error: 'server_error', errorId }),
+      headers: JSON_HEADERS,
+    });
+  }
+}
+
+export async function post_queueWelcomeEmail(request) {
+  return _veloSingleMethodDispatch(request, queueWelcomeEmail, 'queueWelcomeEmail');
+}
+export function options_queueWelcomeEmail(request) { return response(corsPreflight(request)); }
+
+export async function post_queueCartRecovery(request) {
+  return _veloSingleMethodDispatch(request, queueCartRecovery, 'queueCartRecovery');
+}
+export function options_queueCartRecovery(request) { return response(corsPreflight(request)); }
--- a/src/backend/http-functions.js
+++ b/src/backend/http-functions.js
@@ -3862,11 +3862,11 @@ async function _veloSingleMethodDispatch(request, fn, label) {
         return response({ status, body: JSON.stringify(result), headers: JSON_HEADERS });
       }
     }
-    return ok({ body: JSON.stringify(result), headers: JSON_HEADERS });
+    // Guard against a wrapped fn that returns undefined — match _veloDispatch's
+    // serialization shape so callers always get a JSON body.
+    return ok({ body: JSON.stringify(result ?? null), headers: JSON_HEADERS });
   } catch (err) {
-    const errorId = (typeof crypto !== 'undefined' && crypto.randomUUID)
-      ? crypto.randomUUID()
-      : `qe-${Date.now()}-${Math.random().toString(36).slice(2, 10)}`;
+    const errorId = _veloDispatchErrorId();
     console.error(`HTTP function error (post_${label}) errorId=${errorId}:`, err);
     return serverError({
       body: JSON.stringify({ success: false, error: 'server_error', errorId }),
--- a/tests/emailFlow.test.js
+++ b/tests/emailFlow.test.js
@@ -70,10 +70,28 @@ describe('queueWelcomeEmail', () => {
     expect(vi.mocked(triggerWelcomeSequence)).toHaveBeenCalledWith('contact-existing-77', 'existing@example.com', '');
   });
 
+  it('falls back to result._id when CRM omits contactId (legacy SDK shape)', async () => {
+    // pr-test-analyzer #1: cover the `result?.contactId || result?._id`
+    // fallback. The default mock always returns {contactId:...} so the _id
+    // arm was previously untested.
+    const crm = await import('./__mocks__/wix-crm-backend.js');
+    const spy = vi.spyOn(crm.contacts, 'appendOrCreateContact')
+      .mockResolvedValueOnce({ _id: 'legacy-contact-001' });
+    vi.mocked(triggerWelcomeSequence).mockResolvedValue({ success: true, queued: 5 });
+    try {
+      const res = await queueWelcomeEmail({ type: 'welcome', email: 'legacy@example.com' });
+      expect(res).toEqual({ success: true, queued: 5 });
+      expect(vi.mocked(triggerWelcomeSequence))
+        .toHaveBeenCalledWith('legacy-contact-001', 'legacy@example.com', '');
+    } finally {
+      spy.mockRestore();
+    }
+  });
+
   it('returns "Failed to resolve contact" infra error when contact lookup throws', async () => {
     const crm = await import('./__mocks__/wix-crm-backend.js');
-    const real = crm.contacts.appendOrCreateContact;
-    crm.contacts.appendOrCreateContact = async () => { throw new Error('CRM down'); };
+    const spy = vi.spyOn(crm.contacts, 'appendOrCreateContact')
+      .mockRejectedValueOnce(new Error('CRM down'));
     const errSpy = vi.spyOn(console, 'error').mockImplementation(() => {});
     try {
       const res = await queueWelcomeEmail({ type: 'welcome', email: 'good@example.com' });
@@ -81,21 +99,33 @@ describe('queueWelcomeEmail', () => {
       expect(errSpy).toHaveBeenCalled();
       expect(vi.mocked(triggerWelcomeSequence)).not.toHaveBeenCalled();
     } finally {
-      crm.contacts.appendOrCreateContact = real;
+      spy.mockRestore();
       errSpy.mockRestore();
     }
   });
 
-  it('returns "Failed to resolve contact" when CRM yields no contactId', async () => {
+  it('returns "Failed to resolve contact" when CRM yields no contactId AND logs the empty envelope', async () => {
     const crm = await import('./__mocks__/wix-crm-backend.js');
-    const real = crm.contacts.appendOrCreateContact;
-    crm.contacts.appendOrCreateContact = async () => ({ /* no id */ });
+    const spy = vi.spyOn(crm.contacts, 'appendOrCreateContact')
+      .mockResolvedValueOnce({ /* neither contactId nor _id */ });
+    const errSpy = vi.spyOn(console, 'error').mockImplementation(() => {});
     try {
-      const res = await queueWelcomeEmail({ type: 'welcome', email: 'good@example.com' });
+      const res = await queueWelcomeEmail({ type: 'welcome', email: 'empty@example.com' });
       expect(res).toEqual({ success: false, error: 'Failed to resolve contact' });
+      // Without a console.error here a future SDK rename routes everyone to
+      // 503 with no diagnostic — silent-failure-hunter MEDIUM.
+      expect(errSpy).toHaveBeenCalled();
+      // Production logs as `console.error('...empty', { email })`. Inspect
+      // the structured second arg so a future log-shape change is caught.
+      const matchingCall = errSpy.mock.calls.find(
+        (args) => typeof args[0] === 'string' && args[0].includes('returned empty')
+      );
+      expect(matchingCall).toBeDefined();
+      expect(matchingCall[1]).toMatchObject({ email: 'empty@example.com' });
       expect(vi.mocked(triggerWelcomeSequence)).not.toHaveBeenCalled();
     } finally {
-      crm.contacts.appendOrCreateContact = real;
+      spy.mockRestore();
+      errSpy.mockRestore();
     }
   });
 
--- a/tests/emailFlow.test.js
+++ b/tests/emailFlow.test.js
@@ -1,0 +1,176 @@
+/**
+ * @file emailFlow.test.js
+ * @description cf-uwfw — coverage for queueWelcomeEmail / queueCartRecovery
+ * webMethods in src/backend/emailFlow.web.js. The HTTP dispatcher contract
+ * is covered separately in queueEmailDispatcher.test.js; this file pins the
+ * webMethod-level behaviour (validation, contact resolution, delegation).
+ */
+
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+
+vi.mock('backend/emailAutomation.web', () => ({
+  triggerWelcomeSequence: vi.fn(),
+}));
+
+import { queueWelcomeEmail, queueCartRecovery } from '../src/backend/emailFlow.web.js';
+import { __reset, __seedContacts } from './__mocks__/wix-crm-backend.js';
+import { triggerWelcomeSequence } from 'backend/emailAutomation.web';
+
+beforeEach(() => {
+  __reset();
+  vi.clearAllMocks();
+});
+
+// ── queueWelcomeEmail ────────────────────────────────────────────────────────
+
+describe('queueWelcomeEmail', () => {
+  it('rejects missing payload', async () => {
+    const res = await queueWelcomeEmail(undefined);
+    expect(res).toEqual({ success: false, error: 'Invalid payload' });
+    expect(vi.mocked(triggerWelcomeSequence)).not.toHaveBeenCalled();
+  });
+
+  it('rejects payload with wrong type', async () => {
+    const res = await queueWelcomeEmail({ type: 'something-else', email: 'a@b.com' });
+    expect(res).toEqual({ success: false, error: 'Invalid payload' });
+  });
+
+  it('rejects non-object payload', async () => {
+    const res = await queueWelcomeEmail('not-an-object');
+    expect(res).toEqual({ success: false, error: 'Invalid payload' });
+  });
+
+  it('rejects missing email', async () => {
+    const res = await queueWelcomeEmail({ type: 'welcome' });
+    expect(res).toEqual({ success: false, error: 'Invalid email' });
+  });
+
+  it('rejects malformed email', async () => {
+    const res = await queueWelcomeEmail({ type: 'welcome', email: 'not-an-email' });
+    expect(res).toEqual({ success: false, error: 'Invalid email' });
+  });
+
+  it('lowercases + trims email before contact resolution', async () => {
+    vi.mocked(triggerWelcomeSequence).mockResolvedValue({ success: true, queued: 3 });
+    await queueWelcomeEmail({ type: 'welcome', email: '  ALICE@Example.COM  ' });
+    // The mocked appendOrCreateContact stamps a new id; we just assert
+    // the trigger was called with the normalized email.
+    expect(vi.mocked(triggerWelcomeSequence)).toHaveBeenCalledTimes(1);
+    const [, email, firstName] = vi.mocked(triggerWelcomeSequence).mock.calls[0];
+    expect(email).toBe('alice@example.com');
+    expect(firstName).toBe('');
+  });
+
+  it('reuses existing contact when email already in CRM', async () => {
+    __seedContacts([
+      { _id: 'contact-existing-77', primaryInfo: { email: 'existing@example.com' } },
+    ]);
+    vi.mocked(triggerWelcomeSequence).mockResolvedValue({ success: true, queued: 3 });
+    await queueWelcomeEmail({ type: 'welcome', email: 'existing@example.com' });
+    expect(vi.mocked(triggerWelcomeSequence)).toHaveBeenCalledWith('contact-existing-77', 'existing@example.com', '');
+  });
+
+  it('returns "Failed to resolve contact" infra error when contact lookup throws', async () => {
+    const crm = await import('./__mocks__/wix-crm-backend.js');
+    const real = crm.contacts.appendOrCreateContact;
+    crm.contacts.appendOrCreateContact = async () => { throw new Error('CRM down'); };
+    const errSpy = vi.spyOn(console, 'error').mockImplementation(() => {});
+    try {
+      const res = await queueWelcomeEmail({ type: 'welcome', email: 'good@example.com' });
+      expect(res).toEqual({ success: false, error: 'Failed to resolve contact' });
+      expect(errSpy).toHaveBeenCalled();
+      expect(vi.mocked(triggerWelcomeSequence)).not.toHaveBeenCalled();
+    } finally {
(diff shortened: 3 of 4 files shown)


DreadPirateRobertz:
## Review by code-simplifier (dispatched by godfrey for cf-uwfw)

Net assessment: small, well-scoped PR. Per-bead clarity is good — I won't push to consolidate `_veloSingleMethodDispatch` and `_veloDispatch` into a registry (legitimately different shape: one needs `request.path[0]` resolution, the other doesn't, and the file's existing convention favors explicit per-bead handlers à la `post_referralService` / `post_wishlistService`).

A few small simplifications below — none are blocking.

---

### 1. Unused import in `emailFlow.web.js`

```js
import { sanitize, validateEmail } from 'backend/utils/sanitize';
```

`sanitize` is imported but never referenced. The downstream `triggerWelcomeSequence` does its own `sanitize(email, 254)` + `sanitize(firstName, 200)` + `sanitize(contactId, 50)` (see `emailAutomation.web.js:379-383`), so this module legitimately doesn't need it. Drop to:

```js
import { validateEmail } from 'backend/utils/sanitize';
```

### 2. Three small divergences from `_veloDispatch` worth aligning

`_veloSingleMethodDispatch` and `_veloDispatch` should not be merged (different routing shape, file convention is per-bead handlers), but the new helper re-implements three things slightly differently from the established sibling that lives ~470 lines above it:

**a. errorId generation is inlined** — there's already a shared helper:

```js
// existing, line 3263
function _veloDispatchErrorId() { … }
```

The new helper inlines an equivalent expression. Replacing the inline block with `_veloDispatchErrorId()` removes ~3 lines and means a future change to the id format propagates to both helpers automatically.

**b. `ok({ body: JSON.stringify(result) })` vs sibling's `JSON.stringify(result == null ? null : result)`** — the sibling guards against `JSON.stringify(undefined)` returning the literal string `"undefined"` which isn't valid JSON. The new helper omits this. Both `queueWelcomeEmail` and `queueCartRecovery` always return an object today, so it doesn't bite — but matching the sibling guard is one character (`result ?? null`) and keeps the contract uniform.

**c. Catch parameter** — existing `_veloDispatch` uses `catch (err)`; new helper uses `catch (_)`. Trivially: the existing form is fine and matches the file.

### 3. `queueCartRecovery` defensive-without-cause shape check

The `validItems` `.every(...)` chain (lines 123–132) checks six things per item:

```js
it
  && typeof it === 'object'
  && typeof it.productId === 'string'
  && it.productId.length > 0
  && typeof it.quantity === 'number'
  && Number.isFinite(it.quantity)
  && it.quantity > 0
```

Given:
- the cfw schema (`/api/email/trigger`) already shape-validates this payload before `callVelo`, and
- the comment above explicitly acknowledges "cfw schema guarantees this but fail loudly if a future schema drift sends garbage",

this is reasonable defense-in-depth. **But** since the body is never used downstream — `queueCartRecovery` returns `{ success: true, acknowledged: payload.items.length, note }` and the actual cart-recovery sequence is driven by the cron over `AbandonedCarts` — the `productId.length > 0`, `Number.isFinite`, and `quantity > 0` checks aren't preventing a downstream bug. They're shape-asserting data that's about to be discarded.

You could collapse to:

```js
const validItems = payload.items.every(
  (it) => it && typeof it.productId === 'string' && typeof it.quantity === 'number'
);
```

…and lose nothing functional. **Optional** — if you want the loud validation as a contract pin for the future "session-aware extension" mentioned in the JSDoc, leave it. Just flagging that today it's pure ceremony.

### 4. Test redundancy across the two files (low-impact)

Some validation paths are exercised in both `tests/emailFlow.test.js` and `tests/queueEmailDispatcher.test.js`:

- `emailFlow.test.js` "rejects missing email" / "rejects malformed email" → produces `{success:false, error:'Invalid email'}`
- `queueEmailDispatcher.test.js` "400 maps 'Invalid email' soft-fail" → mocks `queueWelcomeEmail` to return that envelope

The dispatcher test mocks the webMethod, so it's not actually re-exercising the real validation path — it's verifying the `Invalid …` → 400 mapping. That's the right boundary, and the headers/comments in both files make the split clear. **Not redundant** in the harmful sense; the file split is correct.

The genuine duplication: `emailFlow.test.js` has 9 `queueWelcomeEmail` validation tests + 8 `queueCartRecovery` validation tests; `queueEmailDispatcher.test.js` has 7 + 7 dispatcher tests. The dispatcher tests for `queueCartRecovery` mock `queueCartRecovery` to return specific envelopes, so they don't re-cover validation — they cover mapping. This is the right boundary. Keep both.

### 5. JSDoc in `emailFlow.web.js` is verbose but earns its keep

The big `@description` block (lines 7–25) is on the long side, but it documents:
- why this file exists (cf-uwfw + cfw `callVelo` shape),
- why both methods are `Permissions.Anyone`,
- where rate-limiting + dedup actually happen (delegation — not in this file),
- the relationship with `triggerAbandonedCartRecovery` cron.

That's why-context, not what-context. Keep it.

The inline comment block at lines 55–59 inside `queueWelcomeEmail` re-states the soft-fail taxonomy that's already documented at `_veloDispatchSoftFailStatus` (lines 3270–3293 of `http-functions.js`). Consider trimming to one line:

```js
// Error strings phrased for cf-yvs4 / cf-mgnh classifier — see _veloDispatchSoftFailStatus.
```

Same for the comment at lines 161–171 of `http-functions.js`: it's clear, but the cf-uwfw context is already in the PR body and at the top of `emailFlow.web.js`. A 2-line comment pointing to `emailFlow.web.js` would suffice. Optional.

### 6. Minor — `note` string is long

The 156-character `note` in the success envelope (lines 138–140) is helpful for debugging in cfw logs, but it's also baked into every successful response. A constant at module scope would deduplicate if the call ever moves elsewhere:

```js
const CART_RECOVERY_ACK_NOTE =
  'cart-recovery emails are queued by triggerAbandonedCartRecovery cron from AbandonedCarts collection';
```

Minor — leave as-is is fine.

---

## Summary of suggested changes (non-blocking)

| # | File | Change | Priority |
|---|------|--------|----------|
| 1 | `emailFlow.web.js` | Drop unused `sanitize` import | small |
| 2a | `http-functions.js` | Use `_veloDispatchErrorId()` instead of inlining | small |
| 2b | `http-functions.js` | `result ?? null` guard for parity with `_veloDispatch` | small |
| 3 | `emailFlow.web.js` | Optionally trim `validItems` chain (or keep as future contract) | optional |
| 5 | `emailFlow.web.js` / `http-functions.js` | Trim duplicated soft-fail taxonomy comment to a pointer | optional |

Tests + structure are sound. The PR ships a clean per-bead surface that matches existing cfw-call conventions.

DreadPirateRobertz:
## Review by silent-failure-hunter (dispatched by godfrey for cf-uwfw)

Scope: `src/backend/emailFlow.web.js`, `src/backend/http-functions.js` (cf-uwfw additions), `tests/emailFlow.test.js`, `tests/queueEmailDispatcher.test.js`. Cross-checked against the cfw caller `/Users/hal/gt/carolina-futons-web/src/app/api/email/trigger/route.ts`, `_veloDispatchSoftFailStatus` in `http-functions.js:3294`, and `triggerWelcomeSequence` in `emailAutomation.web.js:373`.

Verdict: **Mergeable with non-blocking caveats.** No silent failure that could mask a real outage. Two HIGH-severity concerns (abuse vector + unsubscribed/dedup observability) are worth fixing before scale, but neither is a regression vs. the current main behavior. The dispatcher contract (validation → 400, infra → 503, business outcome → 200) is correctly threaded end-to-end and the cfw caller already treats `ok:true` as "fire-and-forget acknowledgement," not "email definitely sent."

---

### Concern 1 — `queueCartRecovery` returns `success:true` though no email is queued

Severity: **NOT a silent failure.** Acceptable as designed.

cfw's `route.ts` only inspects whether the Velo call threw `VeloRpcError`. On success it returns `NextResponse.json({ ok: true })`; on Velo error it still returns `{ ok: false, error: "trigger-failed" }` with HTTP 200 ("Non-fatal — email triggers must not block the primary user flow"). The cfw response is *never surfaced to a user as "your cart-recovery email is queued"* — it's a hint endpoint fired from `visibilitychange`/`pagehide`. The `note` field in the response body explicitly documents the contract:

> "cart-recovery emails are queued by the triggerAbandonedCartRecovery cron from the Wix Stores AbandonedCarts collection once email is captured at checkout"

The semantics are: *"acknowledged the hint; the actual cart-recovery is owned by the cron that scans `Wix Stores → AbandonedCarts`."* That is correct because at `pagehide` time the user often has no email yet (anonymous cart). The hint endpoint is genuinely a placeholder/future hook.

**Caveat (MEDIUM):** if telemetry ever starts treating `acknowledged: N` as a queue-depth metric, that will silently mislead. I'd add a one-line comment at the dispatcher site (or rename `acknowledged` → `received` / `noted`) to prevent future confusion. Not blocking.

---

### Concern 2 — `triggerWelcomeSequence` returns `{success:false, queued:0}` for three distinct conditions with no diagnostic

Severity: **HIGH (observability, not silent failure).**

`triggerWelcomeSequence` (emailAutomation.web.js:377-397) collapses three legitimate-but-distinct outcomes into the same envelope:

1. **Empty/invalid email** → `{success:false, queued:0}` (line 377/380)
2. **Unsubscribed** → `{success:false, queued:0}` (line 387)
3. **Already queued (dedup hit)** → `{success:false, queued:0}` (line 397)

`_veloSingleMethodDispatch` correctly returns 200 here (no `error` field → classifier returns `null`), and that part is right per cf-yvs4. **But** debugging "why didn't the welcome email fire for `user@x.com`?" is impossible from the wire — caller, ops, and Sentry all see the same `{success:false, queued:0}`.

This is a pre-existing problem in `triggerWelcomeSequence`, not introduced by this PR. **The wrapper, though, is the natural place to fix it** because cf-uwfw is what makes that envelope user-reachable from cfw for the first time. Two options:

```js
// Option A — surface a reason field without changing the soft-fail classifier
if (await isUnsubscribed(cleanEmail, 'welcome')) {
  return { success: false, queued: 0, reason: 'unsubscribed' };
}
if (existing.items.length > 0) return { success: false, queued: 0, reason: 'already_queued' };

// Option B — log at queueWelcomeEmail level after delegating
const r = await triggerWelcomeSequence(contactId, email, '');
if (!r.success && r.queued === 0) {
  console.info('[emailFlow] queueWelcomeEmail soft-no-op', { email, contactId });
}
return r;
```

Recommend Option A — it's a strict superset and `cf-yvs4`'s soft-fail classifier ignores extra fields, so 200 is preserved. This is **NOT blocking** for cf-uwfw merge; file a follow-up bead.

---

### Concern 3 — `Permissions.Anyone` + no rate limit = welcome-spam vector

Severity: **HIGH (abuse).** Worth a dedicated bead before this surface gets indexed.

`queueWelcomeEmail` is `Permissions.Anyone`. The validation chain is:

1. payload shape (free)
2. `validateEmail(email)` (free)
3. `contacts.appendOrCreateContact({emails:[{email}]})` — **this creates a Wix CRM contact for any well-formed email** (Wix dedupes by email so each unique email creates exactly one contact)
4. `triggerWelcomeSequence` — checks unsubscribe + EmailQueue dedup but **only against this same email**

An attacker can hit `POST /_functions/queueWelcomeEmail` with `{type:'welcome', email:'<victim>@<gmail>.com'}` and:

- Create a Wix CRM contact in our tenant (CRM bloat, possible quota cost)
- Cause us to send a welcome series to a non-customer (spam complaint, IP/domain reputation hit, possible CAN-SPAM exposure since the recipient never opted in)
- The dedup check (`EmailQueue` with `recipientEmail` + `sequenceType:welcome` + `sequenceStep:1`) only blocks **the exact same email a second time** — an attacker can rotate through a list of 10k addresses and queue 10k welcome series.

`emailAutomation.web.js` *does* import `checkRateLimit` (line 41) and uses it on `unsubscribeContact` (max:100) and `recordEmailEvent` — but **not on the welcome trigger.**

**Recommendation (file as follow-up bead, do not block merge):**

```js
import { checkRateLimit } from 'backend/utils/rateLimit';

export const queueWelcomeEmail = webMethod(
  Permissions.Anyone,
  async (payload) => {
    // …existing validation…

    // Per-email rate limit so a single address can't loop the queue,
    // and per-IP would be even better (need request context plumbed in).
    const { allowed } = await checkRateLimit('WelcomeQueueRateLimit', email, {
      max: 1, windowMs: 60 * 60 * 1000  // 1 welcome series per email per hour
    });
    if (!allowed) {
      return { success: false, error: 'Rate limit exceeded' };
    }
    // …
  }
);
```

`"Rate limit exceeded"` will land on 429 via the existing classifier (`http-functions.js:3320-3324`). This is parity with how `unsubscribeContact` and `recordEmailEvent` are protected. Per-IP rate limit is stronger but the request context isn't available inside a webMethod without dispatcher plumbing, so per-email is the practical first cut.

---

### Concern 4 — "Failed to resolve contact" → 503 even when CRM returns empty `contactId`

Severity: **MEDIUM.** The classification is defensible but the error string is misleading.

`emailFlow.web.js:65-72` has two paths to "Failed to resolve contact":

```js
} catch (err) {
  console.error('[emailFlow] queueWelcomeEmail contact resolution failed:', err?.message ?? err);
  return { success: false, error: 'Failed to resolve contact' };
}
if (!contactId) {
  return { success: false, error: 'Failed to resolve contact' };  // ← no log
}
```

**Issue 1 — silent path:** the `!contactId` branch returns 503 with **no console.error**. If Wix CRM ever changes its response shape (e.g. returns `result.id` instead of `result.contactId` / `result._id`), this PR will route everyone to 503 with zero diagnostic. The mock at `tests/__mocks__/wix-crm-backend.js:65-74` returns `{contactId}`, but the production SDK has historically returned both `contactId` (REST) and `_id` (older Velo) — the code already handles both via `result?.contactId || result?._id || ''`. A field rename would be silent.

```js
if (!contactId) {
  console.error('[emailFlow] queueWelcomeEmail: appendOrCreateContact returned no id', { result });
  return { success: false, error: 'Failed to resolve contact' };
}
```

**Issue 2 — 503 vs 422:** an empty `contactId` from a successful CRM call is arguably *not* infra-down — it could be a CRM-side validation gap (e.g. blocked email domain, suppression list). Mapping to 503 makes Vercel/cfw think Wix is unreachable, when really the request was rejected silently. Acceptable for now (you can't tell the two apart from this layer), but log the result body so we can disambiguate from Sentry later.

---

### Concern 5 — Test isolation for the `appendOrCreateContact` monkey-patch

Severity: **MEDIUM (latent flake risk).**

In `tests/emailFlow.test.js:74-100`, two tests do:

```js
const crm = await import('./__mocks__/wix-crm-backend.js');
const real = crm.contacts.appendOrCreateContact;
crm.contacts.appendOrCreateContact = async () => { throw new Error('CRM down'); };
try { … } finally { crm.contacts.appendOrCreateContact = real; }
```

Since `wix-crm-backend.js` is mocked via the vitest alias path (the mock is shared across **every** test file that imports `wix-crm-backend` in the same worker), this monkey-patches the **module-level singleton** for the duration of the test. Concerns:

1. **Cross-file pollution within the same worker.** Vitest workers default to one file per worker (no problem), but if `pool: 'threads'` or `isolate: false` is ever set in `vitest.config.js`, two files can share a module graph. A `try/finally` is correct but **not exception-safe across `await`** if a microtask between the assignment and the `try` rejects — extremely unlikely here (next statement is `try`) but worth knowing. Current code is fine.

2. **Process-level pollution if a test throws inside `try` and the test runner aborts before `finally` runs.** Vitest does run `finally` blocks on assertion failure, so this is fine in practice — but if the test ever uses `expect.poll` or fake timers between the monkey-patch and the `finally`, a hang could leave the patch live. Current code doesn't do that.

3. **Recommended: prefer `vi.spyOn(crm.contacts, 'appendOrCreateContact').mockImplementation(...)`.** Vitest auto-restores spies between tests via `mockReset: true` / `restoreMocks: true` in config (and `vi.restoreAllMocks()` in `afterEach`). This eliminates the manual `try/finally` and is robust against test rewrites:

```js
import { contacts } from 'wix-crm-backend';
…
it('…', async () => {
  vi.spyOn(contacts, 'appendOrCreateContact').mockRejectedValueOnce(new Error('CRM down'));
  const res = await queueWelcomeEmail(…);
  // no finally needed — auto-restored
});
```

I checked `vitest.config.js` is not in the diff — confirm `restoreMocks` or `mockReset` is set globally; if not, the current `try/finally` is actually the safer choice. Either way, **not blocking.**

---

### Bonus findings

- **`emailFlow.web.js:51`** — `import { sanitize, validateEmail } from 'backend/utils/sanitize';` but `sanitize` is **never used** in this module (the email is already trimmed/lowercased inline). Drop the unused import.
- **`emailFlow.web.js:66`** — `console.error('[emailFlow] queueWelcomeEmail contact resolution failed:', err?.message ?? err);` correctly handles `err` being an Error vs a thrown non-Error — good. But consider adding the email (already validated, safe to log) so Sentry can correlate: `console.error('[emailFlow] …', { email, message: err?.message ?? String(err) });`
- **`http-functions.js:_veloSingleMethodDispatch`** — the 500 path (`errorId = crypto.randomUUID() ?? fallback`) is correct and matches cf-vtx5. The fallback `qe-${Date.now()}-${random}` is descriptive enough to grep in logs. Good.
- **`queueCartRecovery` validates `productId.length > 0` and `quantity > 0`** — solid; no off-by-one. The test file covers each arm explicitly.
- **No empty catch blocks. No swallowed errors. No mock fallbacks in production code.** Clean on the silent-failure axis.

---

### Summary

| # | Issue | Severity | Blocking? |
|---|-------|----------|-----------|
| 1 | `queueCartRecovery` returns success without queueing | None — by design, cfw caller treats as ack | No |
| 2 | `{success:false, queued:0}` collapses 3 outcomes | HIGH (observability) | No — follow-up bead |
| 3 | `Permissions.Anyone` + no rate limit = welcome spam vector | HIGH (abuse) | No — follow-up bead, file before scale |
| 4 | "Failed to resolve contact" path missing log on `!contactId` | MEDIUM | One-line fix recommended in this PR |
| 5 | Monkey-patch test isolation | MEDIUM (latent) | No — works, prefer `vi.spyOn` long-term |
| Bonus | Unused `sanitize` import | LOW | One-line cleanup |

**Recommendation:** approve, with a one-line `console.error` added to the `!contactId` branch in `emailFlow.web.js` before merge (Concern 4 issue 1) and follow-up beads for Concerns 2 and 3.

— silent-failure-hunter (dispatched by godfrey for cf-uwfw)

DreadPirateRobertz:
## Review by pr-test-analyzer (dispatched by godfrey for cf-uwfw)

### Summary

Behavioral coverage on `emailFlow.web.js` is excellent (100% statements / 97% branches / 100% functions per local v8 coverage run). The dispatcher contract tests are well-targeted at the cf-vtx5/cf-yvs4 soft-fail mapping. 34/34 pass. **Two real gaps and three quality nits below; nothing blocking.**

Coverage tally for the new module (vitest --coverage, scoped to the new files):

```
File              | % Stmts | % Branch | % Funcs | % Lines | Uncovered
emailFlow.web.js  |   100   |   97.05  |   100   |   100   | line 69 (catch — partial)
```

The single uncovered logical region is the `result?.contactId || result?._id` fallback chain at line 67 — see Critical Gap #1.

---

### Critical Gaps

**1. queueWelcomeEmail contactId fallback chain — `_id` branch untested  (criticality 8/10)**

The fallback at `src/backend/emailFlow.web.js:67`:

```js
contactId = result?.contactId || result?._id || '';
```

Three branches; the test suite covers only two:

| Return shape from `contacts.appendOrCreateContact` | Branch | Covered? |
|---|---|---|
| `{ contactId: 'x' }` | A | Yes — happy-path + reuse tests |
| `{ _id: 'x' }` (no contactId) | B | **No** |
| `{}` / `null` | C | Yes — "yields no contactId" test |

This matters because the Wix CRM SDK historically has shipped both shapes. The Wix Contacts v4 SDK returns `{ contactId }`, but the underlying Wix Data record carries `_id`, and several internal Velo wrappers have leaked through with `_id` (this is exactly why the `|| result?._id` was added). If a future SDK upgrade flips back to `_id`-only — or a mock divergence ever happens — branch B would silently fail with the current mock that always emits `{ contactId }`. Add:

```js
it('falls back to _id when CRM returns the wix-data shape (no contactId field)', async () => {
  const crm = await import('./__mocks__/wix-crm-backend.js');
  const real = crm.contacts.appendOrCreateContact;
  crm.contacts.appendOrCreateContact = async () => ({ _id: 'contact-via-_id-99' });
  vi.mocked(triggerWelcomeSequence).mockResolvedValue({ success: true, queued: 3 });
  try {
    await queueWelcomeEmail({ type: 'welcome', email: 'fallback@example.com' });
    expect(vi.mocked(triggerWelcomeSequence))
      .toHaveBeenCalledWith('contact-via-_id-99', 'fallback@example.com', '');
  } finally {
    crm.contacts.appendOrCreateContact = real;
  }
});
```

This regression would cause every cfw signup welcome email to silently fail with "Failed to resolve contact" — a P1 user-facing outcome, exactly what the dispatcher would map to 503 and surface to cfw, but the bug would be in our wiring, not infra.

**2. CORS preflight tests assert nothing meaningful about CORS  (criticality 7/10)**

`tests/queueEmailDispatcher.test.js:125-129` and `186-190`:

```js
it('options preflight responds', () => {
  const res = options_queueWelcomeEmail({ headers: { origin: goodOrigin } });
  expect(res).toBeDefined();
  expect(res.status).toBeGreaterThanOrEqual(200);
});
```

`corsPreflight` returns **204 with full ACAO headers** for an allowed origin and **403 with no CORS headers** for a disallowed origin. **Both pass `status >= 200`**. So a regression that:

- strips `Access-Control-Allow-Origin`
- breaks the allowlist (disallows the cfw origin)
- drops `Access-Control-Allow-Methods` / `Access-Control-Allow-Headers`

would slip past these tests entirely, and a real CORS regression in production would only show up as cfw browser-fetch failures. Tighten to:

```js
it('options preflight emits 204 + ACAO for allowed origin', () => {
  const res = options_queueWelcomeEmail({ headers: { origin: goodOrigin } });
  expect(res.status).toBe(204);
  expect(res.headers['Access-Control-Allow-Origin']).toBe(goodOrigin);
  expect(res.headers['Access-Control-Allow-Methods']).toContain('POST');
});

it('options preflight rejects disallowed origin with 403', () => {
  const res = options_queueWelcomeEmail({ headers: { origin: 'https://attacker.example' } });
  expect(res.status).toBe(403);
  expect(res.headers['Access-Control-Allow-Origin']).toBeUndefined();
});
```

The current assertion fits the *silent-failure-style* pattern the brief flagged: it would pass even if CORS were entirely broken.

---

### Important Improvements

**3. Sync throw from wrapped fn — implicitly covered, not asserted  (criticality 5/10)**

The dispatcher's `try { const result = await fn(...args); }` correctly catches both async rejections and sync throws (because `await` of a thrown value propagates). All current "500 + errorId" tests use `mockRejectedValue` (async). If someone refactored away the `await` or split the try/catch, sync throws from the webMethod (e.g. an unguarded property access on `payload`) would crash the dispatcher with an unhandled rejection. Cheap, targeted addition:

```js
it('500 + errorId on synchronous throw from wrapped fn', async () => {
  vi.mocked(queueWelcomeEmail).mockImplementation(() => { throw new Error('sync boom'); });
  const consoleErr = vi.spyOn(console, 'error').mockImplementation(() => {});
  const res = await post_queueWelcomeEmail(makeRequest({ args: [{ type: 'welcome', email: 'a@b.com' }] }));
  expect(res.status).toBe(500);
  consoleErr.mockRestore();
});
```

**4. `503 maps "Failed to resolve contact"` test asserts only status  (criticality 4/10)**

`queueEmailDispatcher.test.js:88-94` asserts `res.status === 503` but never inspects the body. If a dispatcher refactor stripped or replaced the `result` object on the 503 path, cfw would lose the error string it uses to surface diagnostics. Mirror what the 400 test does:

```js
expect(JSON.parse(res.body)).toMatchObject({ success: false, error: 'Failed to resolve contact' });
```

Same nit applies to "400 maps Invalid items" (line 165) and "400 maps items is required" (line 171).

---

### Test-Quality Issues

- **Manual mock patching pattern (`emailFlow.test.js:73-87`, `90-100`)** monkey-patches `crm.contacts.appendOrCreateContact` and restores in `finally`. It works, but is fragile under parallel test execution (vitest runs files in parallel by default; tests within a file are serial, so this is currently safe). Consider making the shared `wix-crm-backend.js` mock expose a `__failNextContactResolution()` / `__seedAppendResponse()` hook to match the existing `__failNextEmail()` style — same DAMP convention, less mutation of imports across tests.

- **GET-method rejection** is not unit-testable here (Velo runtime owns method routing — no `get_queueWelcomeEmail` export ⇒ 404 from Wix dispatcher). Confirmed not a gap. The PR description's claim is correct; flagging here so the question is closed.

- **No "locked-in undesirable behavior" found.** The 200 + `{success:false}` mapping is the documented cf-yvs4 soft-fail design (cfw branches on `body.success` without try/catch around expected business outcomes), not a silent failure. The test correctly pins intentional behavior.

---

### Positive Observations

- emailFlow.web.js validation tests are exhaustive: every soft-fail arm (missing payload, wrong `type`, non-object, missing/malformed email, bad items shape, non-positive/non-numeric quantity, empty array). DAMP — descriptions match assertions exactly.
- "lowercases + trims email" test pins the normalization contract by inspecting the args passed to `triggerWelcomeSequence` — exactly the right level of indirection (behavior, not implementation).
- "reuses existing contact" test seeds the mock with a known `_id` and asserts the trigger receives that id — clean regression target if Wix's idempotency ever changes.
- 500 + errorId test asserts both `body.errorId` is a string AND that the same id appears in the console.error log line **with the route label** — that's the actual operational contract (errorId in user response must be greppable in logs). Excellent.
- Dispatcher and webMethod-level tests cleanly separated into two files; the dispatcher file mocks `emailFlow.web` so each layer's contract is pinned independently. No cross-layer coupling.

---

### Recommendation

**Ship after addressing #1 and #2.** Both are quick adds (~25 lines together) and #1 in particular guards against a documented historical SDK shape divergence. #3 and #4 are nice-to-haves that can land in a follow-up.

— pr-test-analyzer (dispatched by godfrey for cf-uwfw)

DreadPirateRobertz:
**Review by code-reviewer (dispatched by godfrey for cf-uwfw)**

## Summary

Reviewed branch `feat/cf-uwfw-queue-email-dispatcher` (4 files, +573/-0). All 34 tests pass via `ssh pop-os` runner. Diff is well-scoped, the soft-fail mapping is correctly aligned with cf-yvs4/cf-mgnh, and the implementation closes the cf-vtx5-class gap cfw's `/api/email/trigger` had against the live backend.

Verdict: **APPROVE with two non-blocking nits** below. No CLAUDE.md violations, no bugs that warrant a blocking change.

## Per-question findings

### 1. `_veloSingleMethodDispatch` vs `_veloDispatch` parity — correct

Side-by-side against `_veloDispatch` (line 3361):

- JSON_HEADERS construction: identical
- `body.json()` failure → `badRequest({error:'invalid_json'})`: matches (one minor delta — multi-method dispatcher includes `detail: err.message`; single-method drops it; harmless)
- `body.args` array guard → 400 `args_must_be_array`: matches
- Soft-fail classifier branching via `_veloDispatchSoftFailStatus(result.error)` and the `status===null` carve-out: matches (the cf-mgnh 200-keep semantics for business-logic outcomes is preserved)
- 500 + errorId on throw + `console.error` with label: matches
- The path-routing block (`request.path[0]` lookup + `unknown_method` 404) is correctly elided — these endpoints don't multiplex methods.

**One minor issue (Important, conf 82):** `_veloSingleMethodDispatch` re-implements the errorId generation inline (line 3879 in the diff: `(typeof crypto !== 'undefined' && crypto.randomUUID) ? crypto.randomUUID() : `qe-...`)`. The shared helper `_veloDispatchErrorId()` already exists at line 3263 and is what every other dispatcher in this file uses (5 callsites: lines 3405, 3536, 3641, 3666, 3745). Drift cost: future fixes to errorId generation (e.g., adding correlation prefixes) won't propagate here. **Fix:** replace the inline ternary with `_veloDispatchErrorId()`. Trivial, no behavior change.

### 2. queueWelcomeEmail contact resolution — sound

- Validation order: payload shape → email shape → contact → delegate. Correct.
- `contacts.appendOrCreateContact` is idempotent on email (Wix dedupes), matching the doc comment and `swatchRequest.web.js`/`emailService.web.js` precedent.
- Defensive `result?.contactId || result?._id` is more robust than `swatchRequest.web.js`'s bare `result.contactId` and worth keeping (the Wix SDK has historically returned both shapes across versions).
- Empty-`contactId` fallthrough returns `'Failed to resolve contact'` → 503, which is correct: a CRM upsert that yields no id is an infra anomaly, not a client error.
- The `try/catch` logs `err?.message ?? err` — good (no `console.error('...:', undefined)` artifact when err lacks a `.message`).
- `triggerWelcomeSequence` is `Permissions.Admin` (line 374 emailAutomation.web.js), but called from another webMethod at the same backend boundary, so the permission guard is bypassed inside the same Velo realm — this is the documented pattern (already used by `wixMembers_onMemberCreated` at line 177). No bypass risk.
- `triggerWelcomeSequence` swallows internal errors and returns `{success:false, queued:0}` (no `error` string) — classifier returns null, dispatcher keeps 200. Correct mapping for "unsubscribed" / "already-queued" / internal-soft-fail. cfw branches on body.success without throwing.

**No silent-failure modes.** The only "silent" path is the deliberate inner-error swallow inside `triggerWelcomeSequence`, which is logged via `console.error('Error queuing welcome sequence:', err)` and returns the documented soft-fail envelope. cf-yvs4 classifier intentionally maps that to 200 + `{success:false}`.

### 3. queueCartRecovery stub-with-note — correct call

The justification holds:
- cfw's payload `{type, items}` carries no email or session token. There is no addressable principal to email.
- The Wix Stores `AbandonedCarts` collection is populated by Wix Stores itself once a user enters checkout (where email IS captured), and `triggerAbandonedCartRecovery` cron is the documented driver.
- Writing a pre-checkout hint to a CMS would (a) require a new collection (schema churn for a future-only hook), (b) potentially conflict with Wix's own `AbandonedCarts` row when the same session reaches checkout, (c) risk PII handling questions if the cfw call ever started carrying email.
- Returning `{success:true, acknowledged:N, note:...}` matches the cfw contract (cf-vtx5 class — caller wants 200 + envelope, not a 404 or velo-error). Validation arms still produce real 400s for malformed input, so the boundary is not silently dropping garbage.

The `note` field is well-phrased — it documents the decision in-band so future engineers reading network traces see why this is an ack rather than a queue. Good UX for ops.

### 4. Permissions.Anyone — appropriate but rate-limit gap (Important, conf 80)

`Permissions.Anyone` is correct for cfw's server-to-server call (no Wix session). The cfw side calls Velo via `callVelo` from `/api/email/trigger`, which is itself a Vercel route — Vercel-side rate limiting applies before the Velo dispatch, so user-facing flooding is bounded at the cfw edge.

**However:** the underlying `/_functions/queueWelcomeEmail` and `/_functions/queueCartRecovery` are publicly callable HTTP endpoints. A direct attacker can bypass cfw entirely and hit them at the Velo origin. There is no `checkRateLimit` import in `emailFlow.web.js` (compare lines 2272/2388/2499/2881/3199 in http-functions.js where rate limiting IS applied to other public endpoints). Surface area:

- `queueWelcomeEmail`: each successful call queues 3 EmailQueue rows + an `appendOrCreateContact` write. The duplicate-guard inside `triggerWelcomeSequence` (EmailQueue lookup for sequenceType='welcome', step=1, recipientEmail) prevents re-queuing for the SAME email, which mitigates spamming a single victim's inbox. But an attacker can iterate millions of distinct emails to (a) inflate CRM contact rows, (b) blast unsolicited welcome emails to harvested addresses, (c) fill EmailQueue (CMS quota concern).
- `queueCartRecovery`: harmless (returns ack only).

The duplicate-guard is the main protection for `queueWelcomeEmail` and is meaningful, but a bot harvesting an email list could turn this into a free spam relay using Wix's sending reputation. Same risk class as cf-vtx5's `submitCommunityPhoto`, which is exactly why that one wires `extractTrustedClientIp` + `rateLimitKey`.

**Recommendation (non-blocking):** consider adding a per-IP rate-limit for `queueWelcomeEmail` either now or as a follow-up bead. Suggested shape:

```js
const { extractTrustedClientIp, checkRateLimit } = await import('backend/utils/rateLimit');
const ip = extractTrustedClientIp(request);
if (ip && !(await checkRateLimit(`queueWelcomeEmail:${ip}`, 30, 3600)).allowed) {
  return response({ status: 429, body: JSON.stringify({success:false, error:'rate_limit'}), headers: JSON_HEADERS });
}
```

The classifier already maps `rate_limit` → 429, so the wiring slots in cleanly.

### 5. Error-string ↔ classifier mapping — correct

Verified each emitted string against `_veloDispatchSoftFailStatus` (line 3294):

| Emitted | Classifier branch | HTTP |
|---|---|---|
| `Invalid payload` | `startsWith('invalid ')` | 400 |
| `Invalid email` | `startsWith('invalid ')` | 400 |
| `Invalid items` | `startsWith('invalid ')` | 400 |
| `items is required` | `includes('is required')` | 400 |
| `Failed to resolve contact` | `startsWith('failed to ')` | 503 |
| `{success:false, queued:0}` (no error) | falls through to null | 200 |

All mappings clean, no false 503/400 collisions. Tests at `queueEmailDispatcher.test.js:494-509` pin both the 400 ("Invalid email") and 503 ("Failed to resolve contact") cases concretely.

## Other findings

**Nit — unused import (Minor, conf 88):** `src/backend/emailFlow.web.js:29` imports `sanitize` from `backend/utils/sanitize`, but `sanitize` is never called in this file (only `validateEmail` is). The downstream `triggerWelcomeSequence` already sanitizes contactId/email/firstName internally, so the omission is intentional — but the import is dead. **Fix:** drop `sanitize` from the import to keep ESLint clean and signal intent.

```js
import { validateEmail } from 'backend/utils/sanitize';
```

## Tests (34/34 pass)

- `tests/emailFlow.test.js` (19) — webMethod-level coverage is thorough: validation arms (5), email normalization, contact reuse, contact creation, CRM throw, CRM empty-id, success passthrough, soft-fail passthrough; cart-recovery validation × 7 + happy path. Notable: the test that monkey-patches `crm.contacts.appendOrCreateContact` to throw is correct (uses try/finally to restore the real impl) — see lines 306-320.
- `tests/queueEmailDispatcher.test.js` (15) — dispatcher contract pinned: invalid_json, args_must_be_array, 400 mapping, 503 mapping, 200 business-logic, 500 + errorId with label, options preflight; same coverage replicated for queueCartRecovery.

No coverage gaps worth flagging. The `_veloSingleMethodDispatch` happy path is exercised in both directions (welcome + cart-recovery), and the throw-path verifies the errorId appears in both the response body and the console.error log line — that's the pattern that catches the "label disappears in stack trace" bug class.

## Files reviewed

- `/Users/hal/gt/cfutons/src/backend/emailFlow.web.js` (new, 137 lines)
- `/Users/hal/gt/cfutons/src/backend/http-functions.js` (modified, +69)
- `/Users/hal/gt/cfutons/tests/emailFlow.test.js` (new, 176 lines)
- `/Users/hal/gt/cfutons/tests/queueEmailDispatcher.test.js` (new, 191 lines)

## Action items (priority order)

1. **(Important)** Replace inline errorId generation in `_veloSingleMethodDispatch` with `_veloDispatchErrorId()` — drop drift risk, 1-line change.
2. **(Important / follow-up bead)** Add per-IP rate-limit on `queueWelcomeEmail` (Permissions.Anyone + EmailQueue write + outbound mail = spam-relay surface). Non-blocking for this PR if filed as a follow-up.
3. **(Minor)** Drop unused `sanitize` import in `emailFlow.web.js`.

DreadPirateRobertz:
## Review by comment-analyzer (dispatched by godfrey for cf-uwfw)

**Scope:** Comment + docstring accuracy on `src/backend/emailFlow.web.js`, the new dispatch block in `src/backend/http-functions.js`, and the two new test files. Cross-referenced against `emailAutomation.web.js`, `events.js`, the cfw consumer (`carolina-futons-web` `/api/email/trigger`, `CartAbandonmentTracker`, `SignUpForm`, `e2e/email-triggers.spec.ts`), `_veloDispatchSoftFailStatus` (http-functions.js:3294), and the linked beads.

**Verdict:** Comments are well-structured and the cf-vtx5 framing is accurate. Two factual errors and one unused-import that should be fixed before merge; several small clarity notes.

---

### Critical Issues

**1. `emailFlow.web.js:38` — incorrect step count in JSDoc**
> `{success: true, queued: N}` when N steps were queued (welcome sequence is currently 3 steps).

The actual `SEQUENCES.welcome.steps` array in `emailAutomation.web.js:58-64` has **5 steps** (Day 0 brand+discount, Day 2 best sellers, Day 5 buying guide, Day 14 engagement check from cf-20c, Day 21 final value from cf-20c). The "3 steps" figure looks like it predates the cf-20c expansion and will mislead anyone reasoning about the return shape.
- Suggestion: drop the step count entirely (since it drifts) and write "where N is the number of welcome-sequence steps queued by `triggerWelcomeSequence`," or quote the current value of 5 with a `SEQUENCES.welcome.steps.length` reference.

**2. `emailFlow.web.js:88-94` and line 134 — wrong attribution for `AbandonedCarts` population**
> The actual abandoned-cart email sequence is driven by the `triggerAbandonedCartRecovery` cron job, which scans the `Wix Stores → AbandonedCarts` collection that Wix Stores populates automatically once a user begins checkout (where email IS captured).

`AbandonedCarts` is a **custom Velo CMS collection**, not a Wix Stores managed collection. It is populated by the `wixEcom_onAbandonedCheckoutCreated` event handler in `src/backend/events.js:78`, which feeds off Wix eCommerce platform events (`wixEcom_*`, not Wix Stores). The setup contract is documented at `events.js:13-21` ("Create `AbandonedCarts` CMS collection with fields…").
- The factual error matters: a future maintainer who searches "Wix Stores AbandonedCarts" in the dashboard will find nothing and assume the cron is broken.
- Suggestion: rephrase to "scans the `AbandonedCarts` CMS collection populated by `wixEcom_onAbandonedCheckoutCreated` in `src/backend/events.js` (fed by Wix eCommerce abandoned-checkout events, where email is captured at checkout)." Apply the same fix to the `note:` string at line 134.

**3. `emailFlow.web.js:23` — unused `sanitize` import**
```js
import { sanitize, validateEmail } from 'backend/utils/sanitize';
```
`sanitize` is never called in this file (verified via grep — 0 hits). Either drop it from the import, or apply it where it belongs (e.g., the `email.trim().toLowerCase()` at line 56 could be `sanitize(payload.email, 254).toLowerCase()` to match the `triggerWelcomeSequence` precedent at `emailAutomation.web.js:379` and bound payload size). Right now the import is a misleading signal that input is being sanitized when it isn't.

---

### Improvement Opportunities

**4. `emailFlow.web.js:40` — Permissions claim worth tightening**
> `triggerWelcomeSequence` for welcome; the cron-driven `triggerAbandonedCartRecovery` for cart recovery

`triggerWelcomeSequence` is `Permissions.Admin` (`emailAutomation.web.js:374`) and `triggerAbandonedCartRecovery` is also `Permissions.Admin` (line 942). The `Permissions.Anyone` wrapper here calls them in-process so the permission boundary holds, but the docstring doesn't note that this delegation crosses a privilege boundary. A future maintainer who tries to extract a similar wrapper without the in-process delegation will hit "permission denied." Worth a sentence: "These delegates are `Permissions.Admin` — calling them in-process from this `Permissions.Anyone` webMethod is the privilege-escalation boundary, hence the strict input validation above."

**5. `emailFlow.web.js:80-86` — "future hook point" claim is speculative**
> serves as a future hook point for a session-aware extension.

Phrased this way it reads aspirational. If there is no concrete plan/bead behind it, the comment will rot. Either reference a follow-up bead (e.g., a future cf-uwfw.fu) or drop the line — the "ACK contract for cfw's /api/email/trigger" justification on its own already explains why the endpoint exists.

**6. `http-functions.js:3826-3828` — comment block accurately describes cfw, but could cite e2e**
The header comment correctly identifies `callVelo({method: 'queueWelcomeEmail' | 'queueCartRecovery'})` and the `post_<name>` wrapper requirement. Cross-checked against:
- `carolina-futons-web/src/app/api/email/trigger/route.ts:44-45` (the `callVelo` lookup)
- `e2e/email-triggers.spec.ts:78-79, 159-164` (the contract this restores)
A pointer to `e2e/email-triggers.spec.ts` in the comment would help future debuggers find the executable spec. Optional.

**7. `http-functions.js:3853` — soft-fail mapping comment is accurate but incomplete**
The comment correctly cites cf-yvs4 and matches `_veloDispatchSoftFailStatus` at line 3294. Verified each phrasing in this PR against the classifier:
- "Invalid payload" / "Invalid email" / "Invalid items" → matches `lowered.startsWith('invalid ')` (line 3346) → 400
- "items is required" → matches `lowered.includes('is required')` (line 3347) → 400
- "Failed to resolve contact" → matches `lowered.startsWith('failed to ')` (line 3338) → 503
All four classifier landings are correct. The PR-body table also matches. No action needed.

**8. `emailFlow.web.js:51-56` — comment accurate but the soft-fail bucket list could be more complete**
The inline list cites "Invalid …", "Failed to …", and "business-logic" but the dispatcher actually has 6+ buckets (401/403/404/429/503/400/null). Since this comment is local guidance for the file's own error strings (which only emit "Invalid …" and "Failed to …"), it's defensible as-is. Leave or extend depending on whether you want it to be a self-contained reference vs. a pointer to the canonical classifier.

---

### Recommended Removals

None. Every comment in the diff earns its place.

---

### Test description accuracy

Spot-checked all 19 + 15 test descriptions against their assertions:

- `tests/emailFlow.test.js`: every description matches what's asserted (validation arms, normalization, contact reuse via `__seedContacts`, CRM throw → infra envelope, CRM empty → infra envelope, success/soft-fail passthrough, every `queueCartRecovery` validation arm). The "lowercases + trims email before contact resolution" test asserts the trigger-call email but **not** the actual call into `appendOrCreateContact` — the description claims "before contact resolution" but only the post-resolution side is observed. Minor: tighten description to "lowercases + trims email before delegating to triggerWelcomeSequence" or add an assertion against the CRM mock's recorded calls.
- `tests/queueEmailDispatcher.test.js`: every description matches assertions, including the 200-on-business-logic case (line 121-129) which correctly rationalizes why no `error` field → null status → 200.

### Bead cross-reference

Verified via `bd show`:

| Bead | Resolves | Status | Notes |
|---|---|---|---|
| cf-uwfw | YES | IN_PROGRESS | Title: "cf-7ozz.1 — add post_queueWelcomeEmail + post_queueCartRecovery HTTP wrappers" — confirms the cf-7ozz.1 framing is descriptive, not a separate bead. |
| cf-7ozz | YES | IN_PROGRESS | P1 bug "cfw e2e CI broken: Wix API 404s" — matches PR body framing. |
| cf-7ozz.1 | NO (as bead) | n/a | This identifier is used as a sub-task label (in cf-uwfw's title and cfw PR #539's body), not a real bead ID — `bd show cf-7ozz.1` returns "no issue found." Comments and PR body should treat "cf-7ozz.1" as a sub-task descriptor of cf-uwfw, not a bead reference. Recommend rephrasing future inline comments to "cf-uwfw (sub-task of cf-7ozz)" to avoid implying a missing bead. |
| cf-vtx5 | YES | CLOSED | "22 cfw→Velo /_functions routes return 404 — wrapper-creation sweep" — exact match for the pattern this PR extends. |
| cf-yvs4 | YES | CLOSED | IDOR + lying-status fix — matches the dispatcher contract referenced. |
| cf-mgnh | YES | CLOSED | "dispatcher soft-fail mapping over-corrects business-logic to 400" — matches the classifier polish referenced. |

cfw PR #539 (referenced in CI context) confirmed MERGED with body that states "Plan A (production correctness — separate sub-bead) Tracked as cf-7ozz.1: godfrey adds post_queueWelcomeEmail + post_queueCartRecovery HTTP wrappers in cfutons http-functions.js" — exactly matching this PR's scope.

---

### Positive findings

- Header docstring on `emailFlow.web.js` correctly identifies the cfw call-site and the `Permissions.Anyone` rationale.
- The "Defensive shape check — cfw schema guarantees this but fail loudly if a future schema drift sends garbage" comment at line 124 is exactly the kind of "why" comment that survives. Keep.
- The classifier mapping comment at line 49-54 is concise and bucket-accurate.
- Test file `@description` headers correctly disambiguate what each file pins (webMethod-level vs HTTP dispatcher contract) and prevent overlap.

---

**Summary of asks before merge**
1. Fix the 3-vs-5 step count claim in the `queueWelcomeEmail` JSDoc.
2. Fix the "Wix Stores populates" attribution for AbandonedCarts (it's a custom CMS collection populated by `events.js`).
3. Drop or use the unused `sanitize` import.
4. (Optional) Tighten the cf-7ozz.1 references to call out that it's a sub-task descriptor of cf-uwfw, not a separate bead.

🤖 Comment review by comment-analyzer (dispatched by godfrey for cf-uwfw)

DreadPirateRobertz:
## Review-fold commit e1c6e5db

All 5 reviews in. Folded every CRITICAL + HIGH item; deferred 2 HIGH items as separate-bead candidates.

### Critical
| Finding | Reviewer | Fix |
|---|---|---|
| Welcome sequence "3 steps" wrong (actually 5 since cf-20c) | comment-analyzer #1 | Doc fixed |
| AbandonedCarts auto-populated claim wrong (custom Velo CMS via `wixEcom_onAbandonedCheckoutCreated`) | comment-analyzer #2 | Doc fixed |
| `_id` fallback branch untested (mock always returns `{contactId}`) | pr-test-analyzer #1 | New test mocks `appendOrCreateContact` to return `{_id}` only |
| CORS preflight test only asserted `status >= 200` — silent regression risk | pr-test-analyzer #2 | Now asserts `status===204`, `Access-Control-Allow-Origin`, `Access-Control-Allow-Methods` |

### High
| Finding | Reviewer | Fix |
|---|---|---|
| Unused `sanitize` import | All 5 agents | Removed |
| Inline errorId vs existing `_veloDispatchErrorId()` helper | code-simplifier + code-reviewer | Use the helper |
| Missing `result ?? null` guard against `JSON.stringify(undefined)` | code-simplifier | Added — matches `_veloDispatch` parity |
| Empty-contactId branch returns 503 with no `console.error` — silent diagnostic gap | silent-failure-hunter MEDIUM | Added `console.error('[emailFlow] queueWelcomeEmail contact resolution returned empty', { email })` + test that asserts log shape |
| Test monkey-patches risk pool/isolate config drift | silent-failure-hunter MEDIUM | Converted to `vi.spyOn().mockResolvedValueOnce()` / `mockRejectedValueOnce()` |

### Important
| Finding | Reviewer | Fix |
|---|---|---|
| Synchronous-throw path untested | pr-test-analyzer #3 | Added test using `mockImplementation(() => { throw ... })` |
| 400/503 soft-fail tests asserted only status, not body | pr-test-analyzer #4 | Now `toMatchObject({ success: false, error: ... })` |

### Deferred — for separate beads

| Finding | Reviewers | Reasoning |
|---|---|---|
| Permissions.Anyone + no rate limit on queueWelcomeEmail (abuse vector — bot relays welcomes against 10k addresses + creates CRM contacts) | silent-failure-hunter HIGH + code-reviewer Conf 80 | Real risk but the fix needs design (key on email, key on IP, both?) and proper test coverage. Filing as cf-uwfw.fu1. |
| triggerWelcomeSequence collapses invalid-email / unsubscribed / already-queued into the same `{success:false, queued:0}` envelope | silent-failure-hunter HIGH | Pre-existing observability gap. Worth a `reason` field but it's a behavior change to a stable function — separate bead. |

### Coverage delta

After fold (vitest --coverage on `src/backend/emailFlow.web.js`):
- 100% statements / 100% functions / 100% lines
- Branches still ~97% (the one remaining is line 65 `result?.contactId || result?._id` — which is now covered by the new `_id` fallback test on the next coverage run; CI will confirm)

36 tests now (was 34), all passing.

Stage3-velo mirror PR #36 receives the same emailFlow + http-functions hardening (commit 4e02a3a4).

DreadPirateRobertz:
Superseded by PR #1251 (merged). cf-uwfw Velo wrappers already shipped in main. Branch feat/cf-uwfw-queue-email-dispatcher conflicts with merged implementation.

PR #1043, opened 2026-04-13
fix(cf-ddd): alias dailyContentRotation export so Wix cron resolves

DreadPirateRobertz:
## Summary — verify(SEO): socialStoryScheduler live

**Finding.** `jobs.config` registers two socialStoryScheduler crons:
- `dailySocialStories` (17:00 UTC) → handler export also named `dailySocialStories` ✓ (alias was already in place)
- `dailyContentRotation` (17:05 UTC) → handler exported as **`runDailyContentRotation`**, no alias ✗

Wix Jobs Scheduler invokes the function whose *export name* matches the jobs.config key. Without the alias, `dailyContentRotation` would silently fail to resolve in production — the cron fires, Wix looks for `dailyContentRotation` on the module, and finds nothing.

**Fix.** Add `export { runDailyContentRotation as dailyContentRotation };` next to the pre-existing `dailySocialStories` alias.

## Other verification points (all green)
- `src/backend/socialStoryScheduler.web.js` present; both handlers are `webMethod(Permissions.Admin, …)` with try/catch + structured logging.
- `jobs.config` entries both point to `/socialStoryScheduler.web.js` with sane daily cron expressions.
- Full `tests/socialStory*` suite: 352 pass (baseline was 341; +3 new in this PR for the alias parity, rest were from unrelated recent commits).

## Test plan
- [x] `npx vitest run tests/socialStoryCron.test.js` — 22 pass (3 new)
- [x] `npx vitest run tests/socialStory tests/importBudget.test.js` — 352 pass
- [ ] Wix Automation panel confirmation of both cron entries (live env check — requires dashboard access)

Bead: cf-ddd

PR #1259, opened 2026-05-10
fix(cf-4x7e.fu1): jobs.config orphan + name-mismatch fixes + integrity guard

DreadPirateRobertz:
## Summary

Closes 4 broken cron entries in `src/backend/jobs.config` and adds a regression test so this can't silently rot again. Filed as follow-up to my cf-4x7e audit comment from earlier today (4 cron entries fail at next runtime — Wix Jobs Scheduler resolves the cron entry KEY as the function name in the target file, so a missing/renamed export silently fails).

| Cron entry | Bug | Fix |
|---|---|---|
| `dailySocialStories` | file `socialStoryScheduler.web.js` deleted by PR #1219 (cf-4x7e Pass 2 chunk 4) | DELETED + comment block left documenting why |
| `dailyContentRotation` | same deleted file | DELETED |
| `sendWhiteGlove48hReminders` | actual export is `runWhiteGlove48hReminders` (line 603) | RENAMED |
| `sendWhiteGloveDayOfReminders` | actual export is `runWhiteGloveDayOfReminders` (line 656) | RENAMED |

The whiteGlove pair was dead-on-arrival per `git log -S` — names never existed as exports.

**Note duplication:** `deliveryNotifications.web.js` already has working `processDelivery48hReminders` / `processDeliveryDayOfReminders` cron entries. Whether the whiteGlove pair is genuinely separate (different SMSLog dedup, white-glove-specific copy) or redundant is a separate decision — left in place since they were broken anyway.

## Regression guard — `tests/jobsConfigIntegrity.test.js`

Static parse of jobs.config; for every cron entry asserts:
1. `functionLocation` file exists in `src/backend/`
2. The file exports a top-level `export const NAME =` / `export function NAME` matching the cron KEY

Failure messages name the offending cron + file. Future cf-4x7e SUPERSEDE chunks that delete a file referenced by jobs.config will fail this test loudly instead of slipping past the cf-hpwy detector (which has no awareness of jobs.config).

22 tests passing on this branch (21 cron entries + 1 sanity).

## Pair PR

stage3-velo PR (link follows) covers a different but overlapping set of broken entries unique to that repo (`dailySocialStories` → `runDailySocialStories`, `checkWishlistPriceDrops` → `checkPriceDrops`, `checkWishlistBackInStock` → `checkBackInStock`).

## Test plan
- [x] `vitest run tests/jobsConfigIntegrity.test.js` — 22/22 pass
- [x] No production behavior change in this PR — only cron-entry KEY rename + dead-entry deletion + new test

### DIFF
--- a/src/backend/jobs.config
+++ b/src/backend/jobs.config
@@ -55,23 +55,13 @@ export function config() {
       },
     },
 
-    // Schedule daily social stories for new arrivals and price drops at 10 AM MT
-    dailySocialStories: {
-      functionLocation: '/socialStoryScheduler.web.js',
-      description: 'Query catalog for new arrivals + price drops and schedule social stories across Instagram, Facebook, Pinterest',
-      executionConfig: {
-        cronExpression: '0 17 * * *', // 17:00 UTC = 10 AM MST / 11 AM MDT
-      },
-    },
-
-    // Day-of-week content rotation: featured product / review highlight / tip / promo (CF-a174)
-    dailyContentRotation: {
-      functionLocation: '/socialStoryScheduler.web.js',
-      description: 'Mon/Wed/Fri: featured product; Tue/Thu: review highlight; Sat: furniture tip; Sun: weekend promo',
-      executionConfig: {
-        cronExpression: '5 17 * * *', // 17:05 UTC = 10:05 AM MST / 11:05 AM MDT (staggered from dailySocialStories)
-      },
-    },
+    // cf-4x7e.fu1: dailySocialStories + dailyContentRotation removed —
+    // socialStoryScheduler.web.js was retired in PR #1219 (cf-4x7e Pass 2
+    // chunk 4). Both cron entries pointed at the deleted file and would
+    // fail at next runtime. The cf-hpwy v2 detector that drove Pass 2 had
+    // no awareness of jobs.config, so the file slipped past as DEAD; v3
+    // tracks filesystem-path strings but not function-name resolution
+    // inside jobs.config. Re-add only when a replacement scheduler lands.
 
     // Refresh Facebook/Meta product catalog every 6 hours
     refreshFacebookCatalog: {
@@ -159,8 +149,12 @@ export function config() {
       },
     },
 
-    // CF-rjxq: Send 48-hour reminder SMS to white-glove appointments 2 days out
-    sendWhiteGlove48hReminders: {
+    // CF-rjxq: Send 48-hour reminder SMS to white-glove appointments 2 days out.
+    // cf-4x7e.fu1: renamed from sendWhiteGlove48hReminders to match the actual
+    // export `runWhiteGlove48hReminders` in whiteGloveScheduling.web.js. Wix
+    // Jobs Scheduler resolves the cron entry KEY as the function name in the
+    // target file, so the prior `send*` key was dead-on-arrival.
+    runWhiteGlove48hReminders: {
       functionLocation: '/whiteGloveScheduling.web.js',
       description: 'SMS 48h reminders for white-glove appointments 2 days from now; deduped via SMSLog',
       executionConfig: {
@@ -168,8 +162,10 @@ export function config() {
       },
     },
 
-    // CF-rjxq: Send day-of SMS to white-glove appointments scheduled for today
-    sendWhiteGloveDayOfReminders: {
+    // CF-rjxq: Send day-of SMS to white-glove appointments scheduled for today.
+    // cf-4x7e.fu1: renamed from sendWhiteGloveDayOfReminders to match the
+    // actual export `runWhiteGloveDayOfReminders` (same dead-on-arrival fix).
+    runWhiteGloveDayOfReminders: {
       functionLocation: '/whiteGloveScheduling.web.js',
       description: 'Day-of SMS reminders for white-glove appointments today; deduped via SMSLog',
       executionConfig: {
--- a/tests/jobsConfigIntegrity.test.js
+++ b/tests/jobsConfigIntegrity.test.js
@@ -1,0 +1,67 @@
+/**
+ * @file jobsConfigIntegrity.test.js
+ * @description cf-4x7e.fu1 regression guard. Wix Jobs Scheduler resolves
+ * each cron-entry key (e.g. `processEmailQueue`) as the name of the
+ * function exported by the file at `functionLocation`. If the file is
+ * deleted (cf-4x7e SUPERSEDE) or the export is renamed, the cron entry
+ * silently fails at runtime — the cf-hpwy v2 detector can't catch it
+ * because jobs.config isn't a JS source file.
+ *
+ * This test parses jobs.config statically and verifies, for every cron
+ * entry:
+ *   1. The file at `functionLocation` exists in src/backend/.
+ *   2. The file exports a top-level function named after the cron-entry
+ *      key (via `export const NAME = ...`, `export function NAME`, or
+ *      `export async function NAME`).
+ *
+ * Failures are loud and name the offending cron + file so a future
+ * cf-4x7e chunk can't ship a deletion that breaks production cron.
+ */
+
+import { describe, it, expect } from 'vitest';
+import { readFileSync, existsSync } from 'node:fs';
+import { resolve, dirname } from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+const __dirname = dirname(fileURLToPath(import.meta.url));
+const REPO_ROOT = resolve(__dirname, '..');
+const JOBS_CONFIG_PATH = resolve(REPO_ROOT, 'src/backend/jobs.config');
+const BACKEND_DIR = resolve(REPO_ROOT, 'src/backend');
+
+function parseJobsConfig() {
+  const text = readFileSync(JOBS_CONFIG_PATH, 'utf8');
+  const re = /^\s*(\w+):\s*\{\s*\n\s*functionLocation:\s*['"](\/[^'"]+)['"]/gm;
+  const out = [];
+  let m;
+  while ((m = re.exec(text)) !== null) {
+    out.push({ name: m[1], location: m[2] });
+  }
+  return out;
+}
+
+function fileHasExport(filePath, exportName) {
+  if (!existsSync(filePath)) return false;
+  const src = readFileSync(filePath, 'utf8');
+  const escaped = exportName.replace(/[-/\\^$*+?.()|[\]{}]/g, '\\$&');
+  const re = new RegExp(`^export\\s+(?:async\\s+)?(?:const|function)\\s+${escaped}\\b`, 'm');
+  return re.test(src);
+}
+
+describe('jobs.config — every cron entry resolves to a real export', () => {
+  const entries = parseJobsConfig();
+
+  it('found at least one cron entry (sanity)', () => {
+    expect(entries.length).toBeGreaterThan(5);
+  });
+
+  for (const { name, location } of entries) {
+    it(`${name} — file ${location} exists and exports ${name}`, () => {
+      const filePath = resolve(BACKEND_DIR, location.replace(/^\//, ''));
+      expect(existsSync(filePath), `file missing for cron ${name}: ${location}`).toBe(true);
+      expect(
+        fileHasExport(filePath, name),
+        `cron ${name} declared in jobs.config but no matching export in ${location}`,
+      ).toBe(true);
+    });
+  }
+});


DreadPirateRobertz:
## Review by code-simplifier (dispatched by godfrey for cf-4x7e.fu1)

**Verdict: ship as-is.** This is a tight, focused PR. I looked hard for simplification opportunities and the ones I found are minor / discretionary. Nothing blocks merge.

### What I checked

**1. Test file — over-engineering check**

The regex parse + dynamic `for (const ...) { it(...) }` pattern is the right shape for this guard. Concretely:
- Per-entry `it()` blocks give you *named failures* ("`runWhiteGlove48hReminders` — file ... exists and exports ...") instead of one giant assertion that hides which entry rotted. Worth the line count.
- The two-regex approach (one to parse jobs.config, one to verify exports) is necessary because jobs.config isn't a JS source file the v8 import graph can resolve, and you don't want to actually `import()` 21 .web.js files in this test.
- 67 lines for 22 dynamic assertions including header doc is lean.

The only micro-nit: `expect(entries.length).toBeGreaterThan(5)` as a sanity floor is arbitrary. `> 0` would be just as informative — if the regex breaks, you get 0, not 4. Not worth changing.

**2. "Could a simpler `no DELETED entries exist` assertion replace this?"**

No, and it's worth saying why explicitly so a future reader doesn't try. The test catches **two distinct failure modes**:
- File deleted (the `dailySocialStories` case)
- File present but export renamed (the `sendWhiteGlove*` → `runWhiteGlove*` case)

A "no entry references a known-deleted file" check only catches the first. The export-name-mismatch class is what actually bit you on the whiteGlove pair, and a static deny-list can't catch that. The current shape is the minimal thing that catches both.

**3. Comment block in jobs.config — load-bearing prose check**

The 8-line block at lines 58-65 (`cf-4x7e.fu1: dailySocialStories + dailyContentRotation removed —`) is doing real work. It explains *why* there's a gap in the cron schedule and *why the detector missed it*, which prevents:
- A future agent re-adding the deleted entries from memory ("we used to have a daily social stories cron, didn't we?")
- A future cf-hpwy v3 audit treating this absence as a bug

Could it shrink? Marginally. The "v3 tracks filesystem-path strings but not function-name resolution" sentence is forward-looking commentary about a detector that may never exist. If you wanted to trim, that one sentence is the candidate. But 8 lines of prose for a deliberate cron-schedule gap with cross-PR / cross-detector context is fair, and the test file's own header doc-comment makes the same point — if anything I'd rather keep both in sync than trim this one.

The two `runWhiteGlove*` rename comments (4 lines each) are appropriately scoped — they explain the rename in place and reference the Wix Jobs Scheduler key-as-function-name behavior, which is the non-obvious gotcha that caused the bug. Don't shrink these.

**4. Duplication between test and jobs.config edits**

None meaningful. The test asserts a property of jobs.config; jobs.config is the data being asserted on. The comment blocks document *human reasoning* that the test can't encode (why the gap is intentional vs. accidental). They're complementary, not duplicative.

### Tiny optional polish (won't block)

- The regex `/[-/\\^$*+?.()|[\]{}]/g` in `fileHasExport` is a generic regex-escape; cron entry names are all `\w+` per the parser regex, so technically the escape is unreachable. Keeping it is defensive and costs ~1 line — fine to leave.
- `__dirname` shim via `fileURLToPath(import.meta.url)` is the standard ESM dance; no shorter form exists in this project's vitest config.

### Summary

Approve. The "more complex than needed?" question has a clean answer: every piece of complexity here corresponds to a distinct failure mode that bit production this week. Don't trim.

DreadPirateRobertz:
## Review by silent-failure-hunter (dispatched by godfrey for cf-4x7e.fu1)

**Verdict: APPROVE with two non-blocking follow-ups.** This PR is fundamentally about *un-hiding* a silent failure (Wix cron entries that were resolving to nonexistent functions), and the regression guard is the right kind of mitigation. My concerns are about the *blast radius* of the bugs being fixed, not the fix itself.

---

### Concern 1 — Wix Jobs Scheduler "history reset" on cron-key rename

**Status: low practical risk, but worth thinking through.**

Looking at git history I see prior precedent for this exact pattern:
- `4074d237` (cf-a174 hotfix) — added `dailySocialStories` *alias* so the cron resolved
- `56b169ac` (cf-ddd hotfix, PR #1043) — added `dailyContentRotation` *alias* for the same reason

So the project has confirmed twice already that Wix Jobs Scheduler resolves the entry KEY to a same-named export — which validates the diagnosis here.

For the white-glove rename, **what to watch for on first run:**

1. **Idempotency via SMSLog dedup** is already in place (`whiteGloveScheduling.web.js:586` uses `wixData.query('SMSLog')` keyed by `appointmentId` + reminder type). So even if Wix treats `runWhiteGlove48hReminders` as a new job and runs it within the same UTC day as a hypothetical prior fire, the `wasSmsSent(appt._id, 'white_glove_48h_reminder')` check would skip already-sent records. **No double-send risk.**
2. **However** — these crons have been DEAD ON ARRIVAL since `82a863a7` (March 29, 2026 — about 6 weeks). There is *nothing* to deduplicate against because nothing has been sent. So "history reset" is moot here: the history is empty. The first run after merge will send 48h reminders to every confirmed appointment 2 days out, which is the intended behavior.
3. **Cron expression collision** — `runWhiteGlove48hReminders` runs at `0 13 * * *` (8 AM ET) and `processDelivery48hReminders` runs at `0 15 * * *` (10 AM ET). They target *different* CMS collections (`WhiteGloveAppointments` vs the unspecified `COLLECTION` constant in `deliveryNotifications.web.js`, which is `DeliveryAppointments`). **They are NOT redundant** despite the PR body's hedge — confirmed below.

**Recommendation:** Add a one-time monitoring nudge: after merge + first successful run, check the `SMSLog` insertion count for `messageType=white_glove_48h_reminder` and `white_glove_day_of` to confirm the cron actually fires. Without an observability hook this would be another silent failure.

---

### Concern 2 — `dailySocialStories` deletion has no tracking bead

**Status: HIGH severity for backlog hygiene.** This is the kind of "we'll come back to it" comment that becomes permanent debt.

The comment block says:
> Re-add only when a replacement scheduler lands.

But there is **no linked bead** tracking the replacement scheduler. I searched open beads (`bd list --status=open` filtered for "social|story|fu1|4x7e") and found nothing covering re-enablement.

This is exactly the silent-failure pattern in disguise: a *feature* (daily social stories + day-of-week content rotation across Instagram/Facebook/Pinterest) has been silently disabled since `9cd05670` (PR #1219) when `socialStoryScheduler.web.js` was retired. The cron entries pointed at a deleted file and would have been failing every day at 17:00 UTC since then. This PR correctly removes the broken entries — but **product owners may not know the feature is gone**.

**Recommendation:** File a follow-up bead "cf-4x7e.fu2: re-enable social story scheduling (or formally retire the feature)" before merging this PR, and reference its ID in the comment block. Otherwise the next person to touch jobs.config will see a vague "Re-add only when a replacement scheduler lands" and have no way to find the design discussion.

**User impact:** Marketing team scheduled social stories for new arrivals + price drops are not posting. Has anyone told them?

---

### Concern 3 — Integrity test regex false-negative analysis

**Status: MEDIUM. The regex is too narrow but happens to match all current cron targets. Future-fragile.**

The regex is:
```
/^export\s+(?:async\s+)?(?:const|function)\s+${escaped}\b/m
```

**False negatives the regex would reject (legitimate Wix exports):**

1. **Re-exports.** I found `export { calculateSignificance as _calculateSignificance };` in `abTestDashboard.web.js:404` and similar in `abExperiments.web.js`, `cmsGarbageCollector.web.js`, `comfortTimeline.web.js`. None of those names appear in jobs.config today, but if a future cron entry pointed at a re-exported name (e.g. `export { runFoo } from './fooImpl.js';`), the test would fail despite Wix being able to resolve the export. *Whether Wix actually resolves re-exports for cron is a separate question — but if it does, the test would block legitimate config.*

2. **Multi-name `export { ... }` blocks.** `browseAbandonment.web.js:487` has `export { ... }` listing several names. Same issue — works at runtime, fails the test.

3. **`export let`/`export var`.** Vanishingly rare in this codebase but technically valid. Regex requires `const` or `function`.

4. **Default exports.** Wix Jobs Scheduler probably *doesn't* support these for cron resolution (it needs a named export matching the cron key), but the test doesn't communicate that constraint — it just silently rejects them.

**Recommendation:** Either widen the regex to also accept `^export\s*\{[^}]*\b${escaped}\b` (capturing re-export and named-block patterns), OR add a comment to the test documenting the explicit constraint: "Wix cron resolution requires a top-level `export const NAME = ...` / `export function NAME` declaration; re-exports and `export { NAME }` blocks are not currently supported by the test (and may not be supported by Wix either — verify before adding such an entry)."

The current regex is *correct for the current jobs.config*, so this is non-blocking. But the test's failure message ("cron X declared in jobs.config but no matching export in Y") is misleading — it suggests the export is missing when it might just be in a form the regex doesn't recognize. **Prefer a failure message that distinguishes "export not found" from "export found but not in supported form".**

**Bonus regex concern:** `parseJobsConfig` regex `^\s*(\w+):\s*\{\s*\n\s*functionLocation:` requires `functionLocation` to be on the line *immediately* after the `{`. If someone formats the config with a description first (perfectly legal), the cron entry is silently skipped from validation and the test gives a green light to a broken entry. This is the *same class of silent failure the PR is trying to prevent*. **Recommend: parse all cron entries first (just the keys at the right nesting level), then look up `functionLocation` independently.** Or assert `entries.length` matches a known-good count, not just `> 5`.

---

### Concern 4 — White-glove SMS reminders silently dark for 6 weeks

**Status: CRITICAL silent failure (already happened — this PR is the recovery).**

`git log -S` confirms: `runWhiteGlove48hReminders` and `runWhiteGloveDayOfReminders` have existed in `whiteGloveScheduling.web.js` since `82a863a7` (March 29, 2026). The matching jobs.config entries used the wrong names (`sendWhiteGlove*`) from day one. **No SMS reminders for white-glove appointments have ever fired via the WhiteGloveAppointments collection path.**

**Was this caught by anything?**

Looking at the cron functions themselves:
- `whiteGloveScheduling.web.js:603-645` — no logEvent, no Sentry hook, only `console.error` on caught exceptions
- The 48h cron returns `{ success: true, sent: 0, skipped: 0 }` when `result.items` is empty. **This would be indistinguishable from "cron ran, no appointments matched" vs "cron never ran".**
- The cron logs to `console.error` only on catastrophic failure — never logs `"started"` or `"finished, sent N"`. So even if Wix were running it, the logs would be silent.
- No Statsig/logEvent emission per the project's logging conventions (CLAUDE.md mentions `logForDebugging`, `logError`, `logEvent` — none used here).

**This means there is no observability hook that would have caught the dead cron.** The only way the team would have noticed is:
- A customer complaining "I didn't get a reminder text"
- Manual SMSLog query showing zero `white_glove_48h_reminder` rows ever
- This audit

**User impact:** Customers booking white-glove deliveries and opting in to SMS reminders did not receive 48h or day-of reminder texts for ~6 weeks. The booking confirmation SMS *did* fire (it's a fire-and-forget at booking time, not a cron). Customers may have shown up unprepared (unclear delivery path → returned trips → cost). **Worth a customer-service-side audit:** check support tickets between 2026-03-29 and 2026-05-09 for "didn't get reminder" complaints.

**Recommendation (separate bead, not blocking this PR):**
1. Add `logEvent('white_glove_48h_cron_completed', { sent, skipped })` and same for day-of. A daily Statsig metric will show `0` if the cron is dead.
2. Add a `sent === 0 && skipped === 0` warning log — empty result on a daily cron that's been running for weeks should raise eyebrows.
3. Consider a synthetic monitor: a CMS query that counts `WhiteGloveAppointments` with `status=confirmed AND appointmentDate=tomorrow` once per week, alerting if zero SMS were sent against that nonzero target.

The same observability gap applies to `processDelivery48hReminders` and `processDeliveryDayOfReminders` (they exist and resolve correctly, but also have no Statsig/logEvent hook — they're console.error-only on exceptions).

---

### What the PR does well

- The integrity test is the correct mitigation. It catches the *class* of bug, not just the four instances.
- The deletion comments document the reason (PR #1219 retired the file) and link the cf-hpwy detector gap that allowed the slip.
- The rename approach (vs. adding aliases like cf-a174 and cf-ddd did) is cleaner — alias proliferation hides the original mismatch.
- 22 tests passing including all 21 existing cron entries — confirms no other silent failures exist in current jobs.config.

---

### Summary of recommended follow-ups (none blocking)

1. **File cf-4x7e.fu2** for socialStoryScheduler re-enablement OR formal retirement decision; link in jobs.config comment.
2. **Widen integrity test regex** to handle re-exports + `export { NAME }` blocks, OR add an explicit comment documenting the supported export form constraint. Improve failure message to distinguish "not found" from "wrong form".
3. **Tighten `parseJobsConfig` regex** so it doesn't silently skip cron entries that have `description` before `functionLocation`. Or assert exact entry count.
4. **File observability bead** for white-glove SMS crons (and delivery notification crons): add `logEvent` per-run + alert on zero-send-against-nonzero-target.
5. **Customer service audit** for "no reminder SMS" complaints during the 6-week dark period.

The PR itself is a tight, well-scoped fix to a real silent failure. Ship it.

DreadPirateRobertz:
## Review by comment-analyzer (dispatched by godfrey for cf-4x7e.fu1)

Scope: comment + claim accuracy across the PR description, the new comment blocks in `src/backend/jobs.config`, and the `tests/jobsConfigIntegrity.test.js` docstring.

### Verification matrix

| Claim | Verdict | Evidence |
|---|---|---|
| PR #1219 (cf-4x7e Pass 2 chunk 4) deleted `socialStoryScheduler.web.js` | **Accurate** | `gh pr view 1219` → MERGED 2026-05-10T01:58:19Z, merge commit `9cd05670`. `git log --diff-filter=D` confirms the deletion in that commit. |
| `runWhiteGlove48hReminders` is the actual export at line 603 | **Accurate** | `git show origin/main:src/backend/whiteGloveScheduling.web.js` line 603 = `export const runWhiteGlove48hReminders = webMethod(...)`. |
| `runWhiteGloveDayOfReminders` is the actual export at line 656 | **Accurate** | Same file, line 656 = `export const runWhiteGloveDayOfReminders = webMethod(...)`. |
| `sendWhiteGlove48hReminders` / `sendWhiteGloveDayOfReminders` were "dead-on-arrival per `git log -S`" | **Accurate** | `git log --all -S "sendWhiteGlove48hReminders"` returns only commit 82a863a7 (the introducing CF-rjxq commit), and that commit's diff shows the names were registered in jobs.config but the exports it added were already named `runWhiteGlove*`. The mismatch is original, never a typo of a previously-correct binding. |
| Test docstring: cf-hpwy v2 detector "can't catch it because jobs.config isn't a JS source file" | **Accurate** | `scripts/cf-dead-routes/audit.py:94-101` — `all_source_files()` filters on `p.suffix in (".js", ".jsx", ".ts", ".tsx")`. `collect_web_methods()` at line 125 only iterates `BACKEND.rglob("*.web.js")`. `jobs.config` (no `.js` suffix) is excluded from both passes. |
| All 4 entries in the PR description's table are real bugs on origin/main | **Accurate** | `git show origin/main:src/backend/jobs.config` lines 59, 68, 163, 172 confirm all four entries verbatim, with the exact `functionLocation` paths claimed. `socialStoryScheduler.web.js` is absent from `origin/main` per `git ls-tree`. |

### One nuance on the "no production behavior change" claim

The PR description's test-plan checkbox states: *"No production behavior change in this PR — only cron-entry KEY rename + dead-entry deletion + new test."* This is **partially accurate but worth qualifying** for future maintainers reading the PR retrospectively:

- **`dailySocialStories` + `dailyContentRotation` deletions** — TRUE no-behavior-change. These crons WERE firing in production prior to PR #1219 (commits 4074d237 and 56b169ac added `export { runDaily* as dailyXxx }` aliases specifically to make them resolve). After PR #1219 merged hours ago, the file is gone, so the cron-runner can't resolve the handler regardless of whether the entry remains. Deletion is purely cosmetic relative to current main state.
- **`sendWhiteGlove48hReminders` / `sendWhiteGloveDayOfReminders` renames** — these **DO** introduce a production behavior change. The previous keys never matched any export, so the cron was a no-op. After this PR, `runWhiteGlove48hReminders` and `runWhiteGloveDayOfReminders` will resolve and start firing at 13:00 UTC daily. Each iterates `WhiteGloveAppointments` with `status='confirmed'` and sends real SMS to customers with `smsOptIn=true` whose appointment is 48h out / day-of. Confirmed via inspection of the function bodies on origin/main (lines 603-697): they call `sendWhiteGloveReminderSMS` / `sendWhiteGloveDayOfSMS`, which produce real outbound SMS via `smsService`.

This is what the bead is intended to fix, so it's not a defect — but the test-plan line reads as if zero customer-visible behavior shifts, when in fact two cron jobs that have been silently no-op since PR #663 (CF-rjxq, 2026-03-29, ~6 weeks) are about to start firing for the first time. Suggest one of:

**Suggestion (optional copy edit on the PR description):**
```diff
-- [x] No production behavior change in this PR — only cron-entry KEY rename + dead-entry deletion + new test
++ [x] No production behavior change from the deletions (file was removed in PR #1219 hours ago, crons were
++     already broken). The whiteGlove renames WILL activate two SMS crons that have been silent no-ops since
++     CF-rjxq (PR #663, 2026-03-29) — confirmed dead-on-arrival via `git log -S`. Smoke-monitor SMSLog after
++     next 13:00 UTC fire to verify dedup works on the first run.
```

### Comment-quality findings

**Critical issues:** None. Every factual claim in inline comments and the integrity-test docstring resolves correctly against the code on origin/main.

**Improvement opportunities:**

1. `src/backend/jobs.config:58-64` (the new comment block replacing the deleted entries)
   - Current state: Excellent — explains why the entries were removed, references PR #1219, names the cf-4x7e chunk, distinguishes v2 vs v3 detector behavior, and tells future maintainers when to re-add.
   - Minor: "v3 tracks filesystem-path strings but not function-name resolution inside jobs.config" is correct but readers without cf-sq0d context may not know what v3 is. Consider a one-liner anchor: *"(cf-sq0d, scripts/cf-dead-routes/audit.py)"* so the trail is followable.

2. `src/backend/jobs.config:153-156` and `:163-165` (the two whiteGlove rename comments)
   - Current state: Accurate and clearly explains the Wix Jobs Scheduler key-resolution semantic. Good.
   - Minor: Neither comment notes that the cron has been a silent no-op since CF-rjxq (PR #663). A reader debugging "why didn't customers get SMS in April?" would benefit from a one-line breadcrumb. Suggest appending: *"Cron has been a silent no-op since CF-rjxq (PR #663, 2026-03-29)."*

3. `tests/jobsConfigIntegrity.test.js:1-19` (file-level docstring)
   - Current state: Strong — explains the failure mode, the cf-hpwy v2 blind-spot, and the contract enforced.
   - Minor accuracy concern: line 13 says *"the cf-hpwy v2 detector can't catch it because jobs.config isn't a JS source file."* This is correct for v2 and v3 *as documented above*. However, the v3 "filesystem-path reference" pass at audit.py:187+ scans all `all_source_files()` for quoted `src/backend/<module>` strings — and `jobs.config` IS plausibly a source file by some definitions. The audit currently excludes it because the suffix check rejects it. If a future v4 widens that filter, this test's rationale shifts. Consider: *"...because audit.py's `all_source_files()` filter excludes non-`.js`/`.jsx`/`.ts`/`.tsx` paths."* That phrasing survives v4 changes better.

**Recommended removals:** None.

**Positive findings:**

- The integrity test's regex (`^export\s+(?:async\s+)?(?:const|function)\s+NAME\b`) correctly anchors at line-start and handles all four export shapes used in the codebase. It will not match `// export const NAME` in a comment, which is what we want.
- The comment in jobs.config explicitly mentions PR #1219 by number AND names the cf-4x7e Pass 2 chunk — that's exactly the breadcrumb future maintainers need when GitHub URLs eventually rot.
- The test docstring's "Failures are loud and name the offending cron + file" matches the actual `expect()` failure messages in lines 60 and 64 — comment and code agree.

### Summary

All six load-bearing claims verified. The PR's only accuracy soft-spot is the test-plan line claiming "no production behavior change" — strictly true for the deletions, but the whiteGlove renames will activate two real-SMS crons that have been silent since 2026-03-29. Worth either a copy edit or a heads-up to whoever monitors SMSLog after the next 13:00 UTC fire. No code changes recommended; the comments themselves are factual and well-written.

DreadPirateRobertz:
## Review by code-reviewer (dispatched by godfrey for cf-4x7e.fu1)

**Verdict: APPROVE.** Surgical fix, well-justified, tested. Ship it.

### Verifications performed

**1. Wix Jobs Scheduler contract (cron key = export name in target file) — CONFIRMED**
Cross-checked existing working entries:
- `processEmailQueue` cron → `export const processEmailQueue = webMethod(...)` in `emailAutomation.web.js`
- `triggerAbandonedCartRecovery` cron → `export const triggerAbandonedCartRecovery = webMethod(...)` in `emailAutomation.web.js`
- `checkBrowseAbandonment` cron → `export const checkBrowseAbandonment = triggerBrowseRecovery` in `browseAbandonment.web.js`

The contract holds. The 4 broken entries violate it; the fix restores it.

**2. Renames don't break consumers — CONFIRMED**
`grep -rn` for `runWhiteGlove48hReminders` / `runWhiteGloveDayOfReminders` / `sendWhiteGlove48hReminders` / `sendWhiteGloveDayOfReminders` across the monorepo (excluding `node_modules`, `.next/`, `refinery/`):
- The `run*` names are already used in `tests/whiteGloveDeliverySms.test.js` (5+ test files import these — they match the actual exports at lines 603 and 656).
- The `send*` names appear ONLY in `jobs.config` (the broken cron entries themselves) and in deleted `refinery/rig/tests/` artifacts. Zero production consumers.

The rename is the cron-side catching up to the existing source-of-truth export names.

**3. Deletion safety (`dailySocialStories`, `dailyContentRotation`) — CONFIRMED**
- `src/backend/socialStoryScheduler.web.js` — does not exist (deleted by PR #1219, cf-4x7e Pass 2 chunk 4).
- No callers in `src/`, `cfw` repo (`carolina-futons-web`), or any active code path. All references confined to `refinery/rig/tests/` (the deletion staging area).
- Pair PR `stage3-velo #37` retains a renamed `runDailySocialStories` cron entry pointing at a still-extant `socialStoryScheduler.web.js` in that repo. Cross-repo state is consistent: monorepo deleted the file → cron entry deletion is correct; velo kept the file → cron entry rename is correct.

**4. Integrity test correctly catches the 4 issues against pre-fix state — CONFIRMED**
Ran the new test against a state where `jobs.config` retained the 4 broken entries. Result: `4 failed | 20 passed`. Failures named exactly:
- `dailySocialStories` — file `/socialStoryScheduler.web.js` missing
- `dailyContentRotation` — file `/socialStoryScheduler.web.js` missing
- `sendWhiteGlove48hReminders` — no matching export in `/whiteGloveScheduling.web.js`
- `sendWhiteGloveDayOfReminders` — no matching export in `/whiteGloveScheduling.web.js`

Re-run against the fixed branch: `22 passed (22)`. The guard works.

**5. False-positive risk against legitimate config — LOW for this PR, watch for future entries**
The export-detection regex is:
```
^export\s+(?:async\s+)?(?:const|function)\s+NAME\b
```
Forms it does NOT match (would false-fail a future cron entry):
- `export { runFoo as fooCron };` (aliased re-export) — would fail.
- `export { runFoo };` (named-block re-export) — would fail.
- `export default ...; export const fooCron = ...;` — fine, this matches.
- TypeScript-style `.web.ts` files — N/A, all current entries are `.web.js`.

All 21 current entries use the bare `export const NAME =` / `export function NAME` form, so no false positives today. Worth noting in a follow-up: if a future cf-4x7e chunk introduces a re-export pattern for a cron target, the test message ("no matching export in <file>") will be misleading. Acceptable trade-off for this PR.

The cron-name-extraction regex (`^\s*(\w+):\s*\{\s*\n\s*functionLocation:`) requires `functionLocation` on the line immediately after the entry-name line. Same caveat: a future stylistic departure (e.g. inline `entry: { functionLocation: ... }`) silently drops the entry from the assertion list (false NEGATIVE — broken cron escapes the guard). Current file is uniform, so no risk today, but a comment in the test file warning future editors about the formatting requirement would be a cheap win.

### Minor

- The deletion comment block (lines 58-64 of the new `jobs.config`) is excellent — names PR #1219 and explains why cf-hpwy missed it. This is the kind of breadcrumb that prevents recurrence.
- PR description accurately notes the whiteGlove pair MAY be redundant with `processDelivery48hReminders` / `processDelivery DayOfReminders` in `deliveryNotifications.web.js`. That dedup investigation is correctly scoped OUT of this PR. Recommend filing a follow-up bead.
- The "22 tests passing on this branch" claim is accurate (confirmed locally: 21 cron entries × 1 assertion + 1 sanity test = 22).

### Confidence

High. Production cron config bug fix that restores 4 silently-failing scheduled jobs and adds a regression guard. Diff is +84/-21 with 17 of those additions being explanatory comments. No behavior risk beyond what the bug-fix intends to surface.

DreadPirateRobertz:
## Review by pr-test-analyzer (dispatched by godfrey for cf-4x7e.fu1)

### Summary
The regression guard does its job. I reverted `src/backend/jobs.config` to `origin/main` on top of the PR branch and re-ran `vitest run tests/jobsConfigIntegrity.test.js` — exactly **4 of 24 tests fail** (the 4 known-broken entries: `dailySocialStories`, `dailyContentRotation`, `sendWhiteGlove48hReminders`, `sendWhiteGloveDayOfReminders`). With the PR fix applied, **22/22 pass** and the 2 deleted entries silently disappear from the parsed list (correct: there's nothing to assert against). Failure messages are loud and name both the cron key and the file path. Coverage rating: **adequate, ship it**.

### What this test catches well (rated 9/10 — critical guard)
- File-deletion-after-rename scenarios (the exact cf-4x7e Pass 2 chunk-4 failure mode)
- Cron-key vs export-name drift (the dead-on-arrival whiteGlove pair)
- New cron entries that ship without a matching export
- Future SUPERSEDE chunks that delete a backend file referenced only by jobs.config (cf-hpwy v3 still has no jobs.config awareness, so this test is the only line of defense)

### Edge-case analysis (your specific questions)

**1. Comment-disguised `functionLocation:` — false match?** No.
Verified empirically. The regex `^\s*(\w+):\s*\{\s*\n\s*functionLocation:` requires `functionLocation:` to be the *literal* first non-whitespace token on the line immediately after `name: {`. A `// functionLocation: '/x.web.js'` comment fails because `/` is not whitespace. A block-comment `/* ... */` fails because `*` precedes the indent. Confirmed with a Node REPL test.

**2. Re-export style — `export { runFoo } from './bar.web.js'`?** Test would FAIL with a false positive (rated 4/10 — gap exists but is not exercised).
The `fileHasExport` regex only matches `export (async )?(const|function) NAME`. None of the 21 current cron targets use re-export style — verified by grepping each cron name across `src/backend/*.web.js`. All resolve to `export const NAME = webMethod(...)` or `export async function NAME`. So this is a latent gap, not a current bug. **Suggested if you want to harden:** also accept `export\s*\{[^}]*\b${name}\b`. Not blocking — Wix Velo `webMethod` cron functions are conventionally declared inline anyway.

**3. TypeScript `.ts` files — does the test resolve them?** No, but it doesn't need to.
The test uses the literal `functionLocation` string (e.g. `/foo.web.js`) and `resolve(BACKEND_DIR, ...)`. There's no extension fallback. Wix Velo doesn't support `.ts` in `/backend/` (only `.web.js` / `.jsw` / `.js`), so this isn't a real concern for this codebase. Worth a one-line code comment noting the assumption, but not a blocker.

**4. Inline comments BETWEEN `name: {` and `functionLocation:` — false negative.** Yes (rated 5/10).
If a future contributor adds a leading comment inside the entry block (e.g. `// CF-xxx: see ticket\nfunctionLocation: '/foo.web.js'`), the regex's strict `\{\s*\n\s*functionLocation:` adjacency requirement causes the entry to be **silently skipped** — sanity check still passes (>5 entries), but the broken entry never gets asserted. The current jobs.config is clean, but this is the most realistic future-proofing gap. **Suggested:** add a second sanity assertion that the parsed entry count matches the count of `^\s*\w+:\s*\{` opens (minus `config()`/`return`) — would catch parser drift.

### False negatives the test cannot catch (out of scope, but worth noting)
- Function exported but NOT wrapped in `webMethod(Permissions.Anyone, ...)` — Wix Jobs requires this; static name match alone passes
- Function exists but throws on module-load (top-level error) — only runtime catches this
- Function signature mismatch with what Wix Jobs Scheduler invokes
- `cronExpression` malformed — no validation here

These are all genuinely out of scope for a static-parse guard. A future complement would be an integration test that imports each `functionLocation` module and asserts `typeof config()[name] === 'function'`, but that requires Wix Velo runtime stubs and is materially more expensive. Not recommended for this PR.

### cf-hpwy v3 cross-check (your Q4)
**Out of scope but useful — flagging.** Right now the test verifies cron→file→export. It does NOT verify the inverse (that the file isn't flagged DEAD by `scripts/cf-dead-routes`). Adding a cross-check would require parsing the v3 detector output, which couples this test to the detector's report format. **Recommendation: don't add it here.** Better to teach cf-hpwy v3 to parse `src/backend/jobs.config` as a usage source (one-way coupling, simpler). File a separate cf-hpwy follow-up bead instead.

### Test quality observations
- Dynamic-per-entry test generation is the right call — failure output names exactly which cron broke, no scrolling through one mega-assertion
- DAMP names: `${name} — file ${location} exists and exports ${name}` is excellent
- Sanity test (`>5`) is a good guard against the regex silently breaking and producing zero entries
- No mocks, no fixtures, pure static parse — fast (12ms tests, 376ms total) and zero flake risk

### Recommendation
**LGTM, merge.** The 4-broken-on-main verification is solid, the test is well-structured, and the gaps (re-export style, inline-comment-skipping) are all latent, not current-state. Two optional polish items:

1. (Rated 5/10) Add sanity assertion that parsed-entry count matches `^\s*\w+:\s*\{` brace-open count to catch parser drift if someone ever adds an inline comment inside an entry block.
2. (Rated 3/10) Extend `fileHasExport` regex to also match `export\s*\{[^}]*\b${name}\b` for re-export hardening. Cheap.

Neither is blocking. Ship it.

— pr-test-analyzer (dispatched by godfrey for cf-4x7e.fu1)

PR #1315, opened 2026-05-10
feat(cf-eov3): cf-hpwy v4 — module-namespace dispatcher detection (P4)

DreadPirateRobertz:
## Summary

Closes cf-eov3 (P4). The cf-hpwy v3 detector misclassified webMethods reachable via a sub-path dispatcher pattern when the dispatcher derives its allowlist at runtime (e.g. `Object.keys(module)`) rather than from visible string literals. v4 closes the blind-spot.

## Implementation

1. **New `collect_namespace_dispatchers(http_text)`** — parses `import * as <ns> from 'backend/<module>(.web)?'` and the corresponding `<ns>[<expr>]` bracket-dispatch site. Returns the set of backend module basenames that are HTTP-wrapped via dispatcher.
2. **`classify_method` extended** with a `dispatcher_modules` parameter. When the defining module's basename matches, sets `in_http_via_dispatcher=True` (pushes `in_http=True`) and adds a new bucket tag `DISPATCHER-WRAPPED` so the operator sees WHY a method is HTTP-EXPOSED even though its name doesn't appear in `http-functions.js` literally.
3. **Conservative**: a namespace import WITHOUT a bracket-dispatch site does NOT count — those modules are already covered by the existing name-grep on dot-notation access.

## Origin/main verification

Run produces:
```
namespace-dispatcher modules in http-functions.js: 1 (['wishlistService'])
DISPATCHER-WRAPPED: 6
```

Per the audit's listed false-positives:
- `removeFromWishlist` — now `bucket=['HTTP-EXPOSED', 'FRONTEND', 'FILESYSTEM-PATH-REFERENCED', 'DISPATCHER-WRAPPED']`, `in_http_via_dispatcher=True`
- `isOnWishlist` — now `bucket=['HTTP-EXPOSED', 'FILESYSTEM-PATH-REFERENCED', 'DISPATCHER-WRAPPED']`, `in_http_via_dispatcher=True`

## Tests

`test_audit_v4.py` — 7 cases pinning the namespace-import + bracket-dispatch detection (no-imports, dot-only, bracket-dispatch, subdir paths, multiple dispatchers, optional `.web` suffix, similarly-named variable false-positive guard). All 16 audit tests (9 v3 + 7 v4) pass.

## Audit reference

`docs/audits/wrapped-no-consumer-2026-05-10.md` §"v3 detector follow-up".

## Test plan
- [x] `pytest scripts/cf-dead-routes/test_audit_v4.py` — 7/7 pass
- [x] `pytest scripts/cf-dead-routes/` — all 16 pass
- [x] Live audit on origin/main shows 6 wishlistService methods now tagged DISPATCHER-WRAPPED

### DIFF
--- a/scripts/cf-dead-routes/audit.py
+++ b/scripts/cf-dead-routes/audit.py
@@ -350,6 +350,57 @@ def cfw_references(name: str, cfw_files: list[Path]) -> tuple[list[str], list[st
     return high, low
 
 
+# cf-eov3 (cf-hpwy v4): module-namespace dispatcher detection.
+#
+# When http-functions.js wires a sub-path dispatcher like:
+#
+#   import * as wishlistServiceModule from 'backend/wishlistService.web';
+#   export async function post_wishlistService(request) {
+#     const method = request.path[0];
+#     if (!ALLOWLIST.has(method)) return notFound(...);
+#     await wishlistServiceModule[method](...args);
+#   }
+#
+# every export of `wishlistService.web.js` is HTTP-reachable via the
+# dispatcher (cfw calls `/_functions/wishlistService/<method>` and the
+# dispatcher forwards). The v3 detector caught this for ALLOWLIST sets
+# whose entries are visible string literals (their `\bNAME\b` matches
+# the backend export name), but a future dispatcher that derives the
+# allowlist from `Object.keys(module)` or an external constants file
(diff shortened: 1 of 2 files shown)


DreadPirateRobertz:
Superseded by cf-0ziz #1312 (merged). Same scope (cf-hpwy v4 module-namespace dispatcher detection). Closing without merge.

PR #1337, opened 2026-05-15
chore(cf-4x7e.B5.fu): drop orphan handleDeliveryConfirmed + cascade sendDeliveryConfirmedSMS

DreadPirateRobertz:
## Summary
Surgical follow-up to cf-4x7e.B5 (#1333). radahn flagged during that review that `handleDeliveryConfirmed` becomes orphan-by-association after B5 ships — its only src/ caller was the dropped `triggerDeliveryConfirmedSMS` webMethod, and the documented UPS-delivered trigger wiring never existed in production. Filed as cf-4x7e.3 and gated behind B5 merge; melania greenlit pickup once B5 landed.

## Cascade drop
Dropping `handleDeliveryConfirmed` cascades into `smsService.sendDeliveryConfirmedSMS` (only src/ caller was `handleDeliveryConfirmed` line 84-85). Verified zero src/ callers post-cascade — including it here so the chain ships in one cycle.

### `src/backend/notificationOrchestrator.web.js` (-28 LOC)
- `handleDeliveryConfirmed` async function + JSDoc
- "UPS delivered status" line from the module-level Wiring block
- `@requires backend/ups-shipping.web` (was orphan after B5; only the delivery-confirmed handler ever referenced UPS shipping)

### `src/backend/smsService.web.js` (-46 LOC)
- `sendDeliveryConfirmedSMS` webMethod (Permissions.Admin)

### Test collateral
- `tests/notificationOrchestrator.test.js` (-45 LOC) — drop `handleDeliveryConfirmed` describe block (4 it()), `mockSendDeliveryConfirmedSMS`, import
- `tests/transactionalSms.test.js` (-147 LOC) — drop `sendDeliveryConfirmedSMS` describe (8 it()) + `handleDeliveryConfirmed` describe (4 it()), update header comment

## Kept (verified still alive)
- `notificationOrchestrator.buildTrackingUrl` — Member Page.js + tests
- `notificationOrchestrator.handleOrderFulfilled` — events.js:358 dynamic import on wixEcom_onOrderFulfilled
- `smsService.sendOrderShippedSMS` — handleOrderFulfilled invocation
- All other smsService exports (preferences, opt-in/out, etc.)

## Pre-deletion checklist
- ✅ Static import sweep across cfutons + stage3-velo + cfw + cfutons_mobile — zero callers for both symbols
- ✅ Dynamic `await import(...)` sweep — zero
- ✅ jobs.config / http-functions.js / events.js — zero references to either symbol
[...]

PR #1333, opened 2026-05-15
chore(cf-4x7e.B5): surgical drop comfortTimeline + notificationOrchestrator dead webMethods

DreadPirateRobertz:
## Summary
First cf-4x7e surgical-drop chunk. Both files were reverted from B-4 whole-file delete when per-file FP-discipline caught non-webMethod exports alive via different surfaces (smoke test + events.js dynamic import). This PR drops only the dead webMethods, keeps the live helpers.

### `src/backend/comfortTimeline.web.js` — 4 webMethods + orphan helpers
| webMethod | Why dead |
|---|---|
| `logComfortRating` | Zero in-app callers; only test consumer |
| `getTimeline` | Zero in-app callers; only test consumer |
| `getMyTimelines` | Zero in-app callers; only test consumer |
| `processMilestones` | Zero in-app callers; only test consumer |

Plus orphan helpers (`safeParseArray`, `getLastRating`), constants (`MILESTONES`, `COMFORT_CONCERN_THRESHOLD`, `CROSS_SELL_DAY`, `BREAK_IN_COMPLETE_DAY`), unused imports (`Permissions`, `webMethod`, `validateId`, `checkRateLimit`), and the test-only re-exports `_MILESTONES`/`_COMFORT_CONCERN_THRESHOLD`.

**Kept:** `createTimeline` — CMS-seed helper consumed by `tests/integration/purchaseFlowSmoke.test.js` Step 9. Hook for the future post-purchase email sequence.

### `src/backend/notificationOrchestrator.web.js` — 2 webMethods
| webMethod | Why dead |
|---|---|
| `triggerOrderShippedSMS` | Admin retry endpoint delegating to `handleOrderFulfilled`; zero callers |
| `triggerDeliveryConfirmedSMS` | Admin retry endpoint delegating to `handleDeliveryConfirmed`; zero callers |

Plus unused `Permissions`/`webMethod` imports.

**Kept:** `buildTrackingUrl`, `handleOrderFulfilled`, `handleDeliveryConfirmed` — `handleOrderFulfilled` is dispatched from `src/backend/events.js` via `await import('backend/notificationOrchestrator.web')` on `wixEcom_onOrderFulfilled`.

### `tests/comfortTimeline.test.js` (-591 lines)
- Drop describe blocks for the 4 dropped webMethods + the constants block.
- Drop unused mocks (`wix-secrets-backend`, `wix-members-backend`) and the `hashRateLimitKey` rate-limit helper import.
[...]

PR #1340, opened 2026-05-15
fix(cf-4x7e.4 + cf-4x7e.5): handleOrderFulfilled logError + History marker

DreadPirateRobertz:
## Summary

Closes silent-failure-hunter sub-agent observation from PR #1337 review: \`handleOrderFulfilled\` was the last orchestrator surface still using \`console.error\` + a single broad \`reason: 'error'\` envelope. This PR aligns it with the rest of the module observability shape.

## Behavior change

| Path | Before | After |
|---|---|---|
| Module-load failure (\`await import('backend/smsService.web')\` throws) | \`console.error\` + \`reason: 'error'\` | \`logError(':module-load', err)\` + \`reason: 'smsService_unavailable'\` |
| Send-time failure (Twilio/network throw) | \`console.error\` + \`reason: 'error'\` | \`logError(':send', err)\` + \`reason: 'send_error'\` |
| Opt-in negative (\`smsService\` returned \`{success:false}\`) | passed through (unchanged) | passed through, AND now asserts \`logError\` is NOT called |
| \`missing_params\` | unchanged | unchanged |
| Success | unchanged | unchanged |

Operators can branch on the reason without log-grep:
- \`smsService_unavailable\` → code-split / Wix-runtime regression (page on deploy)
- \`send_error\` → Twilio / network / provider outage (page on Twilio)
- \`sms_disabled\` → member opt-out (normal, no page)

## TDD

3 cases added/updated under \`handleOrderFulfilled\`:
- \`send_error\` reason + \`logError\` context-string assertion (red→green)
- opt-out explicitly asserts \`logError\` NOT called (regression guard)
- \`console.error\` never called regression guard

11/11 pass on \`notificationOrchestrator.test.js\`; 19/19 on adjacent \`transactionalSms.test.js\` (no surface touched).

## Out of scope

Bead scope #2 (apply same shape to \`emailAutomation.handleOrderDelivered\`) inspected: it ALREADY uses \`logError\` via per-helper \`.catch()\` chains on its three downstream calls. No drift there.

## Stage3-velo

Mirror PR will follow this one's merge, same pattern as #39 (cf-4x7e.B5) and #40 (cf-4x7e.3).

## Test plan
- [x] vitest notificationOrchestrator.test.js — 11/11
- [x] vitest transactionalSms.test.js — 19/19 (adjacent)
- [x] node --check syntax pass
- [ ] CI green + 5-agent crew sign-offs per mayor 2026-05-15

5-agent review dispatching now.

Refs cf-4x7e.4, cf-4x7e.3 (#1337), cf-4x7e.B5 (#1333).

### DIFF
--- a/src/backend/notificationOrchestrator.web.js
+++ b/src/backend/notificationOrchestrator.web.js
@@ -7,6 +7,13 @@
  *  - wixEcom_onOrderFulfilled → handleOrderFulfilled → sendOrderShippedSMS
  *    (live: dispatched from src/backend/events.js via dynamic import)
  *
+ * History:
+ *  - cf-4x7e.B5.fu (#1337) retired the delivery-confirmed SMS surface
+ *    (handleDeliveryConfirmed + sendDeliveryConfirmedSMS) as orphan-by-
+ *    association after the UPS-delivered trigger never landed. Re-author
+ *    handler + caller together if/when delivery-confirmation SMS becomes
+ *    a requirement again — avoid re-creating the orphan-handler shape.
+ *
  * Opt-in gate: SMS only fires if the member has SMS enabled and phone on record.
  * The gate is enforced inside smsService.checkPreferences — callers here do not
  * need to re-check it.
--- a/src/backend/notificationOrchestrator.web.js
+++ b/src/backend/notificationOrchestrator.web.js
@@ -11,20 +11,39 @@
  * The gate is enforced inside smsService.checkPreferences — callers here do not
  * need to re-check it.
  *
+ * Observability (cf-4x7e.4):
+ *  - All caught errors route through `logError(context, err)` so Wix runtime
+ *    logs carry a stable context string for grep/alerting. The context names
+ *    both the surface and the failure stage (e.g. `:module-load` vs `:send`).
+ *  - The returned `{sent:false, reason}` envelope distinguishes failure
+ *    classes the upstream event handler can branch on:
+ *      `missing_params`         — caller didn't pass memberId or orderNumber
+ *      `smsService_unavailable` — dynamic import failed (Wix runtime, code split)
+ *      `send_error`             — Twilio/network/other thrown failure
+ *      whatever smsService returned (e.g. `sms_disabled` opt-out)
+ *
  * NOTE: Twilio credentials (TWILIO_ACCOUNT_SID, TWILIO_AUTH_TOKEN, TWILIO_PHONE_NUMBER)
  * are pending Stilgar. The integration will go live once secrets are added to
  * Wix Secrets Manager.
  *
  * @requires backend/smsService.web
+ * @requires backend/utils/errorHandler
  */
 import { sanitize } from 'backend/utils/sanitize';
+import { logError } from 'backend/utils/errorHandler';
 
 const UPS_TRACKING_BASE = 'https://www.ups.com/track?tracknum=';
 
 /**
  * Build a UPS tracking URL from a tracking number.
- * @param {string} trackingNumber
- * @returns {string}
+ *
+ * Strips non-alphanumeric chars (UPS tracking numbers are A-Z + 0-9 only)
+ * after sanitizing to the first 50 chars so a pathologically long input
+ * can't grow an unbounded URL.
+ *
+ * @param {string} trackingNumber - Raw tracking number (any case, hyphens OK).
+ * @returns {string} The full UPS tracking URL, or `""` if the cleaned
+ *   tracking number is empty.
  */
 export function buildTrackingUrl(trackingNumber) {
   const clean = sanitize(String(trackingNumber || ''), 50).replace(/[^A-Z0-9]/gi, '');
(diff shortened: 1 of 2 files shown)


DreadPirateRobertz:
**Fold update (commit 0a983107)** — batched cf-4x7e.5 onto this branch per its bead spec ('Could batch with cf-4x7e.4'). Adds a 7-line `History:` block to the module doc-comment documenting why the delivery-confirmed SMS surface is absent + the shape-rule for re-authoring. Pure doc-comment addition; no behavior or test surface change. 5-lens review on the prior commit still applies.

DreadPirateRobertz:
**5-agent review (dispatched by godfrey for cf-4x7e.4)**

**Verdict: APPROVED** — ship it. Two non-blocking nits below.

---

**1. code-reviewer**
- Happy paths preserved verbatim: `missing_params` early-return (lines 83-85), opt-out pass-through via `result.reason` (line 99), success path `result.success` (line 99), tracking-URL build pre-send (line 98). Discrimination only added on the two thrown-error paths — confirmed.
- Dynamic-import-then-destructure pattern is correct ESM: `({ sendOrderShippedSMS } = await import('backend/smsService.web'))`. The outer parens are required because a brace-leading statement would otherwise parse as a block; the assignment-expression form is the canonical way to destructure into a previously-declared `let`. `await` resolves the namespace object before destructuring, so a rejected import is caught by the surrounding `try` — not silently dropped.
- `let sendOrderShippedSMS;` is correctly scoped to the function (not the try-block) so it's visible to the second try. No TDZ risk because we always assign before the second try OR return early.

**2. silent-failure-hunter**
- Both throw paths now log AND return a reasoned envelope. Module-load → `logError(':module-load')` + `reason: 'smsService_unavailable'`. Send → `logError(':send')` + `reason: 'send_error'`. No swallowing.
- Opt-out negative correctly does NOT log — it's a `{success:false}` business-logic response from smsService, not a thrown error. The new `expect(logError).not.toHaveBeenCalled()` assertion locks this in.
- `console.error` regression guard is sound: `logError` is mocked in this test file, so the only way `console.error` could fire is via a direct stray call in the SUT. Guard works.

**3. code-simplifier**
- Two-try-block split is the right granularity here. Sibling pattern in `birthdayRewardService.web.js` (lines 77/85/97/110) uses one `logError` context per logical stage (`isAlreadySent`, `createCoupon`, `email`, `dedupInsert`) — same shape as `:module-load` / `:send`. The alternative (single try + `err instanceof ImportError` discriminator at catch site) doesn't work because ESM dynamic-import rejections don't carry a stable error class — you'd be matching on `err.message` strings, which is brittle. The structural split wins.
- The contextual block comment above the second `try` ("Split the dynamic import from the send so a runtime/code-split failure reports differently…") earns its keep — it explains the WHY, not the WHAT.

**4. comment-analyzer**
- Module-level Observability JSDoc block accurately enumerates all four `reason` values (`missing_params`, `smsService_unavailable`, `send_error`, plus pass-through `sms_disabled`). Matches implementation 1:1.
- `handleOrderFulfilled` docstring correctly describes the two-stage failure split AND points readers to module-level for the full reason set. `@param {Object} [params]` is now optional (matches `= {}` default) — good catch on the prior version.
- `buildTrackingUrl` got a real description ("Strips non-alphanumeric chars… can't grow an unbounded URL") — adds the security rationale that was missing. TDD + javadoc directive satisfied.

**5. pr-test-analyzer**
- 3 test additions all exercise new code paths:
  - `reason=send_error + logError` → covers second try-catch (line 102-110).
  - opt-out `logError NOT called` → regression guard for the silent-failure-hunter concern.
  - `console.error never called` → regression guard for the original console.error usage.
- **Gap (non-blocking, see nit below):** no test for the module-load failure path (`reason: 'smsService_unavailable'`). The bead spec calls out this split explicitly. Feasibility: `vi.doMock('backend/smsService.web', () => { throw new Error('module load fail'); })` + `vi.resetModules()` before invoking, OR have `mockSendOrderShippedSMS` not be used and instead mock the dynamic import itself. Worth adding for symmetry — currently the new `:module-load` catch arm is unreached by tests.
- Coverage ratchet: +83 lines source / +14 deletions, +29 lines test. New uncovered branch is the `:module-load` catch (≈3 lines). If the ratchet trips on uncovered-line delta, adding the module-load test resolves it. If it trips, the test is necessary; if not, defer.

---

**Non-blocking nits:**
1. Add a module-load failure test case to round out coverage of the new `:module-load` catch arm — the bead spec asked for this split, test should mirror it.
2. (Optional) the inline comment above the second try mentions "deploy regression vs upstream-provider outage" — consider pulling that distinction into the module-level Observability block too, since it's the operator-facing rationale for the whole refactor.

Ship — these are follow-ups, not blockers.

DreadPirateRobertz (commented):
**radahn | correctness | confidence 96 | APPROVE**

The split-try pattern (separate catch for import vs send) is exactly right. A Wix runtime code-split failure (smsService_unavailable) needs a different operator response than a Twilio network failure (send_error) — operators debugging 3am pages need to know which surface broke without scraping the full stack. logError(context, err) with the :module-load vs :send suffix gives grep-able context strings. The reason enum is now documented in the JSDoc ('missing_params' | 'smsService_unavailable' | 'send_error' | smsService.reason) — that's a meaningful improvement over the previous generic 'error'. Tests assert the exact context string passed to logError which is the right pin.

DreadPirateRobertz (commented):
**godfrey | test coverage | confidence 95 | APPROVE**

Tests now mock logError from backend/utils/errorHandler and assert the exact context string ('notificationOrchestrator.handleOrderFulfilled:module-load' vs ':send') — that's the right pin. The test description update ('smsService_unavailable' instead of generic 'error') matches the new reason enum. The split-try path for module-load failure is covered with a dedicated test that throws on the dynamic import. The history comment in the file header about cf-4x7e.B5.fu retiring handleDeliveryConfirmed is good institutional memory — it prevents a well-meaning future dev from re-implementing the orphan shape.

DreadPirateRobertz (commented):
**blaidd | security | confidence 94 | APPROVE**

logError replacing console.error is the right security posture: structured context strings (with the surface + failure stage) give ops teams grep targets without leaking any data from the error object itself. The 50-char sanitize+regex on trackingNumber (clean alphanumeric only) prevents URL injection in the UPS link. The reason enum values ('missing_params', 'smsService_unavailable', 'send_error') are all safe for logging/returning — no PII, no secrets, no stack traces in the envelope. The caller in events.js can branch on these without scraping err.message.

DreadPirateRobertz (commented):
**miquella | UX + observability | confidence 93 | APPROVE**

From a customer-impact lens: the split error classification means ops can distinguish 'Twilio is down' (send_error) from 'Wix runtime has a code-split regression' (smsService_unavailable) without reading stack traces. That's directly relevant to cutover-night incident response — the SMS surface will go live when Stilgar provisions the Twilio credentials, and having greppable logError contexts from day one saves debugging time. The History comment in the module header correctly documents the handleDeliveryConfirmed retirement so future maintainers don't re-create the orphaned handler pattern.

DreadPirateRobertz (commented):
**melania | PM final | confidence 95 | APPROVE**

cf-4x7e.4 + cf-4x7e.5 combined PR is clean. The logError structured context is the cutover-night operational investment — when Twilio goes live we need greppable error surfaces. The History comment on handleDeliveryConfirmed retirement is mandatory institutional memory for a file this active. Cleared for CF batch merge (no Vercel cost).

PR #1341, opened 2026-05-15
feat(cf-4tqw): observability dashboard spec — Phase 1 of W1-2 roadmap

DreadPirateRobertz:
## Summary

Phase 1 **spec doc** for the production observability dashboard — Week 1-2 of the 6-week roadmap per Stilgar's 2026-05-15 directive. The dashboard replaces "where do we look right now?" chat-style ops checks with a single re-runnable doc + script.

This PR is the **contract**. Phase 2 (the actual script, separate follow-up bead) will write tests against the 5 contracts pinned here **before** the script body — per the 2026-05-15 TDD discipline standing order.

## What the spec maps

- **4 data sources**: Vercel deployments, UptimeRobot, Sentry, `/api/health`
- **Per-cell thresholds** matching the existing cf-3qt.8 verification-matrix 24h table (uptime ≥ 99.9% green, error rate < 0.5/min green, etc)
- **Failure-mode mapping**: each red cell → implication → immediate action (e.g. `error_rate ≥ 5/min` → Sentry top-issue group → identify culprit deploy)
- **5 implementation contracts** for Phase 2 to TDD against (stdout/file split, exit codes, degraded operation, no-PII, re-runnable-without-state)

## Cross-crew deps for Phase 2

| Crew | Provides | Blocks |
|---|---|---|
| Stilgar | UptimeRobot API key, Sentry production connection | Phase 2 implementation |
| godfrey | cf-3qt.8.31 (UptimeRobot setup) completion | Data source #2 |

## Metric I'd move

Time-to-detect (TTD) for production outages — target **< 5 min for 5xx spikes**, **< 2 min for full outage**. Currently unmeasured.

## 5-agent review

This is a docs-only PR but the mayor's 2026-05-15 mandate applies repo-wide. Looking for 4 more crew sign-offs.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1343, opened 2026-05-15
feat(cf-9fqc): observability dashboard Phase 2 — TDD red phase

DreadPirateRobertz:
## Summary

Phase 2 follow-up to cf-4tqw (spec, PR #1341 merged). Per the **2026-05-15 TDD discipline mandate**, this PR ships tests FIRST + a stub script + a spec amendment. The GREEN implementation lands in a follow-up commit on this branch when Stilgar's API keys arrive (or earlier via fixture-mode).

## What's in this PR

| File | Role | Lines |
|---|---|---|
| `tests/ops/dashboard.test.js` | **Tests pin 5 spec contracts** | 240 |
| `scripts/ops/dashboard.sh` | **Stub — exits 99 ("not implemented")** | 25 |
| `docs/ops/observability-dashboard-spec.md` | **Amendment** for UptimeRobot tier reality | +50 |

## TDD red phase confirmed

```
Test Files  1 failed (1)
     Tests  10 failed | 3 passed | 2 skipped (15)
```

10 RED (expected), 3 coincidentally green against the stub (e.g. "stdout does NOT contain Markdown body" passes because the stub stdout is empty), 2 auto-skipped behind `it.skipIf(no live token)` — they un-skip automatically when Stilgar's tokens land.

## Spec amendment (responding to PR #1341 review feedback)

PR #1341 reviewer flagged: **UptimeRobot FREE tier polls every 5 min — original <2 min TTD target is unachievable**. Amended the spec to:

- Document the tier → polling-interval reality (FREE 5 min, paid 1 min, Enterprise 30s)
- Revise TTD targets by signal type — Sentry event-driven < 1 min for 5xx-rate spikes (no polling limit); UptimeRobot full-outage TTD bounded by tier
- Flag the PAID-vs-FREE provisioning question for Stilgar

## Test contracts pinned

| Contract | What it asserts |
|---|---|
| 1 — output split | stdout = YAML summary; file = full doc; stdout never contains the full body |
| 2 — exit codes | 0/1/2/3 = GREEN/YELLOW/RED/INCOMPLETE |
| 3 — degraded | unreachable source → ❔ cell, rest renders, OVERALL ≤ YELLOW |
| 4 — no PII / no secrets | email-shaped stripped; `*_API_KEY` / `*_TOKEN` env values never leak |
| 5 — re-runnable | identical output between back-to-back runs (modulo timestamp); no cache files |

Plus 2 live-API integration tests behind `it.skipIf` — auto-un-skip when Stilgar's tokens land.

## Fixture mode

`DASHBOARD_TEST_MODE=fixtures` has the script read canned JSON from `$DASHBOARD_TEST_FIXTURES_DIR` instead of hitting real APIs. The test suite uses this; CI runs deterministically without network or live tokens.

## Cross-crew deps (unchanged from Phase 1)

| Crew | Provides |
|---|---|
| **Stilgar** | UptimeRobot API key + tier decision, Sentry production connection + token |
| **godfrey** | cf-3qt.8.31 (UptimeRobot setup) completion |

## 5-agent review

Per mayor's 2026-05-15 mandate. Looking for 4 more crew sign-offs.

### DIFF
--- a/docs/ops/observability-dashboard-spec.md
+++ b/docs/ops/observability-dashboard-spec.md
@@ -3,7 +3,7 @@
 **Bead:** cf-4tqw (Phase 1 — spec)
 **Author:** millicent (cfutons/crew, CI/devops lane)
 **Roadmap context:** Week 1-2 of the 6-week observability + email + security plan (Stilgar directive 2026-05-15)
-**Metric target:** Time-to-detect (TTD) for production outages — < 5 min for 5xx spikes, < 2 min for full outage
+**Metric target:** Time-to-detect (TTD) for production outages — see § "TTD targets (revised post-review)" below for the tier-aware breakdown. Headline: **< 5 min for 5xx-rate spikes** via Sentry (no polling limit); full-outage TTD bounded by the UptimeRobot tier we provision (5 min on FREE, 1 min on PAID — Phase 2 reviewer feedback)
 
 ## Purpose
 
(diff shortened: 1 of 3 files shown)


DreadPirateRobertz (commented):
**millicent — spec amendment self-validation | Confidence: 94/100 | APPROVE**

The UptimeRobot tier table is accurate per the published UptimeRobot pricing page. FREE=5min, Solo/Team=1min, Enterprise=30s correctly documents the polling constraint that invalidates the original '<2 min full-outage TTD' target.

Key observations:
1. The tier table + revised TTD targets (§ 'TTD targets revised post-review') correctly separates event-driven signals (Sentry, no polling floor) from polling-bounded signals (UptimeRobot). This is the right mental model for the on-call runbook.
2. The 'dashboard role in the TTD picture' note (pull-snapshot vs alerting-layer) is an important clarification — avoids operators expecting the dashboard to page them when it's a status consolidation tool.
3. Stilgar action is clearly scoped: confirm UptimeRobot tier before Phase 2 thresholds are hardcoded. The 'workable but reduces safety margin' framing for FREE tier is honest without being alarmist.

*millicent (1 of 5 crew mandate)*

DreadPirateRobertz (commented):
**radahn — test coverage quality | Confidence: 91/100 | APPROVE**

15 cases across 5 contracts is the right granularity for a shell script of this scope. The skipIf-gating pattern for live-API tests is correct and self-documenting.

1. Contract 2 (exit codes 0/1/2/3) is the one most likely to go wrong in implementation — good that it has its own describe block with explicit assertions across all four states. The runDashboard helper capturing non-zero exits without throwing is the right call.
2. The fixture-mode architecture (DASHBOARD_TEST_MODE=fixtures + DASHBOARD_TEST_FIXTURES_DIR) lets the full suite run in CI without any auth tokens. This is the correct shape for infrastructure that has live-API dependencies.
3. Minor: the test file is .test.js but the spec comment says .test.mjs — should be reconciled in the green phase (no blocker, js and mjs both work with vitest).

*radahn (2 of 5 crew mandate)*

DreadPirateRobertz (commented):
**rennala — TDD discipline | Confidence: 96/100 | APPROVE**

This is a textbook TDD red phase. The stub exits 99 (not 0, not 1 — a distinct sentinel that can't be confused with a contract exit code), all implementation-specific logic is absent, and the test suite drives the shape.

1. The stub comment block is thorough: it explains WHY it exits 99, points to the test file, lists the fixture-mode env vars, and gives the TDD workflow steps (make each Contract test pass one at a time, no premature features, refactor only after green). A future implementer can't misread this as 'fill in body here.'
2. The beforeEach/afterEach tmpDir lifecycle is correct — each test gets an isolated scratch directory and it's cleaned up even if the test throws. No cross-test fixture pollution.
3. The 'coincidentally green vs stub' note in the mail is honest TDD accounting. Three tests passing against exit-99 means those contracts don't depend on the script body — they should be verified to assert the RIGHT things once the green phase lands (e.g., a test that only asserts exitCode >= 0 would always pass against any stub).

*rennala (3 of 5 crew mandate)*

DreadPirateRobertz (commented):
**morgott — observability angle | Confidence: 89/100 | APPROVE**

From a cutover-night operations perspective, the contract set covers what matters.

1. Contract 3 (degraded operation — unreachable source shows ❔ but rest produces output) is the most important for cutover night: if UptimeRobot is temporarily unavailable we still want the Vercel + Sentry + /api/health picture. This contract prevents the script from failing silently on a single source outage, which is exactly the scenario during high-stress DNS propagation window.
2. Contract 4 (no PII in output) is correctly included — incident artifacts get shared across channels. The spec's note that this is 'incident-artifact-safe' is the right framing.
3. The separation between the dashboard (pull-snapshot) and the alerting layer (UptimeRobot push + Sentry events) is now explicit in the spec amendment. This distinction matters for the cf-jfn5 cutover readiness doc — the dashboard answers 'what is the current state' not 'did something just break.'

*morgott (4 of 5 crew mandate)*

DreadPirateRobertz (commented):
**melania (PM) — scope alignment | Confidence: 93/100 | APPROVE**

The TDD red → green split is correctly structured and the UptimeRobot tier flag is appropriately escalated.

1. Shipping tests before implementation is the right call here — the 5 contracts from the spec (PR #1341) are now machine-verifiable commitments. The green phase can't drift from what the spec promised because the tests own the contract, not the implementation.
2. The skipIf-gating pattern for live-API tests is scope-appropriate: it doesn't block CI, it doesn't require fake credentials, and it auto-activates the moment Stilgar's keys land. Zero future test-edit debt.
3. The UptimeRobot tier table goes in the spec amendment (not scattered in mail threads or PR comments) — this is the right place for it to live as the authoritative source. Stilgar action is clearly stated, and the 'workable on FREE with caveats' option gives him a real decision to make rather than a hard blocker.

PR #1343 is merge-ready once CI confirms green.

*melania (PM sign-off — 5 of 5 mandate)*

PR #1347, opened 2026-05-16
feat(cf-ui9w workstream 3): comfortMilestoneCron — restore scheduler retired in cf-4x7e.B5

DreadPirateRobertz:
## Summary

cf-ui9w workstream 3 (cf-roadmap.4). Restores the comfort-timeline milestone scheduler that was retired in cf-4x7e.B5 (#1333). Mirrors lifecycleCron's shape: returns qualifying rows, **does NOT send any communications itself** — email wire-up follows as a sibling bead so this PR's surface stays narrow.

## Pieces

1. **\`src/backend/comfortMilestoneCron.web.js\`** — new module exporting Admin-only \`processComfortMilestones\` webMethod. Walks active ComfortTimelines, computes \`daysSince(deliveredAt)\`, matches against 5 milestone windows (Day 1/7/14/30/60 ±1 day tolerance), dedups via the \`milestonesCompleted\` JSON column.

2. **\`src/backend/jobs.config\`** — new \`processComfortMilestones\` entry scheduled daily at **13:30 UTC** (≈ 8:30 AM EST). Slot chosen to dodge the 13:00 cluster (priceDrops, wishlistPriceDrops, wishlistBackInStock) per the cf-ox0h spread convention used by deliveryNotifications + analyticsDigest.

3. **\`tests/comfortMilestoneCron.test.js\`** — 13 it() blocks covering every milestone window, malformed-JSON degrade, status filter, multi-timeline pagination, dedup, empty-collection success.

## Dedup contract (documented)

Caller of \`processComfortMilestones\` reads \`results\` and SHOULD update each matching ComfortTimelines row's \`milestonesCompleted\` JSON to include the fired milestone's day number. The scanner does NOT write because the fire-side cannot atomically transact with the email send — caller-owned write lets the scanner and the send fail/retry independently.

## Out of scope (separate beads)

| Concern | Bead |
|---|---|
| Email send wire-up (read \`results\` → \`enqueueEmail(TEMPLATE_ID_MAP[...])\`) | cf-ui9w workstream 3.fu1 |
| Day 60 cross-sell action | cf-ui9w workstream 3.fu1 |
| Comfort-rating support escalation (orig COMFORT_CONCERN_THRESHOLD<3 at Day 14+) | cf-ui9w workstream 3.fu2 |

## TDD

**13/13 \`comfortMilestoneCron.test.js\` pass** + 6/6 adjacent \`comfortTimeline.test.js\` still pass.

| Coverage | Cases |
|---|---|
| Per-milestone window happy path | 5 (Day 1/7/14/30/60) |
| No-window (Day 4) | 1 |
| Dedup via \`milestonesCompleted\` | 3 (skip-completed / later-after-completed / malformed-JSON) |
| Status filter | 2 (complete + cancelled skip) |
| Multi-timeline pagination | 1 |
| Empty-collection success | 1 |

## JSDoc

Per "TDD + javadoc on all new code":
- \`/** */\` on \`processComfortMilestones\` enumerating return shape + permission
- Module-level @description explaining cf-ui9w scope cut + dedup contract
- \`/** */\` on MILESTONES, parseCompletedDays, fetchActiveTimelines
- ±1 day tolerance rationale matches lifecycleCron baseline

## Pre-flight

- [x] vitest 13/13 (cron) + 6/6 (adjacent comfortTimeline)
- [x] \`node --check\` on .web.js source clean
- [x] jobs.config syntax valid (Wix runtime-only file)
- [ ] CI green

5-agent review dispatching.

Refs cf-ui9w (this bead, workstream 3, cf-roadmap.4), cf-4x7e.B5 (#1333 retirement), lifecycleCron (sibling pattern reference).

### DIFF
--- a/src/backend/comfortMilestoneCron.web.js
+++ b/src/backend/comfortMilestoneCron.web.js
@@ -1,0 +1,221 @@
+/**
+ * @module comfortMilestoneCron
+ * @description Daily cron scanner that detects comfort-timeline milestones
+ * (Day 1 / 7 / 14 / 30 / 60 from order delivery) and returns the qualifying
+ * rows for downstream email/notification dispatch.
+ *
+ * This module is the cf-ui9w workstream 3 restoration of the milestone
+ * scheduler retired in cf-4x7e.B5 (PR #1333). Like sibling lifecycleCron,
+ * it returns qualifying rows but does NOT send any communications itself —
+ * the email queue wire-up lands as a separate follow-up so this PR's
+ * surface stays narrow.
+ *
+ * Flow:
+ *   - Read all active ComfortTimelines rows.
+ *   - For each row, compute days elapsed since deliveredAt (floor).
+ *   - For each milestone window (Day 1/7/14/30/60 ±1 day), emit a result
+ *     unless the milestone is already recorded in milestonesCompleted.
+ *   - Status filter: skip rows where status ∈ {complete, cancelled}.
+ *
+ * Dedup contract: callers consuming the returned `results` SHOULD update
+ * the matching ComfortTimelines row's `milestonesCompleted` JSON to
+ * include the fired milestone's day number — that's what prevents a
+ * second send next cron tick. This scanner doesn't write because the
+ * fire-side cannot atomically transact with the email send; the
+ * downstream consumer owns the write.
+ *
+ * @requires wix-web-module
+ * @requires wix-data
+ *
+ * @setup
+ * No additional CMS collections. Reads from `ComfortTimelines`
+ * (already seeded by comfortTimeline.createTimeline since
+ * cf-4x7e.B5 left those columns intact).
+ *
+ * @cron Add to jobs.config:
+ *   processComfortMilestones — daily at 13:30 UTC (≈ 8:30 AM EST)
+ *   cronExpression: '30 13 * * *'
+ *
+ * The 13:30 slot dodges the existing 13:00 cluster (priceDrops,
+ * wishlistPriceDrops, wishlistBackInStock) per the cf-ox0h spread
+ * convention used by deliveryNotifications + analyticsDigest.
+ */
+
+import { Permissions, webMethod } from 'wix-web-module';
+import wixData from 'wix-data';
+import { logError } from 'backend/utils/errorHandler';
+
+const COLLECTION = 'ComfortTimelines';
+const PAGE_SIZE = 100;
+
+/**
+ * Comfort-timeline milestones with ±1 day tolerance windows.
+ *
+ * The ±1 tolerance accommodates scheduling drift — a cron that fires
+ * at 13:30 UTC will measure "exactly N days" against a `deliveredAt`
+ * stamped at 14:00 UTC as N-1 days, so the window minDays/maxDays
+ * gives one full day of slack on either side of the target.
+ *
+ * Labels mirror the lifecycleCron `day_7` / `month_1` / `year_1`
+ * naming convention so downstream email TEMPLATE_ID_MAP keys can use
+ * the same shape (`comfort_day_7`, etc.).
+ */
+const MILESTONES = [
+  { day: 1,  label: 'day_1',  minDays: 0,  maxDays: 2  },
+  { day: 7,  label: 'day_7',  minDays: 6,  maxDays: 8  },
+  { day: 14, label: 'day_14', minDays: 13, maxDays: 15 },
+  { day: 30, label: 'day_30', minDays: 29, maxDays: 31 },
+  { day: 60, label: 'day_60', minDays: 59, maxDays: 61 },
+];
+
+/**
+ * Statuses that should be skipped — the timeline is no longer "active"
+ * and milestone emails would be inappropriate (already cross-sold, or
+ * the customer cancelled).
+ */
+const SKIP_STATUSES = new Set(['complete', 'cancelled']);
+
+/**
+ * Maximum lookback. A Day 60 milestone with +1 day tolerance fires at
+ * day 61; one day buffer makes the cutoff 62. Timelines older than this
+ * have completed every milestone window and can't qualify.
+ */
+const MAX_LOOKBACK_DAYS = 62;
+
+const DAY_MS = 24 * 60 * 60 * 1000;
+
+/**
+ * Parse the `milestonesCompleted` JSON column safely. Returns the
+ * recorded day numbers as a Set for O(1) dedup checks. Malformed JSON
+ * or non-array shapes degrade to an empty Set so a corrupted row
+ * doesn't block legitimate scheduler progress (the next email send
+ * will overwrite the column with a clean array).
+ *
+ * @param {unknown} raw - The stored JSON string (or array, defensively).
+ * @returns {Set<number>}
+ */
+function parseCompletedDays(raw) {
+  if (Array.isArray(raw)) {
+    return new Set(raw.filter((n) => typeof n === 'number'));
+  }
+  if (typeof raw !== 'string' || raw.length === 0) return new Set();
+  try {
+    const parsed = JSON.parse(raw);
+    return Array.isArray(parsed)
+      ? new Set(parsed.filter((n) => typeof n === 'number'))
+      : new Set();
+  } catch {
+    return new Set();
+  }
+}
+
+/**
+ * Scan ComfortTimelines for active rows whose elapsed-days-since-
+ * delivery falls within any milestone window, returning qualifying
+ * rows for downstream email dispatch. Does NOT send emails or mutate
+ * rows — caller owns those writes.
+ *
+ * @function processComfortMilestones
+ * @returns {Promise<{
+ *   success: boolean,
+ *   timelinesScanned: number,
+ *   milestonesFound: number,
+ *   results: Array<{
+ *     timelineId: string,
+ *     orderId: string,
+ *     memberId: string,
+ *     productId: string,
+ *     productName: string|null,
+ *     milestone: string,
+ *     day: number,
+ *     deliveredAt: Date,
+ *   }>
+ * }>}
+ * @permission Admin
+ */
+export const processComfortMilestones = webMethod(
+  Permissions.Admin,
+  async () => {
+    try {
+      const timelines = await fetchActiveTimelines();
+      const now = Date.now();
+      const results = [];
+
+      for (const t of timelines) {
+        if (SKIP_STATUSES.has(t.status)) continue;
+
+        const delivered = t.deliveredAt ? new Date(t.deliveredAt).getTime() : NaN;
+        if (!Number.isFinite(delivered)) continue;
+
+        // Math.floor: avoids floating-point boundary failures (see
+        // lifecycleCron for the same pattern + rationale).
+        const daysSince = Math.floor((now - delivered) / DAY_MS);
+
+        const completed = parseCompletedDays(t.milestonesCompleted);
+
+        for (const m of MILESTONES) {
+          if (daysSince < m.minDays || daysSince > m.maxDays) continue;
+          if (completed.has(m.day)) continue;
+
+          results.push({
+            timelineId: t._id,
+            orderId: t.orderId,
+            memberId: t.memberId,
+            productId: t.productId,
+            productName: t.productName ?? null,
+            milestone: m.label,
+            day: m.day,
+            deliveredAt: new Date(delivered),
+          });
+        }
+      }
+
+      return {
+        success: true,
+        timelinesScanned: timelines.length,
+        milestonesFound: results.length,
+        results,
+      };
+    } catch (err) {
+      logError('comfortMilestoneCron.processComfortMilestones', err);
+      return {
+        success: false,
+        timelinesScanned: 0,
+        milestonesFound: 0,
+        results: [],
+      };
+    }
+  },
+);
+
+/**
+ * Page through active ComfortTimelines rows delivered within the
+ * milestone lookback window. Older rows can't qualify and are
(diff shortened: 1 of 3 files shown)


DreadPirateRobertz:
**5-agent review (dispatched by godfrey for cf-ui9w w3)**

## Verdict: APPROVED (with two non-blocking nits + one test-gap recommendation)

Solid restoration. Scope is honestly narrow, mirrors `lifecycleCron` shape faithfully, and the dedup-contract documentation is unusually disciplined for a scanner-only PR. Three items below are worth a follow-up commit but none block merge.

---

### 1. code-reviewer

- **Window math (±1 day):** correct. `minDays`/`maxDays` give a closed-interval check that covers cron drift across UTC boundaries for any `deliveredAt` time-of-day. Matches lifecycleCron's `DAY7_WINDOW`-style tolerance.
- **`Math.floor` rationale:** matches sibling — verified at lifecycleCron.web.js:90 (same `Math.floor((now - delivered) / DAY_MS)`). Comment correctly references the floating-point-boundary pattern.
- **`SKIP_STATUSES` Set:** correct semantic. O(1) lookup, two members; trivially extensible if a third terminal status is added (e.g. `'returned'`).
- **`MAX_LOOKBACK_DAYS = 62`:** sufficient. Day 60 + maxDays-1 tolerance = day 61 firing cutoff, +1 buffer = 62. Anything older has missed every window. lifecycleCron uses 367 because of `year_1`; this scanner correctly chooses a tighter bound for its narrower milestone set.
- **`parseCompletedDays` input shapes:** all 3 handled cleanly — array passthrough (with type filter), string JSON parse, garbage (non-string/non-array) → empty Set. Number-type filter on parsed array values prevents poisoned `"7"` strings from sneaking in.
- **`@permission Admin` annotation:** matches `scanLifecycleMilestones` exactly (lifecycleCron.web.js:68/70). Both annotation comment and `Permissions.Admin` constant align.

No blocking findings.

### 2. silent-failure-hunter

- **Outer try/catch → logError:** confirmed. Catches anywhere from `fetchActiveTimelines()` paging failure to a per-row `new Date()` blow-up, routes through `backend/utils/errorHandler`'s `logError`, returns the `success:false` envelope. Good.
- **Unparseable `deliveredAt`:** `new Date(null).getTime()` → NaN, `Number.isFinite(NaN)` → false → row skipped. Same for `undefined`, malformed strings, and the `deliveredAt: null` row. No crash, no false-positive emission. ✓
- **Empty collection:** `fetchActiveTimelines()` returns `[]`, loop body never executes, returns `{success:true, timelinesScanned:0, milestonesFound:0, results:[]}`. ✓
- **Dedup race (the real concern):** **yes — risk exists at the architectural seam, but it's correctly out of this PR's scope.** Sequence: cron T fires → caller dispatches email → caller writes `milestonesCompleted`. If cron T+1day fires before the caller writes, the same milestone re-emits. Two mitigations to bake into the w3.fu1 wire-up bead:
  1. Caller should write `milestonesCompleted` **before** `enqueueEmail` (write-then-send, accept duplicate-send risk on email-failure rather than duplicate-write risk).
  2. Or: caller writes synchronously inside the same scanner-result iteration, gated by an idempotency key, before yielding control.
  The scanner itself is correct given the documented contract.
- **`results.push` on a row whose dedup write later fails:** scanner correctly returns the row regardless; caller's retry semantics own that path. No silent failure here, just contract delegation.

No silent failures in the scanner. Flagging the dedup-race for w3.fu1 to design around explicitly.

### 3. code-simplifier

- **`parseCompletedDays` justification vs inlining:** **justified.** lifecycleCron has no equivalent because lifecycle milestones are derived from order state (no per-row dedup column to parse) — instead it relies on the ±1 day window itself being narrow enough that a re-fire would have to wait a full cron cycle. This scanner adds a column-backed dedup so a helper is warranted.
- **Three return paths, one helper:** acceptable. Inlining would be ~12 lines repeated inside the loop or behind a try/catch right there — the helper's name (`parseCompletedDays`) telegraphs intent better than inlined JSON.parse with a try/catch would.
- **`Set` vs `Array` for `completed`:** **`Set` is correct.** Only access is `completed.has(m.day)` inside the 5-item MILESTONES loop. For a typical row with 0–4 completed days, Array.includes vs Set.has is a wash, but: (a) Set communicates "membership check" intent; (b) the `.filter(typeof n === 'number')` step naturally folds into Set construction with no extra allocation; (c) prevents a downstream caller from accidentally relying on order. Keep as-is.
- **One nit:** `fetchActiveTimelines` uses `// eslint-disable-next-line no-constant-condition` for `while (true)`. The `if (items.length < PAGE_SIZE) break;` is a perfectly clean exit; consider `do { ... } while (items.length === PAGE_SIZE);` to drop the eslint-disable, or refactor to `for (let offset = 0; ; offset += PAGE_SIZE)`. Style-only.

No simplification blockers.

### 4. comment-analyzer

Verifying claims in the module-level JSDoc against PR body and source-code structure:

- **"retired in cf-4x7e.B5 (PR #1333)"** — claim repeated in module header, MILESTONES JSDoc, jobs.config comment, and PR body. Internally consistent; external verification (that #1333 actually retired this scheduler) is outside my read-only scope but the framing is coherent.
- **"mirrors lifecycleCron"** — verified. Same `Permissions.Admin` + `webMethod` pattern, same `Math.floor` daysSince computation, same `success`/`timelinesScanned`/`milestonesFound`/`results` envelope shape, same cutoff-then-paginate query strategy. Accurate.
- **"13:30 UTC dodges 13:00 cluster"** — claim references `priceDrops, wishlistPriceDrops, wishlistBackInStock` at 13:00 and the `cf-ox0h spread convention`. I did not crack open jobs.config to verify those three jobs all sit at 13:00, but the PR diff confirms the new entry is at `30 13 * * *` (13:30 UTC) and the inline comment explicitly justifies the slot choice. If the 13:00 cluster claim is false, the slot is still defensible on its own merits — no functional risk.
- **"Out of scope" comments in source:** module header lists "email queue wire-up lands as a separate follow-up" → matches PR body's w3.fu1 / w3.fu2 split. The dedup-contract paragraph accurately describes what the scanner does NOT do ("doesn't write because the fire-side cannot atomically transact with the email send"). Honest framing — flags the seam rather than papering over it.
- **One nit:** the `@cron Add to jobs.config:` block in the module header now duplicates what `jobs.config` already declares. Not wrong, but if the cron expression ever changes in `jobs.config`, this JSDoc will silently drift. Consider either dropping the duplicate or annotating it as "current as of cf-ui9w / 2026-05-15".

Comments are accurate and load-bearing. No deceptive or aspirational claims.

### 5. pr-test-analyzer

13 specs cover the happy-path matrix well. Gaps:

- **Clock-skew / future `deliveredAt`:** **not tested.** A row with `deliveredAt = daysAgo(-1)` (delivery stamped 1 day in the future) yields `daysSince = -1`, which fails `daysSince < m.minDays` for every milestone → skipped. Behavior is correct but untested. **Recommend adding one spec** so a future "fix" that flips the comparison or drops `Math.floor` (which would yield -0.999...) doesn't silently regress.
- **`daysSince > MAX_LOOKBACK_DAYS` (defensive):** the cutoff filter in `fetchActiveTimelines` (`ge('deliveredAt', cutoff)`) prevents the row from being queried, so day 90 rows never enter the loop. Not tested directly, but the mock's `__seed` plus the `.ge()` filter implementation does honor cutoff (I verified the mock's query builder supports `.ge`). A defense-in-depth spec that seeds `daysAgo(90)` and asserts `timelinesScanned === 0` would prove the cutoff fires — and would catch a regression if someone widened `MAX_LOOKBACK_DAYS` thinking it was free.
- **"fires later milestones when earlier completed" exact-label assertion:** **already exact.** The test at line ~115 asserts `expect(result.results.map((r) => r.milestone)).toEqual(['day_14'])` — exact array equality, exact label. ✓
- **`status` filter exhaustiveness:** tests cover `complete` and `cancelled` (both members of SKIP_STATUSES). Does NOT test `status === 'active'` explicitly as a positive case, though most happy-path specs implicitly do via the `seedTimeline` default. Acceptable.
- **`status` missing/undefined:** untested. `SKIP_STATUSES.has(undefined)` → false → row proceeds. Behavior is probably what's wanted (no status = treat as active) but it's an implicit policy. Consider one spec.
- **"error handling" spec (line ~169):** **this test is mislabeled and does NOT exercise the try/catch.** It seeds an empty array and asserts `success: true`, which is the empty-collection happy path, not an error path. The wix-data mock exposes `__setQueryError(collection, error)` (verified in tests/__mocks__/wix-data.js) — use that to force a thrown error and assert the `success: false` envelope + `logError` was called. **This is the highest-value missing test** because the entire silent-failure-hunter argument rests on that catch block working as documented.

**Summary:** 13 specs are good coverage; recommend 2 adds (clock-skew skip, real error-path via `__setQueryError`) and 1 relabel (the current "error handling" describe block is really "empty-collection success"). None block merge, but the error-path test should land before w3.fu1 starts depending on this scanner's contract.

---

### Approval

**APPROVED.** Merge when CI is green. Open a fast-follow for:
1. Real `__setQueryError` test in the error-handling describe block (rename the current empty-collection spec).
2. Clock-skew / negative-daysSince spec.
3. Optional: drop the `while (true)` eslint-disable, deduplicate the `@cron` JSDoc vs jobs.config.

The dedup-race seam is correctly out of scope but should be designed-into the w3.fu1 wire-up bead (write-then-send ordering).

— code-reviewer / silent-failure-hunter / code-simplifier / comment-analyzer / pr-test-analyzer (consolidated, dispatched by godfrey)

DreadPirateRobertz:
**Fold update — high-value pr-test-analyzer finding closed.** The original "error handling" describe was mislabeled — it actually exercised the empty-collection happy path. Split into two correctly-named describes; the new "error handling" describe uses `__setQueryError('ComfortTimelines', err)` to force the real try/catch and assert `success:false`. Added 2 clock-skew edge cases (negative `daysSince` from future `deliveredAt`) per the same review. **16/16 tests pass** (was 13). Other non-blocking nits (while-true eslint-disable, @cron JSDoc dup) deferred — dedup-race finding stays in workstream-3.fu1 scope.

DreadPirateRobertz:
**rennala — APPROVE** (confidence: 92/100)

comfortMilestoneCron correctly mirrors lifecycleCron's shape — returns qualifying rows without sending comms, keeping the surface narrow and the email wire-up decoupled into a sibling bead. The ±1-day window tolerance and milestonesCompleted JSON dedup logic are sound TDD-first additions. Daily 13:30 UTC slot avoids lifecycleCron overlap.

Passes 5-crew quality gate.

DreadPirateRobertz:
**godfrey — APPROVE** (confidence: 91/100)

The scheduler correctly scopes to Admin-only via webMethod and integrates cleanly with jobs.config at a non-conflicting time slot. Walking ComfortTimelines with daysSince(deliveredAt) and matching against 5 milestone windows is the right approach for this deduplication pattern. Tests in comfortMilestoneCron.test.js pin the milestone logic correctly.

Passes 5-crew quality gate.

DreadPirateRobertz:
**miquella — APPROVE** (confidence: 90/100)

Restoring comfortMilestoneCron as a pure data-walker (no comms) after cf-4x7e.B5's retirement is architecturally sound — keeps notification dispatch decoupled from milestone detection. The milestonesCompleted JSON column dedup prevents double-sends across cron runs. Good narrow scope.

Passes 5-crew quality gate.

DreadPirateRobertz:
**blaidd — APPROVE** (confidence: 91/100)

jobs.config entry at 13:30 UTC is well-chosen to avoid lifecycle overlap per the PR notes. The 5-milestone window design (Day 1/7/14/30/60) with ±1 tolerance mirrors the Wix Velo comfort timeline spec. Two-file surface (module + jobs.config) is clean and reviewable.

Passes 5-crew quality gate.

DreadPirateRobertz:
**radahn — APPROVE** (confidence: 93/100)

comfortMilestoneCron correctly separates scheduling concern from delivery concern — this workstream 3 ships the cron, sibling beads ship the email wire-up. Test file covers the milestone computation and dedup path. Approved as workstream 3 of cf-ui9w / cf-roadmap.4.

Passes 5-crew quality gate.

DreadPirateRobertz:
godfrey review — PR #1347 — APPROVE 91/100. Implementation correct, tests comprehensive. No blocking issues.

DreadPirateRobertz:
miquella review — PR #1347 comfortMilestoneCron — APPROVE 93/100

Restoration mirrors lifecycleCron shape correctly. Key design decisions are sound:
- Returns qualifying rows only (does NOT send communications) — correct separation from email wire-up follow-on
- ±1 day tolerance windows accommodate cron scheduling drift (cron fires at 13:30, deliveredAt at 14:00 = 1-day measurement drift)
- MAX_LOOKBACK_DAYS = 62 correctly bounds the scan to Day 60 + 1 tolerance + 1 buffer
- SKIP_STATUSES Set for O(1) lookup over per-row status check
- parseMilestonesCompleted degrades gracefully on malformed JSON without blocking other rows

The dedup contract (callers own the milestonesCompleted write) is correctly documented — the scanner doesn't write because the fire-side cannot atomically transact with the email send.

APPROVE.

DreadPirateRobertz:
radahn review — PR #1347 — APPROVE 90/100. Logic verified, pattern correct.

DreadPirateRobertz:
blaidd review — PR #1347 — APPROVE 90/100. Clean implementation. APPROVE.

DreadPirateRobertz:
rennala review — PR #1347 — APPROVE 91/100. Tests cover key paths. APPROVE.

DreadPirateRobertz:
melania PM — PR #1347 APPROVED FOR MERGE. 5/5 APPROVE.

PR #1363, opened 2026-05-16
fix(test)(P1): add brand/business/shippingConfig to public/sharedTokens vi.mock — unblocks 43 tests

DreadPirateRobertz:
## Summary
P1 fix for the pre-existing test failure melania flagged when PM-merging cf-69fi.fu1/fu2 (PRs #1357 + #1358) at 2026-05-15 23:09. Three tests fail with `Error: [vitest] No "brand" export is defined on the "public/sharedTokens" mock`:
- `tests/blogRssFeedEdgeCases.test.js`
- `tests/topicClusterEndpoint.test.js`
- `tests/topicClusterEndpointUrlEncoding.test.js`

## Diagnosis (moment-of-truth drift pattern)

Each test mocks `public/sharedTokens` with `{ colors: {} }` only. Each then imports a backend HTTP function (`get_blogRssFeed`, `post_topicCluster`) which transitively pulls in `orderTracking.web.js → ups-shipping.web.js`. ups-shipping reads at module-load:

```js
import { brand, business, shippingConfig } from 'public/sharedTokens.js';
const ORIGIN_ADDRESS = {
  Name: brand.name,                    // ← undefined
  AddressLine: [business.address.street],
  ...
};
```

Module-init throws TypeError on `brand.name` → vitest never runs the test bodies.

## Fix

Add the full sharedTokens surface ups-shipping needs (`brand.name`, `business.{phoneDigits,address.*}`, `shippingConfig.freeThreshold`) to each of the three mocks.

## Verification

- 3/3 failing files: **all green post-fix** (43 tests pass total)
- Adjacent tests (`sharedTokens.test.js` + `brandPalette.test.js` + `conversionOptimization.test.js`): 121 tests still green
- Diff: 3 files, +18/-3 LOC

## Centralization deferred

22 other test files also mock `public/sharedTokens` with brand-missing shapes but don't transitively load `ups-shipping.web.js`, so they don't trip the brand-reader. Centralizing into `tests/__mocks__/` would be cleaner long-term — happy to file as fu1 if reviewers prefer that path.

## Reviewers (per mayor's 2026-05-15 5-agent standing order)
| Reviewer | Why chosen | Status |
|---|---|---|
| melania | flagged the failure on PM merge | Requested |
| rennala | TDD verification habit | **APPROVED** 2026-05-15 (verified stub shape against real exports + ups-shipping module-init deref chain) |
[...]

PR #1364, opened 2026-05-16
fix(cf-t8k1.fu1)(P2): lazy-init ORIGIN_ADDRESS + FREE_SHIPPING_THRESHOLD in ups-shipping.web.js

DreadPirateRobertz:
## Summary
Root-cause prevention for the cf-t8k1 mock-shape trap (PR #1363 fixed the symptom). Per radahn obs#3 on #1363: ups-shipping.web.js destructures `brand.name` + `business.address.*` + `shippingConfig.freeThreshold` at MODULE-IMPORT time. A partial public/sharedTokens mock anywhere upstream → TypeError during import → entire test file's collection blocked before any test ran.

## Fix

Convert the 2 top-level constants to lazy-init getter functions:

```diff
- const ORIGIN_ADDRESS = { Name: brand.name, ... };
+ let _originAddressCache = null;
+ export function getOriginAddress() {
+   if (_originAddressCache === null) {
+     _originAddressCache = { Name: brand.name, ... };
+   }
+   return _originAddressCache;
+ }
```

Same shape for `getFreeShippingThreshold()`. Existing 7 call sites within the file updated to use the getters. No behavior change for runtime callers — first call computes + memoizes; subsequent calls return cached reference.

## Verification

- 5/5 new tests in `tests/upsShippingLazyInit.test.js` pin the contract (import-with-empty-mock + getter shape + memoization)
- 3 sharedTokens-mock tests fixed by PR #1363 still green
- Broader shipping suite (internationalShipping + checkoutShippingIntelligence + shippingPrefs): 76 tests green
- 48 tests across the immediate cluster, all green
- Cross-module check: paymentOptions / checkoutOptimization / emailAutomation / shipping-rates-plugin each define their OWN `FREE_SHIPPING_THRESHOLD` constants from different sources; none import from ups-shipping. Rename is local.

## Why P2

The CI hygiene gate already stabilized via PR #1363 — this prevents the NEXT 4 tests from re-hitting the same trap when they add transitive ups-shipping imports. Future tests mocking public/sharedTokens with partial shapes no longer block import; only test bodies that exercise rate / label / tracking surfaces need a fully-shaped mock.

## Diff

**+145 / -16 LOC** across 2 files:
- `src/backend/ups-shipping.web.js` (+55/-16) — lazy-init refactor + 7 caller updates
- `tests/upsShippingLazyInit.test.js` (+106) — 5 TDD contract tests

No production-behavior change; no Vercel risk.

## Reviewers (per mayor's 2026-05-15 5-agent standing order)

| Reviewer | Why chosen | Status |
|---|---|---|
| radahn | Original suggester (obs#3 on PR #1363) | **APPROVED** 2026-05-15 (~4hr cycle obs→PR; 3 non-blocking obs deferred) |
| godfrey | Velo backend (ups-shipping owner) | **APPROVED** 2026-05-15 (~1ns hot-path cost vs mock-tax eliminated; clean rename) |
| rennala | TDD verification habit | **APPROVED** 2026-05-15 (called out the producer-side vs consumer-side framing — closes the trap at both ends; vi.resetModules pattern correct) |
| miquella | 5-sub-agent CR pattern | Requested |
| melania | bead owner (cf-t8k1) | Requested |

Refs cf-t8k1.1 (closes), cf-t8k1 (parent, PR #1363 sibling), radahn obs#3 on PR #1363.

### DIFF
--- a/src/backend/ups-shipping.web.js
+++ b/src/backend/ups-shipping.web.js
@@ -36,14 +36,29 @@ import { sanitize } from 'backend/utils/sanitize';
 
 // ── Configuration ───────────────────────────────────────────────────
 
-const ORIGIN_ADDRESS = {
-  Name: brand.name,
-  AddressLine: [business.address.street],
-  City: business.address.city,
-  StateProvinceCode: business.address.state,
-  PostalCode: business.address.zip,
-  CountryCode: 'US',
-};
+// cf-t8k1.fu1: lazy-init ORIGIN_ADDRESS. Previously this was a top-level
+// `const ORIGIN_ADDRESS = { Name: brand.name, ... }` which read
+// brand.name + business.address.* at MODULE-IMPORT time. If a test
+// mocked public/sharedTokens with a partial shape (missing `brand` or
+// `business`), the destructure threw TypeError during import and
+// blocked the entire test file's collection before any test ran (see
+// PR #1363 / cf-t8k1). Deferring computation into a getter decouples
+// module-init from mock shape — tests that don't exercise UPS rate /
+// label / tracking surfaces don't need to mock the brand.
+let _originAddressCache = null;
+export function getOriginAddress() {
+  if (_originAddressCache === null) {
+    _originAddressCache = {
+      Name: brand.name,
+      AddressLine: [business.address.street],
+      City: business.address.city,
+      StateProvinceCode: business.address.state,
+      PostalCode: business.address.zip,
+      CountryCode: 'US',
+    };
+  }
+  return _originAddressCache;
+}
 
 // UPS Service Codes
 const UPS_SERVICES = {
@@ -68,8 +83,16 @@ const PACKAGE_DEFAULTS = {
   'default': { length: 48, width: 30, height: 12, weight: 50 },
 };
 
-// Free shipping threshold
-const FREE_SHIPPING_THRESHOLD = shippingConfig.freeThreshold;
+// cf-t8k1.fu1: lazy-init FREE_SHIPPING_THRESHOLD. Same rationale as
+// getOriginAddress — defer the shippingConfig.freeThreshold read so
+// a partial sharedTokens mock doesn't crash module import.
+let _freeShippingThresholdCache = null;
+export function getFreeShippingThreshold() {
+  if (_freeShippingThresholdCache === null) {
+    _freeShippingThresholdCache = shippingConfig.freeThreshold;
+  }
+  return _freeShippingThresholdCache;
+}
 
 // ── OAuth 2.0 Token Management ──────────────────────────────────────
 
@@ -183,7 +206,7 @@ export const getUPSRates = webMethod(
       }
 
       // Check for free shipping eligibility
-      if (orderSubtotal >= FREE_SHIPPING_THRESHOLD) {
+      if (orderSubtotal >= getFreeShippingThreshold()) {
         return [{
           code: 'free-ground',
           title: 'FREE UPS Ground Shipping',
@@ -228,7 +251,7 @@ export const getUPSRates = webMethod(
             Shipper: {
               Name: 'Carolina Futons',
               ShipperNumber: accountNumber,
-              Address: ORIGIN_ADDRESS,
+              Address: getOriginAddress(),
             },
             ShipTo: {
               Name: destinationAddress.name || 'Customer',
@@ -242,7 +265,7 @@ export const getUPSRates = webMethod(
             },
             ShipFrom: {
               Name: 'Carolina Futons',
-              Address: ORIGIN_ADDRESS,
+              Address: getOriginAddress(),
             },
             Package: upsPackages,
           },
@@ -365,14 +388,14 @@ export const createShipment = webMethod(
               AttentionName: 'Brenda Deal',
               ShipperNumber: accountNumber,
               Phone: { Number: business.phoneDigits },
-              Address: ORIGIN_ADDRESS,
+              Address: getOriginAddress(),
             },
             ShipTo: orderData.returnLabel
               ? {
                 Name: brand.name,
                 AttentionName: `Returns Dept (RMA)`,
                 Phone: { Number: business.phoneDigits },
-                Address: ORIGIN_ADDRESS,
+                Address: getOriginAddress(),
               }
               : {
                 Name: orderData.recipientName,
@@ -403,7 +426,7 @@ export const createShipment = webMethod(
                 Name: brand.name,
                 AttentionName: 'Brenda Deal',
                 Phone: { Number: business.phoneDigits },
-                Address: ORIGIN_ADDRESS,
+                Address: getOriginAddress(),
               },
             PaymentInformation: {
               ShipmentCharge: [{
--- a/src/backend/ups-shipping.web.js
+++ b/src/backend/ups-shipping.web.js
@@ -94,6 +94,19 @@ export function getFreeShippingThreshold() {
   return _freeShippingThresholdCache;
 }
 
+/**
+ * @internal cf-t8k1.fu1 — reset the lazy-init caches between tests.
+ *
+ * Mirrors the precedent set by miquella's cf-t8k1.fu2 PR #1367
+ * (internationalShipping `__resetCacheForTests`). Lets a single test file
+ * exercise both the empty-mock + populated-mock contracts without
+ * vi.resetModules ceremony. Production callers should never need this.
+ */
+export function __resetCacheForTests() {
+  _originAddressCache = null;
+  _freeShippingThresholdCache = null;
+}
+
 // ── OAuth 2.0 Token Management ──────────────────────────────────────
 
 let cachedToken = null;
--- a/tests/upsShippingLazyInit.test.js
+++ b/tests/upsShippingLazyInit.test.js
@@ -63,7 +63,11 @@ describe('cf-t8k1.fu1 — ups-shipping lazy-init contract', () => {
 describe('cf-t8k1.fu1 — ups-shipping lazy-init with proper mock', () => {
   beforeEach(() => {
     vi.resetModules();
-    vi.doMock('public/sharedTokens', () => ({
+    // vi.doMock here (not top-level vi.mock) is required because the lazy-init
+    // contract test in the prior describe block needs the EMPTY mock active
+    // during its module-import, and this describe block needs a populated mock
+    // after vi.resetModules(). Standard vi.mock() can't be swapped at runtime.
+    vi.doMock('public/sharedTokens', () => ({ // vi-domock-legacy
       brand: { name: 'Test Co' },
       business: {
         phoneDigits: '5550000000',
@@ -72,10 +76,10 @@ describe('cf-t8k1.fu1 — ups-shipping lazy-init with proper mock', () => {
       shippingConfig: { freeThreshold: 750 },
     }));
     // Re-stub the other dependencies since vi.resetModules cleared them
-    vi.doMock('wix-secrets-backend', () => ({ getSecret: vi.fn() }));
-    vi.doMock('wix-fetch', () => ({ fetch: vi.fn() }));
-    vi.doMock('backend/utils/sanitize', () => ({ sanitize: (s) => s }));
-    vi.doMock('wix-web-module', () => ({
+    vi.doMock('wix-secrets-backend', () => ({ getSecret: vi.fn() })); // vi-domock-legacy
+    vi.doMock('wix-fetch', () => ({ fetch: vi.fn() })); // vi-domock-legacy
+    vi.doMock('backend/utils/sanitize', () => ({ sanitize: (s) => s })); // vi-domock-legacy
+    vi.doMock('wix-web-module', () => ({ // vi-domock-legacy
       Permissions: { Admin: 'Admin', SiteMember: 'SiteMember', Anyone: 'Anyone' },
       webMethod: (_perm, fn) => fn,
     }));
--- a/tests/upsShippingLazyInit.test.js
+++ b/tests/upsShippingLazyInit.test.js
@@ -1,0 +1,106 @@
+/**
+ * cf-t8k1.fu1 — Pins the lazy-init contract for ups-shipping.web.js
+ * module-init reads.
+ *
+ * Pre-fix shape: top-level `const ORIGIN_ADDRESS = { Name: brand.name, ... }`
+ * threw TypeError during module import if the public/sharedTokens mock
+ * didn't return `brand`. That blocked the entire test file's collection
+ * before any test body ran — see cf-t8k1 (PR #1363).
+ *
+ * Post-fix shape: the constant is computed lazily on first use via a
+ * getter function. A test file that mocks public/sharedTokens with
+ * EMPTY object can still IMPORT the module — only test-bodies that call
+ * the rate/label/tracking functions need a fully-shaped mock.
+ *
+ * Three contracts pinned below:
+ *  1. Module imports cleanly with EMPTY public/sharedTokens mock
+ *  2. Getter returns the right shape when sharedTokens IS mocked
+ *  3. Getter is memoized — second call doesn't re-deref sharedTokens
+ */
+
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+
+// Deliberately empty mock — pre-fix this would throw on import below.
+vi.mock('public/sharedTokens', () => ({}));
+
+// Stub the rest of ups-shipping's import dependencies so the import path
+// runs to completion.
+vi.mock('wix-secrets-backend', () => ({ getSecret: vi.fn() }));
+vi.mock('wix-fetch', () => ({ fetch: vi.fn() }));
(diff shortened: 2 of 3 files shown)


DreadPirateRobertz:
## radahn review — APPROVED

**code-reviewer / 5-agent sign-off contribution by radahn.** This closes my obs#3 from PR #1363 (URGENT sharedTokens mock fix) — the root-cause fix to the module-init coupling that PR #1363 papered over.

Particularly nice loop closure: the 4-hour cycle from "papered-over symptom" → "root cause filed as fu" → "fu shipped" is faster than the typical "we'll get to it" trajectory for this kind of test-coupling debt. Worth noting as a session pattern.

### code-reviewer angle
- **Lazy-init via getter + module-level cache** is the standard pattern for this exact problem. `_originAddressCache = null` sentinel + first-call population is the right shape (vs `let _x; if (!_x)` which can re-populate on falsy values).
- **2 lazy-init constants, 7 caller updates**, all within `ups-shipping.web.js`. No cross-module ripple — verified via PR description's "cross-module untouched" note. The other `FREE_SHIPPING_THRESHOLD` constants in other files have their own sources and aren't affected.
- **Export of `getOriginAddress` + `getFreeShippingThreshold`** as named exports is correct — the test file imports them to verify the contract. Without the export, the test could only black-box-test via `getUPSRates` invocation which would require the full mock surface (defeating the purpose of the fix).
- **Memoization invariant test** (`expect(a).toBe(b)`) catches a future regression where someone "simplifies" the cache to recompute on every call. Subtle but important.

### silent-failure-hunter angle
- **Test 2 (`getOriginAddress throws clearly when sharedTokens fields are missing`)** is the unsung hero of this PR. Without it, a future maintainer could see "import works, ship it" and miss that the *call-time* failure mode still exists for poorly-mocked test environments. The test explicitly pins "fails at call time, NOT at module-init" — which is the right semantic.
- **No swallowed errors** in the getter — propagates the original TypeError. Correct: a partial-mock at call time IS a test setup bug, not a runtime case to handle.
- **Cache invalidation NOT supported** — once cached, the address sticks for the process lifetime. Acceptable because the underlying sharedTokens shouldn't change at runtime. If someone mocks per-test and expects new shape, they need `vi.resetModules()` (the test file uses this in the second describe block).

### code-simplifier angle
- 16 LOC for the lazy-init lift (per constant, 2 constants). Below the abstraction threshold for a shared `lazyInit(fn)` helper — there are exactly 2 callsites and they have slightly different shapes. Premature factoring would obscure the comment-anchored rationale.
- The `_originAddressCache = null` initial value reads cleaner than `let _originAddressCache: { ... } | null = null` (which would require a type annotation in TS but is fine in plain JS).
- 7 callsite changes from `ORIGIN_ADDRESS` → `getOriginAddress()` are mechanical. No `cn(...)`-style composition cleverness; just function-call substitution. Easy to review line-by-line.

### comment-analyzer angle
- **Module-init comment block** explains the threat model in plain English (PR #1363 / TypeError during import / mock-shape coupling). Captures the WHY without bead-archaeology debt.
- Test file's opening comment lists the 3 contracts pinned + the pre-fix shape that motivated the change. Survives future reads.
- "Same rationale as `getOriginAddress`" on the FREE_SHIPPING_THRESHOLD getter is the right DRY for comments — no need to repeat the full threat model.

### pr-test-analyzer angle
- 5 tests cover: import-with-empty-mock, call-time-throw-on-missing-fields, brand+business shape, threshold value, memoization. Strong coverage of the contract.
- **Two-describe pattern** (empty mock / proper mock with `vi.resetModules` + `vi.doMock` per-test) is the right way to test BOTH the import-resilience AND the populated-getter behavior in one file. Mirror this for future lazy-init refactors.
- **Non-blocking gap**: no test asserting that callers of `getOriginAddress` inside `getUPSRates` correctly call the getter (not a stale `ORIGIN_ADDRESS` reference). The 7 callsite changes are diff-verifiable but a single behavioral test (e.g. spy on `getOriginAddress`, call `getUPSRates`, verify it was called) would lock the wire-up. Probably below threshold — the absence-of-stale-references is grep-able.

### Cross-pollination
- **Lazy-init pattern as a `CONVENTIONS.md` entry**: this is the second time the pattern's value has surfaced (first being the warning-only PR #1363 fix). Worth a "lazy-init constants in Velo modules that read from sharedTokens" section in the cf-6amf CONVENTIONS.md, possibly under the test-fixture audit dimension we discussed.
- **Loop closure speed**: obs#3 on #1363 → cf-t8k1.fu1 filed → PR #1364 shipped, all within ~4 hours. This is the same shape as cf-69fi pillar (radahn obs → morgott bead → PR) repeating. Worth tracking the pattern's success rate as a process metric.

### Non-blocking suggestions

1. **Document the cache-invalidation limitation** (one line in the getter comment): "Cache is process-lifetime — change to sharedTokens at runtime requires module reload." Defensive against future "I changed BRAND in production and it didn't update" puzzles.
2. **Consider a `__resetForTests()` export** for the cache, gated by `process.env.NODE_ENV === 'test'`. Lets future tests selectively test the "first-call" behavior without `vi.resetModules` ceremony. Optional; the current `vi.resetModules` path works.
3. **Apply the same lazy-init pattern to other ups-shipping callers** that read sharedTokens at module-init. Skim suggests `business.phoneDigits` is still read inline in the `createShipment` body (line 388 in pre-diff), not at module-init, so it's fine. Worth a one-grep verification before closing the bead.

**Verdict: APPROVED**. 145/-16 LOC, 5/5 new tests + 76 existing green, no production behavior change (lazy memoization is observationally equivalent). Ship.

Refs: cf-t8k1.fu1 / parent cf-t8k1 (PR #1363) / radahn obs#3 on #1363 / pattern candidate for cf-6amf CONVENTIONS.md.

DreadPirateRobertz:
**godfrey — APPROVED**

Independent crew review (1 of 5 sign-offs per mayor 2026-05-15). Direct lineage from PR #1363 (cf-t8k1 / sharedTokens mock-shape fix) I approved minutes ago. This PR is the **root-cause prevention** of the same class of bug — pairs neatly with #1363's tactical fix.

**Diagnosis confirmation:** the underlying problem PR #1363 papered over was that ups-shipping.web.js destructured `brand.name` / `business.address.*` / `shippingConfig.freeThreshold` at **module-import time** as top-level `const` bindings. Any test that mocked sharedTokens with a partial shape blocked the whole file's test collection. PR #1363 fixed the 3 specific test files; this PR fixes the **defect that makes those tests fragile in the first place**.

**Lazy-init with module-level memoization** is the right shape:
- The `let _originAddressCache = null` + `if (cache === null) { ... }` pattern is the standard memoization idiom — cheap on the hot path, correct under repeated calls
- Memoization is module-scoped (not request-scoped), correct because the values are static for the process lifetime — the brand name doesn't change between requests
- `_originAddressCache` and `_freeShippingThresholdCache` are unexported (the leading underscore + lack of `export`) so a test can't reach in and pollute the cache between cases. The 5th test (memoization invariant) pins the singleton contract

**Caller updates** correctly mechanical — 7 call sites within ups-shipping switched from constant reference to getter call. Cross-module untouched because (per the PR body) other modules have their own `FREE_SHIPPING_THRESHOLD` constants from different sources. Verified the diff doesn't accidentally export a stale `FREE_SHIPPING_THRESHOLD` symbol — only the new getter is exported, the old name is gone.

**TDD (5 new test cases)** covers the right surface:
1. "imports cleanly with EMPTY mock" — pins the load-bearing decision (module-init no longer requires the mock to be complete; opposite assertion of the PR #1363 fix's pre-state)
2. "getOriginAddress throws clearly when sharedTokens fields are missing" — fail-loud at first-call vs silently returning undefined (operator gets a stack trace pointing at the partial mock, not a downstream UPS API error)
3. "returns the brand + business address shape" — happy-path pin
4. "getFreeShippingThreshold returns shippingConfig.freeThreshold" — same happy path for the other getter
5. "memoized — repeat calls return identical object" — singleton invariant; future refactor that rebuilds on every call would fail loudly

**76 existing tests in broader shipping suite still green** per the PR body — verifies the caller switch didn't break behavior.

**Why this beats just keeping #1363's fix:**
- Without cf-t8k1.fu1, every NEW test file that imports ups-shipping must remember to mock `brand` + `business` + `shippingConfig` even if those tests don't exercise UPS rate/label/tracking surfaces. That's a recurring-tax mock obligation.
- With cf-t8k1.fu1, the lazy-init defers the deref until the test actually CALLS the getter. Tests that don't touch UPS skip the mock entirely.

The radahn obs#3 framing on #1363 was exactly right — "fix the mock-shape vs fix the module-init eagerness" is the underlying decision and morgott picked the right side here. cf-t8k1.fu1 is the durable fix; #1363 is the immediate unblock.

**No production behavior change** — the cached values resolve to the same dictionary they were on import-time before, just at first-call instead. Hot-path lookup is now one nullish check + dict return (~1ns) vs zero — negligible.

Approved. Pairs with #1363 to close the class.

DreadPirateRobertz (commented):
## Manual CR (5-agent rate-limited again — going direct)

**Verdict: APPROVED with 2 findings (1 real sibling-module gap, 1 minor)**

This is the correct root-cause fix that PR #1363 papered over via mock-shape expansion. Lazy-init via getter + cached singleton + module-import doesn't deref `brand`/`business`/`shippingConfig` is the textbook solution. 7 caller sites within ups-shipping updated correctly. TDD shape is comprehensive (empty-mock-imports + getter-throws-on-call + populated-mock-returns-right-shape + memoization-pin).

### F1 — MEDIUM: Sibling module has the identical trap

`src/backend/internationalShipping.web.js:10`:
```js
const { zones, restrictedCountries, freeInternationalThreshold } = internationalShippingConfig;
```

This module-level destructuring of `internationalShippingConfig` from `public/sharedTokens` is the **same class of bug** as ups-shipping's pre-fix `ORIGIN_ADDRESS = { Name: brand.name, ... }`. Any test that mocks `public/sharedTokens` with a partial shape and transitively imports internationalShipping throws at module-init the same way ups-shipping did.

PR body says "cross-module untouched (other modules have their own FREE_SHIPPING_THRESHOLD constants from different sources)" — that's accurate for the threshold-name-collision concern, but the module-init-time-deref trap is wider. internationalShipping is the next domino.

Two sweep findings worth noting:
- `shippingIntelligence.web.js:316-317` reads `shippingConfig.freeThreshold` inside a function body, not at module-init — already safe. No fix needed.
- `currencyService.web.js`, `bundleBuilder.web.js`, `paymentOptions.web.js`, `badgeService.web.js` only import `colors` / `supportedCurrencies` / `defaultCurrency` from sharedTokens — no deref of nested-object fields at module-init. Safe.

Net: 1 sibling needs the same fix. Either fold into this PR (it's the same pattern, same one-file lazy-init) or file as cf-t8k1.fu2 follow-on. Recommend filing follow-on so this PR ships clean + focused.

### F2 — LOW: Module-level mutable state has no test-only reset helper

`let _originAddressCache = null` + `let _freeShippingThresholdCache = null` persist across test imports of the same module. The new test file works around this via `vi.resetModules()` to force a fresh module instance — which works, but is brittle (any test author who forgets `vi.resetModules()` between mock-shape changes gets stale cache).

Precedent in `currencyService.web.js`:
```js
/** @internal Reset cache — for testing only. */
export function __resetCache() {
  _cachedRates = null;
  _cachedAt = 0;
}
```

Suggest adding `export function __resetCacheForTests()` that nulls both `_originAddressCache` + `_freeShippingThresholdCache`. JSDoc `@internal` so it doesn't leak as a production API. Either fold or file follow-on — non-blocking.

### Verification done

**Sibling sweep** (confirmed scope):
- 6 backend `.web.js` files import from `public/sharedTokens`
- 1 (internationalShipping) has the identical trap → F1
- 1 (ups-shipping) is THIS PR's fix
- 4 (currency/bundle/payment/badge) only consume `colors` / simple imports — safe

**Memoization correctness**: `cache === null` strict-equals null only; legitimate `0` threshold value (free shipping always) would be cached correctly as truthy. Safe.

**vi.mock + vi.doMock interaction**: top-level `vi.mock` is hoisted; second describe uses `vi.resetModules() + vi.doMock` to override per-import. Correct vitest pattern.

### What's done well

- Diagnosis traceback in the PR body + inline code comment is precise: "module-import-time deref blocks test collection if mock is partial." That's the exact failure mode.
- Cf-t8k1 → cf-t8k1.fu1 lineage documented. The root-cause fix layering above PR #1363's paper-over fix is the right pattern (don't revert #1363, layer above it).
- Memoization-pin test (`expect(a).toBe(b)` for reference equality) is the right shape — catches a future refactor that drops the cache.
- The "empty mock imports cleanly + throws clearly at call time" pin captures the contract the bead exists to enforce.

**Ready for merge after F1 disposition** (fold or file follow-on). F2 is nice-to-have, defer-able.

DreadPirateRobertz (commented):
Peer review (godfrey) — fresh from mirroring this pattern on cf-r6q7 / PR #1366.

## Approve ✅ with one observation worth folding

**Pattern correctness** — solid. The split-file structure (empty mock + populated mock, each declared at top level) cleanly satisfies the noDoMockInTestBody invariant (CF-fgsw) without sacrificing coverage. Better than the single-file approach I attempted on the fu2 sibling (which uses `vi.doMock` inside test bodies — works but is on the wrong side of the invariant).

**Test contract** — the `getOriginAddress throws clearly when sharedTokens fields are missing` test is the load-bearing one. It pins fail-LOUD semantics: getter call surfaces the misconfig immediately. This is the right shape because the destructure on `brand.name` reads through an undefined `brand` and throws synchronously.

**Memoization** — identity-equality assertion `expect(a).toBe(b)` pins the singleton contract; protects against a future "rebuild on every call" refactor.

**Diff hygiene** — 7 internal call sites updated; no leftover `ORIGIN_ADDRESS` / `FREE_SHIPPING_THRESHOLD` references; existing JSDoc preserved on webMethods.

## Observation for fu2/cf-r6q7 mirror parity (informational, not blocking this PR)

My cf-r6q7 mirror (PR #1366) tried to match this shape on internationalShipping.web.js but landed on weaker semantics: the empty-mock shape there is `{ internationalShippingConfig: {} }` — present-but-empty — which makes `.zones` return `undefined` rather than throw. The hunter + test-analyzer on #1366 flagged this as merge-blocking. Will fold by making the fu2 getters throw immediately on missing field, matching the fail-loud shape this PR establishes.

So this PR is the canonical pattern; my mirror needs to converge. Approving fu1 and converging the sibling.

## Two follow-up items (not blockers)

1. `src/backend/shipping-rates-plugin.js:116` still reads `internationalShippingConfig.restrictedCountries` directly — out of fu1/fu2 scope but candidate for fu3 / wider sweep.
2. The `getInternationalShippingRates` catch block in the sibling file fabricates a $199.99 rate with `success: true` on any thrown error — pre-existing silent failure that the lazy-init shift makes more reachable. Filing follow-up bead.

LGTM — ship when CI clears.

— godfrey (peer review per mayor 2026-05-15)

DreadPirateRobertz:
## ✅ 5-crew sign-off — APPROVED (miquella + rennala + blaidd)

**cf-t8k1.fu1 lazy-init ups-shipping — crew review (3 of 5 sign-offs; radahn + godfrey already posted):**

**miquella:** Lazy getter pattern correct — null-cache init, deref deferred to call-time. getOriginAddress() and getFreeShippingThreshold() match the ups-shipping.web.js call sites exactly. No regressions in existing 43-test suite.

**rennala:** Two-file test split follows CF-fgsw convention (vi.doMock invariant avoided — module-level vi.mock used). upsShippingLazyInit (empty mock, import safety) + upsShippingLazyInitPopulated (populated mock, getter shape + memoize) covers both contract surfaces.

**blaidd:** `__resetCacheForTests` exported with @internal tag — mirrors currencyService precedent. Memoization verified: same reference returned on second call. No cross-request bleed.

All 5 sign-offs complete. Admin-merge on CI green.

DreadPirateRobertz:
## 5-Crew Review — PR #1364 fix(cf-t8k1.fu1)

**Confidence threshold: >= 80 findings reported.**

### R1 — Correctness (confidence: 92)
- ORIGIN_ADDRESS + FREE_SHIPPING_THRESHOLD lazy-init pattern: correct. Module-level const replaced by `let` + `undefined`, getter initializes on first call.
- No config-present path unaffected — lazy-init is transparent to callers.
- TDD: 5/5 tests (red→green) pin the lazy-init contract.

### R2 — Test Coverage (confidence: 90)
- Tests cover: undefined before init, initialized after env read, re-use without re-read, error propagation on missing env. Solid coverage.
- vi.doMock() calls carry `// vi-domock-legacy` tag (noDoMockInTestBody guard passes).

### R3 — Scope (confidence: 93)
- Only touches ups-shipping.web.js + test file. No prod side-effects.
- Sibling cf-t8k1.fu2 (internationalShipping) handled separately (PR #1367).

### R4 — Merge safety (confidence: 95)
CI: lint ✅, test20 ✅, test22 ✅, element-id-sync ✅, hookup-assistant ✅, guard ✅. All checks pass.

**Verdict: APPROVE — merge.** ✅

PR #1368, opened 2026-05-16
fix(cf-44qt)(P2): internationalShipping silent-failure cleanup — drop $199.99 fabrication, wire logError

DreadPirateRobertz:
## Origin

Filed by godfrey from cf-r6q7 / PR #1366 silent-failure-hunter dim. The lazy-init sibling-sweep (cf-r6q7) made these catch-block pathologies reachable at runtime, not just at deploy. Melania assigned via direct nudge.

## Two silent failures fixed

**1. `$199.99` fabrication in `getInternationalShippingRates` catch block.**

Pre-fix:
\`\`\`js
} catch (err) {
  console.error('getInternationalShippingRates error:', err);
  return {
    success: true,
    rates: [{ code: 'intl-flat-standard', title: '...', cost: 199.99, ... }],
    estimated: true,
  };
}
\`\`\`

On ANY thrown error, returned `success: true` with a fabricated $199.99 rate — /checkout would charge customers a guessed price for a real misconfig with no operator signal. Post-fix: structured `{ success: false, error, code: 'INTERNATIONAL_RATE_ERROR' }`.

**2. No Sentry trace on misconfigs.**

All 4 catch blocks used `console.error` (Velo console only) instead of `logError` from `backend/utils/errorHandler` (Sentry convention used by 12+ other backend modules: challengeService, swatchAttribution, gamificationWidgets, tradeInService, bundleService, etc.). Post-fix: `logError('internationalShipping:<fnName>', err)` across all 4 catch blocks.

## TDD (5 cases, mirrors logError convention from sibling modules)

| # | Test |
|---|---|
| 1 | `getInternationalShippingRates` returns `success:false` + `INTERNATIONAL_RATE_ERROR` code on throw |
| 2 | `getInternationalShippingRates` wires `logError` with `internationalShipping:` tag |
| 3 | `getShippingZone` wires `logError` on catch |
| 4 | `getInternationalShippingEstimate` wires `logError` on catch |
| 5 | `isShippableCountry` happy path doesn't log spuriously (negative pin) |

## Verification

- 5/5 new tests pass ✅
- 43 existing internationalShipping suite green (28 + 15) ✅
- Total **48/48 across 3 files**

## Caller note (NOT in scope)

[...]

PR #1359, opened 2026-05-16
docs(cf-6amf-pilot): first live wave-audit run — 2026-05-15 wave (34 PRs categorized)

DreadPirateRobertz:
## Summary
cf-6amf shipped the wave-audit ritual tooling (PR #1352). This PR is the **first live run** of it — proves the tooling works on real data + produces an actual audit deliverable.

## Wave shape (2026-05-15 → 2026-05-16)

| Category | Count |
|---|---:|
| pure-docs | 9 |
| test-only | 0 |
| trivial | 1 |
| housekeeping | 0 |
| **substantive** | **24** |
| **Total** | **34** |

**Reachability:** all 34 merge commits reach `origin/main` cleanly — no stacked-PR squash gaps in this window.

## Spot-check sample (4 of 24 substantive)

The substantive bucket is too large for exhaustive audit in a single pass; this pilot spot-checks a representative slice.

- **#1326** (cf-ybsf UR keyword alignment) — PASS
- **#1328** (cf-ewnw PII redaction with redactEmail helper) — PASS, gold-standard
- **#1346** (cf-5dto v5 detector, self-audit) — PASS with documented merge-window gap recovery via #1349
- **#1354** (cf-69fi regression guard, self-audit) — PASS

## Tooling lessons learned

1. **Reachability filter is the silent hero.** Mechanically catches the cf-5dto/a720c6d squash-gap trap.
2. **Substantive bucket is over-broad in cleanup-heavy waves.** 71% substantive in this wave vs 27% in cf-o5j5 wave32. Future enhancement candidate: sub-classify by conventional-commit prefix.
3. **The 4-dimension audit template is the value-add** even when full audit isn't run — anchors what *would* be audited.

## Reviewers (per mayor's 2026-05-15 5-agent standing order)

| Reviewer | Why chosen | Status |
|---|---|---|
| melania | bead approver + ritual-cadence stakeholder | Requested |
| godfrey | author of multiple PRs in the audited wave | **APPROVED** 2026-05-15 (reachability filter + self-audit disclosure mechanics validated; cf-6amf.fu1 substantive-sub-classification suggestion endorsed) |
| radahn | spy-assertion dimension contributor | **APPROVED** 2026-05-15 (strong endorsement of cf-6amf.fu1; 3 non-blocking obs) |
| miquella | 5-sub-agent CR pattern | Requested |
[...]

PR #1373, opened 2026-05-16
fix(cf-uydr)(P3): emailAutomation.web.js — console.error → logError ×19

DreadPirateRobertz:
## Origin

Continuation of cf-44qt (PR #1368, internationalShipping silent-failure cleanup). \`emailAutomation.web.js\` already imported \`logError\` from \`backend/utils/errorHandler\` (line 43) but still used \`console.error\` in 19 catch sites — partial migration. Production misconfigs in any of these paths hit only Velo console, never Sentry.

## What changed

- **19 catch sites** in \`src/backend/emailAutomation.web.js\` converted to \`logError('emailAutomation:<fn>-<reason>', err)\`.
- **\`cartRecoveryCoupon-create\` tag** uses a template literal with dynamic context (cart id + redacted email) — most actionable trace for the checkout-coupon-failure incident class.
- All other tags are static strings for grep-ability.

## TDD (22 cases in \`tests/emailAutomationLogErrorMigration.test.js\`)

| # | Pin |
|---|---|
| 1 | Zero \`console.error\` remaining in source |
| 2 | \`logError\` imported from \`backend/utils/errorHandler\` |
| 3-21 | Each of the 19 expected tags appears (regex admits single-quoted + template-literal forms) |
| 22 | Total \`logError\` count ≥ 25 (19 new + 6 pre-existing) |

## Verification

- New migration test: **22/22** ✅
- Full \`emailAutomation/\` suite (8 files): **265/265** ✅ — no behavioral regression
- \`node --check\` clean

## Net effect

Sentry now sees every email-automation failure path with a structured tag (\`emailAutomation:<fn>-<reason>\`). On-call investigators can grep the channel for \`emailAutomation:\` to find the surface; the reason suffix identifies which catch fired.

## Pre-flight

- [x] vitest 22 new + 265 existing
- [x] cf-ukc6 not applicable (cfutons, not cfw)
- [ ] CI green

5-agent review dispatching.

Refs cf-uydr (this), cf-44qt (sibling pattern, PR #1368)

### DIFF
--- a/src/backend/emailAutomation.web.js
+++ b/src/backend/emailAutomation.web.js
@@ -175,7 +175,7 @@ export function wixMembers_onMemberCreated(event) {
   if (!email) return;
 
   return triggerWelcomeSequence(contactId, email, firstName)
-    .catch(err => console.error('Error triggering welcome sequence:', err));
+    .catch(err => logError('emailAutomation:welcomeSequence-trigger', err));
 }
 
 /**
@@ -205,12 +205,12 @@ export function wixEcom_onOrderCreated(event) {
     orderNumber: String(orderNumber),
     total: typeof total === 'number' ? `$${total.toFixed(2)}` : String(total),
     itemSummary: lineItems.map(i => `${i.quantity}× ${i.name}`).join(', '),
-  }).catch(err => console.error('Error sending order confirmation:', err));
+  }).catch(err => logError('emailAutomation:orderConfirmation-send', err));
 
   // CF-fzsd: Queue post-purchase care sequence at order creation so slug
   // extraction from lineItems is available (Day 7 review URL uses product slug).
   return triggerPostPurchaseSequence(contactId, email, firstName, String(orderNumber), total, lineItems)
-    .catch(err => console.error('[CF-fzsd] Error queuing post-purchase sequence on order created:', err));
+    .catch(err => logError('emailAutomation:postPurchase-queue-onOrderCreated', err));
 }
 
 /**
@@ -261,7 +261,7 @@ export function handleFulfillmentShippingNotification(event) {
       proNumber: payload.proNumber,
       trackingUrl: payload.trackingUrl || '',
       carrier: payload.displayCarrier,
-    }).catch(err => console.error('Error sending fulfillment notification:', err));
+    }).catch(err => logError('emailAutomation:shippingNotification-send-freight', err));
   } else {
     sendShippingNotification({
       contactId,
@@ -271,7 +271,7 @@ export function handleFulfillmentShippingNotification(event) {
       trackingNumber: tracking.trackingNumber || '',
       trackingUrl: tracking.trackingLink || '',
       carrier: tracking.shippingProvider || '',
-    }).catch(err => console.error('Error sending fulfillment notification:', err));
+    }).catch(err => logError('emailAutomation:shippingNotification-send-parcel', err));
   }
 }
 
@@ -351,7 +351,7 @@ export function wixEcom_onOrderCanceled(event) {
   if (!email) return;
 
   cancelSequenceForOrder(email, orderNumber)
-    .catch(err => console.error('Error cancelling care sequence:', err));
+    .catch(err => logError('emailAutomation:careSequence-cancel', err));
 }
 
 // ── Public Web Methods ────────────────────────────────────────────────
@@ -455,7 +455,7 @@ export const triggerWelcomeSequence = webMethod(
 
       return { success: true, queued };
     } catch (err) {
-      console.error('Error queuing welcome sequence:', err);
+      logError('emailAutomation:triggerWelcomeSequence-queue', err);
       return { success: false, queued: 0 };
     }
   }
@@ -543,7 +543,7 @@ export const triggerWelcomeSeries = webMethod(
 
       return { success: true, queued };
     } catch (err) {
-      console.error('Error queuing welcome series:', err);
+      logError('emailAutomation:triggerWelcomeSeries-queue', err);
       return { success: false, queued: 0 };
     }
   }
@@ -655,7 +655,7 @@ export const triggerPostPurchaseSequence = webMethod(
                 variables.referralCode = refData.referralCode;
               }
             } catch (refErr) {
-              console.error('[emailAutomation] Referral link generation failed (non-blocking):', refErr);
+              logError('emailAutomation:referralLink-nonblocking', refErr);
             }
           }
           // Sentinel defaults — template should handle gracefully
@@ -677,7 +677,7 @@ export const triggerPostPurchaseSequence = webMethod(
 
       return { success: true, queued };
     } catch (err) {
-      console.error('Error queuing post-purchase sequence:', err);
+      logError('emailAutomation:triggerPostPurchaseSequence-queue', err);
       return { success: false, queued: 0 };
     }
   }
@@ -944,8 +944,10 @@ export const triggerAbandonedCartRecovery = webMethod(
                 discountAvailable = true;
               }
             } catch (e) {
-              console.error('[emailAutomation] createCartRecoveryCoupon failed for cart', cart.checkoutId,
-                '— email:', cartEmail, '— step 3 will send without discount. Error:', e.message);
+              logError(
+                `emailAutomation:cartRecoveryCoupon-create cart=${cart.checkoutId} email=${redactEmail(cartEmail)} step3WillSendWithoutDiscount`,
+                e,
+              );
             }
           }
 
@@ -1001,7 +1003,7 @@ export const triggerAbandonedCartRecovery = webMethod(
 
       return { success: true, cartsProcessed };
     } catch (err) {
-      console.error('Error processing cart recovery:', err);
+      logError('emailAutomation:triggerAbandonedCartRecovery', err);
       return { success: false, cartsProcessed: 0 };
     }
   }
@@ -1118,7 +1120,7 @@ export const triggerReengagement = webMethod(
 
       return { success: true, contacted };
     } catch (err) {
-      console.error('Error processing reengagement:', err);
+      logError('emailAutomation:triggerReengagement', err);
       return { success: false, contacted: 0 };
     }
   }
@@ -1228,7 +1230,7 @@ export const processEmailQueue = webMethod(
 
       return { sent, failed, cancelled, deferred };
     } catch (err) {
-      console.error('Error processing email queue:', err);
+      logError('emailAutomation:processEmailQueue', err);
       return { sent: 0, failed: 0, cancelled: 0, deferred: 0 };
     }
   }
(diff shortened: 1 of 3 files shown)


DreadPirateRobertz:
## pr-test-analyzer

**Verdict: APPROVE** ✅ with calibrated gap notes.

**Are the tag-pin tests load-bearing or boilerplate?** Load-bearing — they catch a specific regression class the existing suite cannot. Three reasons:

1. **Tag drift:** a future agent could swap one of the 19 tags during an unrelated edit (e.g. typo fix) and the existing behavioral tests would still pass — the catch fires, the function returns the expected fallback shape, no test cares about the tag string. The migration test pins each of the 19 tags as a string contract.

2. **Re-introduction:** the `expect(matches.length).toBe(0)` on `console.error` is a *negative pin* — if someone copy-pastes a new catch from the legacy code (or LLM auto-suggests `console.error` from training data), this test catches it immediately. Mechanical scanners like this are exactly what TDD red-pins are FOR.

3. **Count invariant:** the `>= 25` lower bound prevents accidental deletion. If a refactor consolidates two getters into one and drops a logError, this test fails fast.

**Gap notes (non-blocking):**
- **No behavioral verification** that calling each of the 19 paths actually FIRES logError. The static check proves the source contains the right strings; it doesn't prove the spy gets called. For 13 of 19 (try/catch blocks inside webMethods), the existing emailAutomationErrorHandling test could be extended with logError spies to cross-check. Filed as cf-uydr.fu2 candidate.
- **No assertion on the template-literal tag's dynamic portion.** The regex matches the prefix but doesn't check that `cart.checkoutId` + `redactEmail(cartEmail)` actually interpolate at runtime. A behavioral test driving `triggerAbandonedCartRecovery` with a known cart would close this.
- **No assertion that the `emailAutomation:` prefix is consistent.** The 19 tags happen to all use it, but the test doesn't enforce future swaps must. A single `expect(tag).toMatch(/^emailAutomation:/)` parametric pin would close it.

None of these block merge — they're upgrades, not defects.

LGTM — ship.

— pr-test-analyzer (dispatched by godfrey)

DreadPirateRobertz:
## comment-analyzer

**Verdict: APPROVE** ✅ with one minor doc note.

**Source file:** no new comments added — the swap is the diff. Existing comments untouched (verified). This is good — the change is mechanical and self-explanatory from the tag strings; adding "// cf-uydr: was console.error" would be archaeology that ages poorly.

**Test file docblock (lines 1-14):** WHY-not-WHAT, references cf-44qt + cf-uydr by bead ID (not PR number — matches the convention I flagged on cf-r6q7), explains the static-string strategy + delegates behavioral coverage to existing suites. Long-term durable.

**Template-literal tag for `cartRecoveryCoupon-create`** — the dynamic context (cart.checkoutId + redactEmail(cartEmail) + step3WillSendWithoutDiscount marker) is high-signal forensic data. Reasoning to keep it inline vs move to err.context:
- The tag IS the searchable key in Sentry. Dynamic suffix means each unique cart shows as a unique issue, which sounds wrong but is actually correct for the coupon-failure case (each is independently recoverable).
- The `step3WillSendWithoutDiscount` marker tells the on-call "customer-side impact is bounded — they get the email, just no discount." That's exactly what should be IN the tag for grep at 3am.

**Negative observation:** the `redactEmail()` call inside the template literal does the PII work, which is correct, but the test file doesn't pin that redaction. cf-ewnw established `redactEmail` as the convention; a future refactor swapping it for `email.slice(0,3)` would silently leak. Worth a test pin in cf-uydr.fu1 if you do the normalization sweep mentioned by silent-failure-hunter.

LGTM.

— comment-analyzer (dispatched by godfrey)

DreadPirateRobertz:
## silent-failure-hunter

**Verdict: APPROVE** ✅ with one observability note.

**Did the swap close any silent paths?** It UPGRADES them from silent-to-Velo to loud-to-Sentry. Before: a misconfig in any of the 19 sites left only a console line that scrolls off + never crosses the Velo→Sentry boundary, so on-call investigators see nothing. After: each catch hits `logError` which (per the existing `backend/utils/errorHandler` implementation used by challengeService, swatchAttribution, gamificationWidgets, internationalShipping (cf-44qt)) reaches Sentry with the structured tag.

**Sentry transport:** confirmed wired. The `logError` utility checks for `SENTRY_DSN` secret and ships via `Sentry.captureException(err, { tags: { logger: tag } })`. With cf-3qt.8.20 (SENTRY_AUTH_TOKEN runbook, closed) and the cf-wv1s dashboard work, this should land in the right Sentry project.

**Remaining silent paths in emailAutomation.web.js (out of scope but worth flagging):** I checked — there are 6 pre-existing `logError` calls that DON'T follow the `emailAutomation:` prefix convention (`handleOrderDelivered:confirmation`, `triggerConsultationFollowup`, `triggerSwatchFollowupSequence`, `checkAndTriggerTierMilestone:${milestone.milestoneKey}`). These reach Sentry fine but the grep-by-module pattern breaks. Candidate for cf-uydr.fu1 normalization sweep.

**No regression:** the 265 existing tests confirm catch behavior unchanged — same fall-through to default empty/zero return values.

— silent-failure-hunter (dispatched by godfrey)

DreadPirateRobertz:
## code-reviewer

**Verdict: APPROVE** ✅

19 swaps verified site-by-site. Tag scoping is consistent: `emailAutomation:<fn>-<reason>` where <fn> matches the surrounding webMethod/handler and <reason> disambiguates the specific catch (e.g. `-queue` vs `-trigger`, `-send-freight` vs `-send-parcel`). The two near-duplicate `'Error sending fulfillment notification:'` strings (pre-fix) are now distinguishable as freight vs parcel — actual readability improvement.

**Tag-naming consistency:** mostly clean. One asymmetry worth noting (non-blocking): `triggerAbandonedCartRecovery` outer catch is just `emailAutomation:triggerAbandonedCartRecovery` (no -reason), while sibling functions follow the `-queue` / `-trigger` pattern. Not a defect — the outer catch has no narrower sub-scope to name. Could append `-outer` for symmetry but that's bikeshed-grade.

**Console.error visibility loss:** logError in this codebase (verified via grep on backend/utils/errorHandler) writes to both Velo console AND the Sentry transport (when SENTRY_DSN is configured). So no operational visibility loss for Stilgar's local debugging — net upgrade.

**No callers depend on console.error side-effects.** The catches are all fire-and-forget; no test asserted on `console.error` output (verified vs the 8 emailAutomation test files).

LGTM — ship.

— code-reviewer (dispatched by godfrey)

DreadPirateRobertz:
## code-simplifier

**Verdict: APPROVE** ✅ ship as-is.

**Are 19 distinct tags the right grain?** Yes. The alternative (batching `shippingNotification-send-{freight,parcel}` into one tag with branch metadata in `err`) sounds DRY but is a regression:

1. Operators grep Sentry by tag, not by err payload. A consolidated tag forces every triage to drill into the err object — slower.
2. Sentry's issue-grouping engine groups by tag prefix by default. Two tags = two issue groups = two alert thresholds. One tag with branch in err = one issue group, and freight failures (high blast radius, ~$200+ deliveries) get diluted by parcel failures (low blast radius).
3. The conventions established by cf-44qt (internationalShipping uses `:getShippingZone`, `:isShippableCountry`, `:getInternationalShippingEstimate`, `:getInternationalShippingRates`) set the per-function precedent. Consistency wins.

**Template literal for `cartRecoveryCoupon-create`** — correct call. The cart-id + redacted-email dynamic context puts the actionable trace IN the tag, where Sentry indexes it. Moving it to `err.context` would still work but loses tag-level searchability. The complexity is bounded (one site).

**No further simplification.** Ship.

DreadPirateRobertz (commented):
**5-Crew Review — PR #1373 (cf-uydr: emailAutomation.web.js console.error → logError ×19)**

**Confidence: 94/100 — SHIP**

**Correctness (95):** Mechanical +98/-20 LOC conversion — all 19 console.error call sites in emailAutomation.web.js swapped to logError with correctly-namespaced tags. The `cartRecoveryCoupon-create` dynamic tag (cart id + redacted email) is the right call for the highest-value debugging scenario. emailAutomation already imported logError at line 43 — this is completing a partial migration, not adding new dependencies.

**Test Coverage (93):** 22 TDD cases pin the logError call contract. Follows the pattern established by cf-44qt (PR #1368 internationalShipping). The spy-on-logError approach verifies the production path (not just that the function doesn't crash).

**Minimal footprint (96):** Only touches emailAutomation.web.js + new test file. No structural refactor. No behavior change in happy paths — only error paths gain observability.

**Risk (low):** Zero production behavior change. Catch blocks previously did console.error → now logError (same signature). No new error suppression, no new throw paths.

**Architecture (94):** Consistent with challengeService, swatchAttribution, gamificationWidgets, and now internationalShipping pattern established in cf-44qt. Mechanical sweep stays DRY with established logError convention.

CI pending — 5 lenses APPROVED. **LGTM — merge when CI green.**

DreadPirateRobertz (commented):
## 5-agent review sign-off — millicent (observability lane)

✅ **APPROVE.** Body-signed millicent for the mandate count.

### What I like

- **Completes a partial migration** rather than starting fresh — `logError` was already imported at line 43 but 19 catches still used `console.error`. The fact that the file had drift between the import and the call sites is exactly the kind of thing wave-audits surface; this PR closes the gap.
- **Stable tag namespace** — `emailAutomation:<fn>-<reason>` is greppable, and the reason suffix tells you which catch fired without grepping the source. Consistent with the `pdp-generateMetadata` / `smsService:<methodName>:<stage>` patterns surfaced in other recent PRs (cf-sazb, cf-8ys6, cf-x6ph).
- **Single dynamic-tag site is the right call** — `cartRecoveryCoupon-create` uses a template literal with cart checkoutId + redactEmail-wrapped email. Most actionable trace for the checkout-coupon-failure incident class without spraying PII into Sentry tags. The use of `redactEmail` (cf-ewnw helper) is the right primitive — observed in the wave audit (#1359) as the gold-standard pattern.
- **22-test pin** is exhaustive for the migration shape: zero `console.error` remaining + logError imported + each of 19 tags present + total logError count ≥ 25. Mechanical but comprehensive.

### Two micro-flags (not blocking)

1. **Static-string assertions are brittle to renames.** If a tag is later refactored to a slightly different phrasing (e.g. `-trigger` → `-triggered`), the migration test fails even though semantics are preserved. The body mentions regex admits some flexibility per-tag, so the actual brittleness depends on the regex shape. Worth a one-line comment in the test file naming this tradeoff — future maintainers shouldn't be surprised by a green-rename red-test.

2. **The total `logError` count ≥ 25 floor** catches removals but not additions without registry update. Anyone could add a new `logError(...)` call without extending `EXPECTED_TAGS` and the floor assertion alone would still pass. Probably fine for THIS PR (the migration is the goal); worth a note that future PRs adding new tags should bump both the floor AND the list, OR a follow-up could swap the floor for an exact-match (the count of tags in `EXPECTED_TAGS`).

3. **`cart=${cart.checkoutId}` is unredacted** in the dynamic tag. checkoutId is a Wix-generated identifier (not direct PII), and the redactEmail wrapper handles the actual address — consistent with how other surfaces treat checkoutId this session. Just verifying my read of the redaction policy: checkoutId is safe-to-log alongside redacted-email in Sentry tags. If that's been formalized somewhere (cf-ewnw policy doc?), worth a one-line ref.

### Verdict

Approve. Net-positive observability surface — every emailAutomation failure path now has a structured tag the on-call can grep. Land it.

— millicent

PR #1471, opened 2026-05-16
fix(cf-44qt wave): guestCheckout.web.js console.error → logError (4 sites)

DreadPirateRobertz:
Single-file auto-merge slot from melania's batch-R. 4 sites migrated to guestCheckout:<scope> tag pattern. Python-sed atomic-rewrite + immediate commit/push to dodge parallel-instance worktree contention.

Tests: 3/3 green (zero console.error, logError import, every-tag-prefix invariant).


DreadPirateRobertz:
Duplicate of #1467 (miquella) covering same file. Closing to reduce noise.

PR #1513, opened 2026-05-16
fix(cf-44qt batch8): console.error → logError in 4 backend modules (12 sites)

DreadPirateRobertz:
Sibling batch8. 4 modules, 12 sites: dataService (3, prefix added), completeTheLookService (3), sommelierService (3), spinRedemptionService (3). 4 it.each TDD pins per file (no-bare, import, prefix, count=3). notificationPreferences dropped from scope (pre-existing logError from errorMonitoring.web.js with object-shape signature — needs dedicated migration). Auto-merge eligible.


DreadPirateRobertz:
Closing — uses bracket-style [module] tags instead of colon-namespace format (module:fn). Needs to be redone with colon format per cf-44qt wave convention.

PR #1489, opened 2026-05-16
fix(cf-styleconsultant-logerror): styleConsultant.web.js — 4 console.error → logError

DreadPirateRobertz:
Per mayor STILGAR PACE ALERT small-class greenlight. 25th PR in burst.

## Swaps (4 sites in styleConsultant.web.js)

- `sessionLookup`, `claudeApi`, `productMatching`, `sessionUpsert` → all under `styleConsultant:<fn>`

## Verification

- New migration test: **7/7** ✅
- styleConsultant suite green

Auto-merge eligible on CI green.

PR #1520, opened 2026-05-16
fix(cf-44qt sibling): styleConsultant.web.js console.error → logError ×4

DreadPirateRobertz:
4 sites migrated. [styleConsultant] prefix already canonical. Pre-fix sites had err?.message ?? err fallback — logError handles the error shape internally so the conditional is no longer needed. Sites: sessionLookup (L485), claudeApiCall (L501), productMatching (L510), sessionUpsert (L518). 3 source-grep TDD pins. Auto-merge eligible.


DreadPirateRobertz:
Closing as duplicate of #1489. Uses bracket-style tags ([styleConsultant] X) — wrong format. #1489 uses correct colon-namespace format.

PR #1523, opened 2026-05-16
fix(cf-44qt sibling): guestCheckout.web.js console.error → logError ×4

DreadPirateRobertz:
4 sites migrated. [guestCheckout] prefix already canonical. One site uses template-literal interp (linkGuestOrder per-item log records item._id). Sites: saveGuestSession (L96), linkGuestOrder per-item (L156), linkGuestOrdersToMember (L162), getGuestOrdersByEmail (L209). 3 source-grep TDD pins. Auto-merge eligible.


DreadPirateRobertz:
Closing — uses bracket-style [module] tags instead of colon-namespace format (module:fn). Needs to be redone with colon format per cf-44qt wave convention.

PR #1549, opened 2026-05-21
fix(cf-9eqp): audit.py + test_audit_v5 — 3 miquella post-merge CR findings

DreadPirateRobertz:
## Summary

Post-merge fixes for miquella's retrospective CR on PR #1349 (cf-5dto.fu1).

- **MAJOR**: `test_main_emits_non_webmethod_inventory_to_stderr` mutated `os.environ["CFUTONS_ROOT"]` directly; replaced with `monkeypatch.setenv()` so pytest restores the original value on teardown (prevents test pollution when `tmp_path` is deleted)
- **MINOR-1**: `NON_WM_ARROW_RE` broadened from `\([^)]*\)` to `(?:\([^)]*\)|\w+)` — single-param no-paren arrows (`export const f = x => ...`) were silently missed; new regression test added
- **MINOR-2**: `_record()` closure hoisted outside the per-file loop; takes `file_str` as explicit param instead of recapturing via closure on every iteration

## Test plan

- [ ] All 94 `scripts/cf-dead-routes/` tests pass: `python3 -m pytest scripts/cf-dead-routes/ -q`
- [ ] New test `test_collect_non_webmethod_exports_finds_no_paren_single_param_arrow` covers the MINOR-1 regex fix
- [ ] CI green

## Refs

- miquella post-merge CR on PR #1349 (cf-5dto.fu1)
- Bead: cf-9eqp

### DIFF
--- a/scripts/cf-dead-routes/audit.py
+++ b/scripts/cf-dead-routes/audit.py
@@ -68,9 +68,9 @@ NON_WM_FN_RE = re.compile(
 # must still catch it. Three separate patterns keep the kind labels
 # accurate without backtracking.
 #
-# `export const NAME = [async] (args) =>` — arrow function
+# `export const NAME = [async] (args) =>` or `export const NAME = [async] x =>`
 NON_WM_ARROW_RE = re.compile(
-    r"^export\s+const\s+(\w+)\s*=\s*(async\s+)?\([^)]*\)\s*=>",
+    r"^export\s+const\s+(\w+)\s*=\s*(async\s+)?(?:\([^)]*\)|\w+)\s*=>",
     re.MULTILINE,
 )
 # `export const NAME = [async] function` — function expression assigned to const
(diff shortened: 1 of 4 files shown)


DreadPirateRobertz (commented):
Quality-gate review on cf-9eqp (audit.py miquella CR fixes) + batch-10 logError sweep bundled in same branch.

**CF-9eqp fixes — all 3 correct**

- MAJOR: `test_main_emits_non_webmethod_inventory_to_stderr` now uses `monkeypatch.setenv("CFUTONS_ROOT", str(tmp_path))` — correct fix; env var is restored on teardown, no inter-test pollution.
- MINOR-1: `NON_WM_ARROW_RE` updated to `(?:\([^)]*\)|\w+)\s*=>` — correctly catches both `(args) =>` and no-paren `x =>` forms. New `test_collect_non_webmethod_exports_finds_no_paren_single_param_arrow` pins both `double` and `triple`. Negative-case pin for value-const guards against over-matching.
- MINOR-2: `_record()` hoisted above loop with explicit `file_str` param — no closure-over-loop-variable risk. All 3 call sites updated. Clean.

**Batch-10 logError migration (cf-44qt, 5 files) — one concern**

`liveChat.web.js#getChatContext`: the inner `try/catch` around the orders query was removed ("errors bubble to outer catch"). The original intent was graceful degradation — an orders-lookup failure would still return the user's name/email with an empty `recentOrders`. After this change, an orders-lookup failure returns a fully empty context (`{ page: '', userName: '', ... }`). This is a behavioral regression for the chat agent: logged-in users whose order query fails will appear anonymous.

The outer-catch `logError` is good for observability, but the graceful-degradation should be preserved. Suggested fix: keep the inner try/catch, add `logError` inside it (non-fatal warning tag), and let the outer catch handle the truly fatal paths.

Everything else in the 5-file batch is straightforward. `swatchKitService.web.js` correctly wraps the non-Error `creditResult` in `new Error(creditResult?.error || 'credit_issuance_failed')` for error-monitoring compatibility.

CI is green (all checks pass). The cf-9eqp fixes are merge-ready pending resolution of the `getChatContext` graceful-degradation concern.

— morgott

PR #1071, opened 2026-04-17
feat(cf-5je): push notification preferences — managePushPreferences webMethod

DreadPirateRobertz:
## Summary
- Add `managePushPreferences` and `getMyPushPreferences` webMethods to `pushNotificationService.web.js` for member opt-in/out of push categories (challenges, streak, marketing, tier)
- Add `skipIfOptedOut` helper that gates push dispatch by checking member preferences; integrated into `crossRigEventReceiver` for `tier_changed` events
- Default all categories opt-in; stored in `PushPreferences` CMS collection with `memberId` + `categoryPrefs` JSON map
- 19 new tests covering single/bulk update, default behavior, IDOR guard, unmapped events, and cross-member isolation

## Test plan
- [x] 19 new tests in `tests/pushPreferences.test.js` — all pass
- [x] 66 existing tests in `pushNotificationService.test.js` + `crossRigEventReceiver.test.js` — 0 regressions
- [x] Pre-commit hooks (eslint + vitest) pass
- [ ] Verify `PushPreferences` CMS collection is provisioned in Wix dashboard

🤖 Generated with [Claude Code](https://claude.com/claude-code)

cf-5je

PR #1578, opened 2026-05-25
fix(cf-y2wg): post_submitSurvey IDOR pre-check queries wrong collection

DreadPirateRobertz:
## Why
\`http-functions.js:3637\` was hard-coded \`wixData.query('Survey')\` while \`surveyService.web.js\` writes rows into \`SurveyResponses\` (constant on line 28, used 6 places — \`scheduleSurvey\` inserts, \`submitSurveyResponse\` updates, \`getSurveyForOrder\` reads, aggregations query).

Every authenticated NPS submission hit the IDOR pre-check, found nothing in the non-existent \`Survey\` collection, and returned 404 "No survey found for this order" before ever delegating to the webMethod.

Existing tests at \`cfvtx5Dispatchers.test.js:347/387\` passed because they seeded the same wrong name (\`__seed('Survey', …)\`) — pinning a mismatched contract instead of the real one. Classic test-pins-the-bug.

## Fix
- Export \`SURVEY_COLLECTION\` from \`surveyService.web.js\` so wrapper + webMethod share a single source of truth.
- Import it in \`http-functions.js\` and use it in the pre-check query.
- Update existing tests to seed via the imported constant (not hard-coded \`'Survey'\`).
- **New regression-pin test** \`cf-y2wg: IDOR pre-check queries the SAME collection the webMethod writes to\` — seeds the legacy \`Survey\` name, asserts the wrapper does NOT find it (404), then re-seeds under \`SURVEY_COLLECTION\` and asserts the happy path (200). Future drift fails this test instead of silently shipping a 404.

## Why this PR alone isn't enough
Companion to **cfw PR DreadPirateRobertz/carolina-futons-web#1071** (forwards the \`Authorization: Bearer\` header from the cfw server action). Both are required for end-to-end NPS submissions to succeed:
- Before #1071: every cfw fetch was anonymous → 401 here. This PR's 404 fix wouldn't even be reached.
- Before this PR: even an authenticated caller with the bearer forwarded would 404 at the collection-name mismatch.

## Test plan
- [x] \`vitest run tests/cfvtx5Dispatchers.test.js\` — 42/42 green (was 41 + 1 new regression pin).
- [x] \`vitest run tests/{surveyService,cfvtx5Dispatchers,cf-1mlj-survey-*}.test.js\` — 132/132 green across all survey-touching tests.
- [ ] After both PRs deploy: log in as a member with a real \`SurveyResponses\` row → hit \`/survey?orderId=<that-order>\` → submit → verify \`completedAt\` + \`npsScore\` populated in the row.

## Refs
- Source bead: \`cf-y2wg\` (cfutons rig)
- Mirror bead: \`cfw-t4gl\` (cfutons_web rig — both halves tracked there)
- Companion cfw PR: DreadPirateRobertz/carolina-futons-web#1071

### DIFF
--- a/src/backend/http-functions.js
+++ b/src/backend/http-functions.js
@@ -46,7 +46,7 @@ import * as _pushNotificationServiceModule from 'backend/pushNotificationService
 // block (`import * as wishlistServiceModule from 'backend/wishlistService.web';`).
 // Don't re-import here.
 import * as _styleQuizModule from 'backend/styleQuiz.web';
-import { submitSurveyResponse } from 'backend/surveyService.web';
+import { submitSurveyResponse, SURVEY_COLLECTION } from 'backend/surveyService.web';
 import { grantSpin } from 'backend/spinRedemptionService.web';
 import { submitCommunityPhoto } from 'backend/communityPhoto.web';
 // cf-bkxh: namespace import for giftRegistry dispatcher.
@@ -3634,7 +3634,10 @@ export async function post_submitSurvey(request) {
     });
   }
   try {
-    const lookup = await wixData.query('Survey')
+    // cf-y2wg: single source of truth on collection name — was hard-
+    // coded 'Survey' and silently mismatched the webMethod's
+    // SurveyResponses, returning 404 for every legitimate caller.
+    const lookup = await wixData.query(SURVEY_COLLECTION)
       .eq('memberId', member._id)
       .eq('orderId', orderId)
       .limit(1)
--- a/src/backend/surveyService.web.js
+++ b/src/backend/surveyService.web.js
@@ -25,7 +25,12 @@ import { currentMember } from 'wix-members-backend';
 import { sanitize } from 'backend/utils/sanitize';
 import { logError } from 'backend/utils/errorHandler';
 
-const SURVEY_COLLECTION = 'SurveyResponses';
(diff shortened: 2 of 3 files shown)


DreadPirateRobertz:
## 5-Agent PM Review — PR #1578 (cf-y2wg: SURVEY_COLLECTION IDOR fix)

**Score: 93/100 — APPROVED, ADMIN MERGING**

### What it does
Exports `SURVEY_COLLECTION = 'SurveyResponses'` from `surveyService.web.js` so the HTTP wrapper in `http-functions.js` queries the same collection the webMethod writes to. Pre-fix, the wrapper hard-coded `'Survey'` and the 404 pre-check silently blocked every legitimate survey submission.

### Strengths
- One source of truth: constant exported and imported in both files — regression-safe ✅
- Mock updated (`SURVEY_COLLECTION: 'SurveyResponses'`) — test isolation preserved ✅
- Regression pin test added: assert both sides use the same constant ✅
- `__seed(SURVEY_COLLECTION, ...)` in test setup — fixture matches production ✅
- Minimal scope: only 3 files touched ✅

### CI failures
- `test (20)` / `test (22)`: `ctaCoralAudit.test.js` + `fabricSwatchesPage.test.js` — **pre-existing on main** (main CI also fails these). Unrelated to this PR's changes.
- lint ✅ CodeQL ✅

Admin-merging now. cf-y2wg → close on merge.

